PR #22, opened 2017-12-17
Fix #21: High CPU usage in a small project

### LINKED ISSUE #21: High CPU usage in a small project
Holy cow this is some high CPU usage:

image

Rojo needs to _not_ do that. My hunch is that the file watcher either has a bug or is falling back to polling because of some missing API support.

Quenty:
The infinite loops that are keeping the threads alive are what really was killing CPU usage. Sleeping for 100 ms seems to work. Not sure how exactly rust async works right now, so this may cause delay of up to 1 second before polling, but filesystem already uses that delay.

Preferably, the webserver also just uses the main thread, instead of creating its own. Busy polling is still bad.

### DIFF
--- a/src/bin.rs
+++ b/src/bin.rs
@@ -22,7 +22,6 @@ pub mod vfs_watch;
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, Mutex};
 use std::thread;
-use std::time::Duration;
 
 use core::Config;
 use pathext::canonicalish;
--- a/src/bin.rs
+++ b/src/bin.rs
@@ -195,7 +195,7 @@ fn main() {
             println!("Server listening on port {}", port);
 
             loop {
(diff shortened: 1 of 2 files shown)


Quenty:
Note: There should also be some sort of integration test to verify this isn't an issue in the future.

Quenty:
See #21

src/bin.rs
```diff
@@ -22,6 +22,7 @@ pub mod vfs_watch;
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, Mutex};
 use std::thread;
+use std::time::Duration;
```
LPGhatguy: We can get rid of this import now -- Rust should issue a warning

Quenty:
Removed the unnecessary import. Should be ready for pull.

PR #56, opened 2018-04-04
WIP: Protocol v2 (0.5.0)

LPGhatguy:
This is basically a continuation of #44, since something mysterious happened that killed the pull request.

This time, the design spec will live in `DESIGN.md` for the entirety of this development, instead of part of a pull request description.


Kampfkarren:
You probably accidentally deleted the branch when pushing, somehow.

LPGhatguy:
I think I'm just about done stripping out all of the stuff from the `master` branch that won't be used in the new protocol.

LPGhatguy:
The branches have diverged pretty badly, so I imagine that in general moving forward, I'll cherry-pick or manually merge commits from master into this branch.

LPGhatguy:
I stripped and stalled pretty hard trying to organize the dance between FS data and Roblox instances -- the threading and channel synchronization is sort of tricky!

LPGhatguy:
An interesting potential design note: what if there was a single port that Rojo used (something large and random and unlikely to conflict) that ran only once per computer?

The first Rojo session you run would start that server if it isn't already running. Each instance could tell the existing running server about its projects, which would mean we don't need to do port negotiation and can have a really nice GUI for selecting projects when syncing.

Some potential problems:

* There could be a race condition with multiple Rojo instances checking for the server and then starting it themselves. Especially with Rouille failing to report if a port is in use, that could be bad!
* Handing off control of the server seems like it would be tricky, since if you start two Rojo sessions and terminate the first, you don't want to lose your session for any of your projects. We basically want a reference-counted process or potentially a process that starts and *never* terminates except when requested. This is how Facebook's Flow checker works, for example.

LPGhatguy:
Rewrite is to a pretty good spot right now:

* Server now keeps an in-memory filesystem that (should?) be correctly synced up with the real filesystem.
* Prototype-level implementation of Roblox object generation and reconciliation is implemented in the server. There are cases where it falls apart pretty quickly, and I think some of the data structure might need to be rethought.
* Plugin can communicate via long-polling and is guaranteed not to miss any messages. Plugin also has a simple 'reconciler' that should be accurate, but is definitely prototype-level code.

LPGhatguy:
This is going to become `master` somewhat soon, and the current `master` branch will be demoted to a new branch named `v0.4.x`.

PR #58, opened 2018-04-07
Open Http Properties upon HttpEnabled prompt

Validark:


LPGhatguy:
Neat, that's a cool way to point people to a solution.

The one issue I see is that the recommended way to change `HttpEnabled` is now under game settings instead of the (now hidden by default) `HttpService`. It's possible that `HttpEnabled` will start being hidden on `HttpService` in Studio someday soon.

Validark:
I disagree. Here is the description of the [Set function](http://wiki.roblox.com/index.php?title=API:Class/Selection/Set):

> Set's the client's selected objects to all of those in the table. You can select objects you wouldn't normally be able to like DataModel, SocialService, and other non-browsable Instances.

Quenty:
This issue here isn't that the `SelectionService:Set()` would stop letting you select the service. It's that Roblox can control what properties show up in the properties window. So even if you select the non-browsable instance `HttpService` there's no guarantee that `HttpEnabled` is a property you can see. 

That being said, I wish there was a more elegant solution.

Validark:
Well, we all use this, so we can easily patch it should Roblox ever change it, but I HIGHLY doubt it will be changed for many, many years. What reason would they have for going back and removing this specific use case? Even if HttpEnabled eventually goes away, this code won't break anything. Worst case scenario is it opens an empty property box and you still have to go to Game Settings.

LPGhatguy:
Fair point, we can always remove this feature if `HttpEnabled` isn't visible anymore.

PR #109, opened 2019-01-19
Do Not Merge: Server plugins prototype

LPGhatguy:
This is a prototype implementation of server plugins.

In this implementation, there are a couple caveats:
* Plugins are defined statically; they should be a project key
* Plugins take bincode input and output JSON; these formats should match and probably not be bincode
* The plugin interface should probably be a little more flexible, see comments inline

I also implemented a functional MoonScript plugin that calls out to the user's installed `moonc` executable.

For a real plugin system, we probably want to support different plugin actions as well as some sort of plugin capabilities API so that plugins can indicate what plugin protocol version they want to use, whether they support two-way sync, etc.


LPGhatguy:
Superseded by #125.

PR #120, opened 2019-02-01
Implement new project file name, default.project.json

### LINKED ISSUE #118: Change project name to something uniquely identifiable as a Roblox project
We should change the project names from `roblox-project.json` to something like `roblox.project.json` so that we're able to distinguish alternate project configurations from arbitrary json files.

This would open up the ability for someone to depend on `SomeLibrary/debug.project.json`, which might have some additional logging or do some transformation via server plugins.

LPGhatguy:
Closes #118.

Changes the default project file name from `roblox-project.json` to `default.project.json`. This opens the door to have other project file names while still allowing Rojo to differentiate project files from random JSON files.

Introduces a simple project compatibility mechanism so that the old default project file name (`roblox-project.json`) is still supported, but users will be warned to upgrade.


server/src/project.rs
```diff
@@ -405,6 +413,25 @@ impl Project {
...
+            .to_str().expect("Project file path was not valid Unicode!");
+
+        if file_name == COMPAT_PROJECT_FILENAME {
+            warn!("Rojo's default project file name changed in 0.5.0-alpha3.");
+            warn!("Support for the old project file name will be dropped before 0.5.0 releases.");
+            warn!("Your project file is named {}", COMPAT_PROJECT_FILENAME);
+            warn!("Rename your project file to {}", PROJECT_FILENAME);
+        } else if !file_name.ends_with(".project.json") {
+            warn!("Starting in Rojo 0.5.0-alpha3, it's recommended to give all project files the");
+            warn!(".project.json extension. This helps Rojo differentiate project file from");
```
ZoteTheMighty: "...helps Rojo differentiate project **files** from"

PR #122, opened 2019-02-05
WIP: Support nested partitions and partitions directly targeting services

### LINKED ISSUE #102: Allow nested partitions
With 0.5.x, it should be possible to _intentionally_ nest partitions.

### LINKED ISSUE #112: Allow overriding root instance of partition
In order to support syncing directly into services or into other interesting instance types, it'd be useful to specify the name, class, and properties of a partition root in addition to a path to sync from.

I imagine it would look something like this in a project:

```json
"ReplicatedStorage": {
	"$path": "src/replicated-storage"
}
```

The extra metadata available to configure project nodes will be relevant here most likely, like `$ignoreUnknownInstances`.

LPGhatguy:
Closes #112. Closes #102.

This makes the project format significantly easier to reason about by turning both `SourceProjectNode` and `ProjectNode` into structs instead of enums with associated data, like they are today.

Error messages should be much better, and it opens up the possibility to specify `$className`, `$children`, and `$properties` on the same partitions that you can specify `$path` on. It should be a backwards-compatible change, except in the case of previously malformed projects.

## TODO
* [ ] Custom `Deserialize` impl to give better error messages on project structure
* [ ] Documentation
* [ ] More exhaustive tests


LPGhatguy:
It looks like we might have to start storing actual `ProjectNode` objects in the path-per-instance metadata object so that we can correctly reconstruct children and properties specified by the project on each change.

LPGhatguy:
Build looks like it might be failing due to how pretty assertions does its comparison?

PR #125, opened 2019-02-13
WIP: Server plugins via rlua (Lua 5.3)

### LINKED ISSUE #55: Support custom plugins
Implementing this, alongside plugin config, would unblock a lot of people and let them have custom workflows.

LPGhatguy:
Almost finishes #55. Further work will be done, as well as a feature stabilization post-merge.

This is a work-in-progress implementation of server plugins. This approach embeds rlua, a safe wrapper around Lua 5.3, and uses Lua as the execution strategy.

Compared with the previous plugin prototype, #109, this approach:

- Does not require additional tools to be installed for simple scripting
- Has the correct mechanisms for loading plugins from project files

The plugin API is still not defined. I imagine it'll be a combination of:

- Injecting a `rojo` global for extra APIs if they're necessary (extra filesystem APIs, HTTP, etc)
- Plugins returning tables that contains a description of features, a basic file/instance filter, and methods to invoke to transform instances or files.


LPGhatguy:
I added a feature gate for plugins so that we can move forward with merging this PR without nailing a lot of the details down. The `plugins` field on projects will be ignored unless it's set, and no Lua state will be created unless it's set.

Once the Rust portion of the test suite passes (Lua has been broken because of Lemur for a bit), I think this will be good to go.

PR #155, opened 2019-04-05
Update mention of VS Code extension in docs

evaera:
Also changed verbiage from "add partitions" to "sync files"


LPGhatguy:
Cool, thanks!

PR #162, opened 2019-04-22
Plugin autostart

### LINKED ISSUE #151: Plugin autostart
I'd like to be able to generate place files that have a special marker that the plugin recognizes, causing it to automatically try to connect to a running Rojo session.

It's really two tightly-coupled features:
- A flag to `rojo build` that injects a non-archivable instance into the `DataModel` with a special name
- A plugin feature that looks for that special instance on startup and automatically starts a session. This should probably have a filter on the project's name or maybe use an ephemeral port to avoid an accidental collision.

This would lay the groundwork for projects to use a single command to build a place file, open it in studio, and start live syncing. We could bundle that together into a command with a name like `rojo start` if it proves ubiquitous enough!

LPGhatguy:
Closes #151.

This PR makes the Rojo plugin recognize a special marker value (`ROJO_AUTOSTART_PORT`), causing it to immediately connect to a running Rojo server.

It also adds a new flag to `rojo build`, `--plugin_autostart`, which will attach a special marker value when used to build a place file.

This means that it's possible to get up and running with a place using Rojo with an easily-scriptable sequence of a couple commands:

```sh
rojo build --plugin_autostart -o Place.rbxlx
start Place.rbxlx # Open Roblox Studio
rojo serve
```


LPGhatguy:
I think before this change, it probably needs to be locked to a project name to prevent accidentally syncing the wrong place in.

I'm afraid there might be some debt to clean up before it's easy to make that change.

LPGhatguy:
I think this change should also not affect the generated place file, ideally. Experimenting with this feature, I have accidentally re-opened a place only to accidentally immediately sync.

PR #165, opened 2019-04-25
Use rbx_dom_weak for reconciler

Kampfkarren:
Has issues with Ref, but works with tags.


plugin/src/Reconciler.lua
```diff
@@ -27,6 +28,29 @@ function Reconciler:applyUpdate(requestedIds, virtualInstancesById)
 	end
 end
 
+function Reconciler:mapProperties(virtualProperties, virtualInstancesById)
+	-- Maps properties from the Rojo format to one rbx-dom-lua uses
```
LPGhatguy: Ideally these should be the same, since Rojo's format is just `rbx_dom_weak`'s format round-tripped through JSON.

plugin/src/Reconciler.lua
```diff
@@ -27,6 +28,29 @@ function Reconciler:applyUpdate(requestedIds, virtualInstancesById)
...
+	local properties = {}
+
+	for propertyName, property in pairs(virtualProperties) do
+		if property.Type == "BinaryString" then
+			local binaryString = ""
+			for _, characterCode in pairs(property.Value) do
+				binaryString = binaryString .. string.char(characterCode)
+			end
+			properties[propertyName] = binaryString
```
LPGhatguy: Hmm, this looks like something to fix in the way `rbx_dom_weak` is serialized/deserialized! Maybe we can switch to base64 encoding along the lines of how the XML format works, and then keep them as strings.

plugin/src/Reconciler.lua
```diff
@@ -53,11 +77,9 @@ function Reconciler:reconcile(virtualInstancesById, id, instance)
...
 
 	-- Some instances don't like being named, even if their name already matches
-	setCanonicalProperty(instance, "Name", virtualInstance.Name)
-
-	for key, value in pairs(virtualInstance.Properties) do
-		setCanonicalProperty(instance, key, rojoValueToRobloxValue(value))
-	end
+	writeInstance(instance, {
+		Name = virtualInstance.Name,
+	})
```
LPGhatguy: We should probably have an API that writes a single property instead of using this general purpose one for names and stuff

plugin/src/Reconciler.lua
Kampfkarren: Usually it is, only the hack for BinaryString and the thing used to get the instance from the Ref ID are different.

plugin/src/Reconciler.lua
Kampfkarren: And then what, decode from rbx_dom_lua?

plugin/src/Reconciler.lua
Kampfkarren: So write an inline function in the Reconciler or this should be in rbx_dom_lua?

LPGhatguy: Probably something in rbx\_dom\_lua. I'd like to revisit these APIs before we dive so deep into trying to integrate them into Rojo.

Kampfkarren: I've sent a PR for this on rbx_dom_lua a bit ago.

Kampfkarren:
Closing as per request of LPG

PR #168, opened 2019-05-01
Merge instances between filesystem and project nodes

### LINKED ISSUE #167: Rojo should merge instances from the filesystem and project
Given this node from a project:

```json
"ReplicatedStorage": {
    "$path": "src/ReplicatedStorage",
    "$ignoreUnknownInstances": true,

    "Subfolder": {
        "$path": "src/ReplicatedStorage/Subfolder",
        "$ignoreUnknownInstances": true
    }
}
```

If `src/ReplicatedStorage/Subfolder` exists on the filesystem, Rojo will currently create two instances called `Subfolder` and cause all sorts of weird havoc in Roblox.

Instead, Rojo should probably merge the two instances. This would let users overwrite properties on deeply-nested instances specified on-disk!

LPGhatguy:
Fixes #167.

This change allows users to specify project nodes that'll be applied to existing instances, like ones that come from folders or from within model files.

I added a test project (`test-projects/fs_project_merging`) that illustrates this:

```json
{
  "name": "fs_project_merging",
  "tree": {
    "$path": "outer",

    "inner": {
      "$path": "outer/inner",
      "$ignoreUnknownInstances": true
    }
  }
}
```

---

Before this PR:

image

---

After this PR:

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 * Added support for implicit property values in JSON model files ([#154](https://github.com/LPGhatguy/rojo/pull/154))
+* Fixed minor bugs when filesystem paths defined in projects alias eachother
+* Added support for patching objects defined on the filesystem by defining an instance in your project with the same name. ([#168](https://github.com/LPGhatguy/rojo/pull/168))
 
 ## [0.5.0 Alpha 9](https://github.com/LPGhatguy/rojo/releases/tag/v0.5.0-alpha.9) (April 4, 2019)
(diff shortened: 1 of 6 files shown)


LPGhatguy:
After trying to stress the edge cases of this idea, there are some interesting problems:
- Should child nodes specify `$path`? Right now, if they do, it doesn't do anything until the path is updated
- Should this be narrowed to just overriding items on the filesystem?
- Is this approach worthwhile, or should I just implement `init.meta.json` instead?

LPGhatguy:
This should be superseded by #183.

PR #170, opened 2019-05-14
Added MoonScript support to Rojo

Airwarfare:
This PR allows the use of MoonScript by simply adding .moon to your files, note this does require you to have `moonc` in your PATH if you want to use moonscript, also sends compile errors to the console and VS Code if you use the rojo extension, it does not send the error to studio (yet).

### DIFF
--- a/server/src/rbx_snapshot.rs
+++ b/server/src/rbx_snapshot.rs
@@ -16,6 +16,7 @@ use maplit::hashmap;
 use rbx_dom_weak::{RbxTree, RbxValue, RbxInstanceProperties, UnresolvedRbxValue};
 use serde::{Serialize, Deserialize};
 use rbx_reflection::{try_resolve_value, ValueResolveError};
+use std::process::Command;
 
 use crate::{
     imfs::{
@@ -376,6 +377,7 @@ fn snapshot_imfs_file<'source>(
                 None
             }
         },
+        Some("moon") => snapshot_moon_file(file)?,
         Some(_) | None => None,
     };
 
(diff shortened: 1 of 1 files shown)


Kampfkarren:
I believe this is better fit for the upcoming plugins rather than being in the actual build of Rojo.

Airwarfare:
That seems like a bit bunch for something quite small in content (45 lines), I think this should be a standard feature in Rojo, its actually one of the reasons I don't use Rojo as much and instead use RbxSync since it supports this feature, it also says in the readme of the repo "Soon, Rojo will be able to: . . . Compile MoonScript and other custom things for your project". Just my thought, thanks for the feedback I appreciate it!

LPGhatguy:
I'm not entirely opposed to MoonScript support to Rojo core, but what @Kampfkarren said is accurate overall.

The current plan for Rojo plugins is to run Lua, which should mean it'll be around 30 lines of Lua to do the same thing in your own project if you're interested in MoonScript or anything else! 😄 

Sadly, I've been stretched pretty thin working on other parts of Rojo and haven't gone back around to finish up plugin support. If someone wanted to champion that (breaking out the Lua glue code + state into its own module, fleshing out a prototype API, etc) we could probably get MoonScript working through that avenue pretty quickly.

Airwarfare:
Alright, thanks for the info, unfortunately I don't have enough experience with the codebase and rust to be comfortable with contributing that but hopefully sometimes soon does and we can have that feature up and working 😃

LPGhatguy:
The codebase has undergone a lot of changes since this PR was filed and it will no longer merge, sadly!

I think based on the discussion we had, we can close this PR and revisit the discussion as part of some sort of plugins.

Issue #103, opened 2019-01-16
Preserve port number in connection dialog


LPGhatguy:
At least within the same session, Rojo should begin with the settings used in the last connect.

That way, if the previous connection disconnected, it's easier to reconnect to the same server without having to dig up the port number again.

Sleitnick:
I would really like to see this. I see this has been open for over a year now though. Is there any roadmap for this @LPGhatguy?

PR #178, opened 2019-05-31
Create a PluginAction for toggling connection

vocksel:
Add's a user-configurable shortcut to start/stop connections to the Rojo server.

image

Had to hoist up some state from ConnectPanel which I would rather have not done, but not sure if there's a better way without Rodux.

### DIFF
--- a/plugin/src/Components/App.lua
+++ b/plugin/src/Components/App.lua
@@ -68,6 +68,8 @@ local App = Roact.Component:extend("App")
 
 function App:init()
 	self:setState({
+		address = "",
+		port = "",
 		sessionStatus = SessionStatus.Disconnected,
 	})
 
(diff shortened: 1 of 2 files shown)


LPGhatguy:
Hey, this fixes #103, too!

LPGhatguy:
Finally getting a chance to look at this!

Mechanically, this change looks pretty much perfect. I'm not super sure about the use case for toggling the connection with a hotkey though! What makes you want to do this kind of thing, and is there something else we can fix that might be more direct?

vocksel:
My biggest motivation is the workflow of opening the ribbon bar, clicking the plugin button, then moving the mouse to select the connect button is tedious. I know that yall are looking to have multiple connections tho, so I understand a single "connect" may not be entirely possible anymore.

If the plugin had an icon so I could cleanly have it on the quick access bar without having the "Connect" text there, that would help a lot, and could be a nice alternative if you don't want to use plugin actions.

vocksel:
This PR is well over a year old and I don't see this workflow as necessary. Gonna go ahead and close it down so it's not floating around anymore.

PR #183, opened 2019-06-03
init.meta.json support

### LINKED ISSUE #123: Support .meta.json
This is an alternative to `init.model.json` support (#114) and opens up a lot of neat features like allowing scripts to be disabled (#34) and giving us a place to stick metadata about uploaded assets (#41).

It'll look like a pared-down version of `.model.json` files, supporting just properties, and maybe asset-specific metadata:

```json
{
    "properties": {
        "Disabled": {
            "Type": "Bool",
            "Value": true
        }
    },
    "uploadedAssetId": "12342124"
}
```

Kampfkarren:
Closes #123.

So far only supports `className`, want to know if this is what you expected before I work on this more.


Kampfkarren:
Can't figure out why it's failing?

server/src/rbx_snapshot.rs
```diff
@@ -315,13 +336,13 @@ fn snapshot_imfs_directory<'source>(
...
-            properties: HashMap::new(),
+            properties: meta.properties,
             children: Vec::new(),
             metadata: MetadataPerInstance {
                 source_path: None,
-                ignore_unknown_instances: false,
+                ignore_unknown_instances: meta.ignore_unknown_children.unwrap_or(false),
                 project_definition: None,
             },
         }
```
LPGhatguy: Instead of applying the metadata in the else branch here, can we apply the metadata unconditionally after this branch? It might even be a good idea to defer loading the `init.meta.json` file until that point.

This branch is only applied when there is no `init.lua`, `init.server.lua`, or `init.client.lua` file in the folder. It's still useful to apply metadata to the instance even if it becomes a script by one of those mechanisms!

server/src/rbx_snapshot.rs
```diff
@@ -351,6 +372,20 @@ fn snapshot_imfs_directory<'source>(
...
+struct InitMetaJson {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    class_name: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    ignore_unknown_children: Option<bool>,
+    #[serde(
+        default = "HashMap::new",
+        skip_serializing_if = "HashMap::is_empty",
+    )]
+    properties: HashMap<String, RbxValue>,
```
LPGhatguy: Do you want to add an empty line before each property that has an attribute?

I've found that it's easier to read structs when they're spaced out once attributes like these are involved.

server/src/rbx_snapshot.rs
Kampfkarren: I believe I did what you asked.

server/src/rbx_snapshot.rs
```diff
@@ -348,9 +374,29 @@ fn snapshot_imfs_directory<'source>(
         }
     }
 
+    if let Some(meta_ignore_instances) = meta.ignore_unknown_instances {
+        snapshot.metadata.ignore_unknown_instances = meta_ignore_instances;
+    }
```
LPGhatguy: Can we bump this block to be above where we set children? We're already assigning to the snapshot's metadata and already have a change there.

server/src/rbx_snapshot.rs
```diff
@@ -348,9 +374,29 @@ fn snapshot_imfs_directory<'source>(
...
+
     Ok(Some(snapshot))
 }
 
+#[derive(Debug, Default, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct InitMetaJson {
```
LPGhatguy: I don't think we need the `Json` part of this struct name.

Since we don't serialize the struct, we might be able to drop the `Serialize` derive and all of the `skip_serializing_if` attributes as well.

server/src/rbx_snapshot.rs
```diff
@@ -299,6 +309,17 @@ fn snapshot_imfs_directory<'source>(
...
+
+    let meta: InitMetaJson = if let Some(ImfsItem::File(file)) = imfs.get(&init_meta_path) {
+        serde_json::from_slice(&file.contents)
+            .map_err(|inner| SnapshotError::InitMetaError {
+                inner,
+                path: file.path.to_path_buf(),
+            })?
+    } else {
+        InitMetaJson::default()
+    };
```
LPGhatguy: I think we can fold this code down even further!

Instead of assigning the struct out here and defaulting to `InitMetaJson`, can we check for the existence of this file later, and then pull properties directly from it onto the snapshot?

Then we can localize the changes to this function to exactly one region, and limit the impact in cases that _don't_ use an `init.meta.json` file.

Kampfkarren:
@LPGhatguy Done in latest commit

Issue #179, opened 2019-05-31
Infer ClassName for children of DataModel in project


LPGhatguy:
Suggested by @osyrisrblx!

Given this project:
```json
{
  "name": "Cool Demo",
  "tree": {
    "$className": "DataModel",

    "ReplicatedStorage": {
      "Code": {
        "$path": "src"
      }
    }
  }
}
```

Right now, Rojo will complain that `ReplicatedStorage` doesn't have a `$className` set. The only correct choice there is `"ReplicatedStorage"` though, so we should be able to infer it!

We should be able to use the `is_service` tag from rbx_reflection, and maybe pass an extra tag down the tree to let the algorithm know that the last instance was a `DataModel`. This might get kind of tricky when we're trying to do reconciliation of a small part of the tree.

PR #196, opened 2019-06-13
Default class name of Folder

### LINKED ISSUE #157: Make project nodes default to having a ClassName of Folder
@Osyrisrblx suggested this idea!

If neither `$path` nor `$className` are specified, it would be useful if the node defaulted to having a `$className` of `Folder`. Since this kind of node isn't valid right now, this would be a handy way to cut down on boilerplate.

Kampfkarren:
Closes #157.

Seemed a bit too easy, so curious if I am missing something, I just removed the match.


LPGhatguy:
I'm not sure we want this as a feature yet. It closes off our ability to start doing `ClassName` inference, since it's super possible that people will rely on things becoming folders.

At the very least, I want to get #179 done before we start thinking about this.

Kampfkarren:
:+1:

PR #198, opened 2019-06-14
Reconciler Mk4

LPGhatguy:
This is an attempt to simplify and unify a lot of the changes that have happened to Rojo's reconciler 
recently.

I'll flesh out docs as more of the design comes into place.

As part of this change, or shortly after, we should be able to mostly remove the reconciler from the plugin in favor of having smaller patches and fewer network round-trips.


server/src/instance_snapshot/patch_compute.rs
```diff
@@ -0,0 +1,147 @@
...
+    snapshot: &'a InstanceSnapshot,
+    tree: &RbxTree,
+    id: RbxId,
+    patch_set: &mut PatchSet<'a>,
+) {
+    let instance = tree.get_instance(id)
+        .expect("Instance did not exist in tree");
+
+    if instance.class_name != snapshot.class_name {
+        panic!("NYI: changing class name of an instance?");
```
Kampfkarren: NYI = Not Yet Implemented, so why not use `unimplemented!("changing class name of an instance?")`

LPGhatguy: I didn't know `unimplemented!` accepted a string argument, good idea!

LPGhatguy:
Replaced by #217.

PR #210, opened 2019-07-03
Infer class name

### LINKED ISSUE #179: Infer ClassName for children of DataModel in project

jeparlefrancais:
Closes #179 

This will remove the necessity to write the `$className` field for services.


server/src/rbx_snapshot.rs
```diff
@@ -14,6 +14,7 @@ use failure::Fail;
 use log::info;
 use maplit::hashmap;
 use rbx_dom_weak::{RbxTree, RbxValue, RbxInstanceProperties, UnresolvedRbxValue};
+use rbx_reflection::{get_class_descriptor};
```
Kampfkarren:
```suggestion
use rbx_reflection::get_class_descriptor;
```

server/src/rbx_snapshot.rs
```diff
@@ -194,6 +195,21 @@ pub fn snapshot_project_tree<'source>(
     snapshot_project_node(context, imfs, &project.tree, Cow::Borrowed(&project.name))
 }
 
+pub fn infer_class_name(
+    instance_name: Cow<str>,
+) -> Option<Cow<str>> {
+    match get_class_descriptor(&instance_name) {
```
Kampfkarren:
```
match get_class_descriptor(&instance_name) {
    Some(instance) if instance.is_service() => Some(instance_name),
    _ => None,
}
```

jeparlefrancais: I did not know about this syntax! Thanks for the tip

jeparlefrancais:
I added a snapshot test, but I'm not sure I got everything that is going on in the expected-snapshot file so I really need someone to double check this carefully. The test passes but I just want to make sure it actually test something!

Kampfkarren:
Yeah that test looks fine to me, if there's no `expected-snapshot.json` then Rojo makes one, and if there is it uses that. So long as you can run `cargo test` twice without it failing, then you're good.

server/src/rbx_snapshot.rs
```diff
@@ -202,24 +212,33 @@ pub fn snapshot_project_node<'source>(
...
+            let class_name = match &node.class_name {
+                Some(_class_name) => Cow::Borrowed("Folder"),
+                None => {
+                    match infer_class_name(instance_name.clone()) {
+                        Some(inferred_class_name) => inferred_class_name,
+                        None => {
+                            return Err(SnapshotError::ProjectNodeUnusable);
+                        }
+                    }
+                },
```
LPGhatguy: I think we can drop the braces here so that the `None` match cascades right into another `match`.

We _also_ should be able to write the inner match as

```rust
infer_class_name(instance_name.clone()).ok_or(SnapshotError::ProjectNodeUnusable)?
```

so something like

```rust
            let class_name = match &node.class_name {
                Some(_) => Cow::Borrowed("Folder"),
                None => infer_class_name(instance_name.clone())
                    .ok_or(SnapshotError::ProjectNodeUnusable)?,
            }
```

server/src/rbx_snapshot.rs
```diff
@@ -202,24 +212,27 @@ pub fn snapshot_project_node<'source>(
...
     let maybe_snapshot = match &node.path {
         Some(path) => snapshot_imfs_path(context, imfs, &path, Some(instance_name.clone()))?,
-        None => match &node.class_name {
-            Some(_class_name) => Some(RbxSnapshotInstance {
+        None => {
+            let class_name = match &node.class_name {
+                Some(_class_name) => Cow::Borrowed("Folder"),
+                None => infer_class_name(instance_name.clone())
+                    .ok_or(SnapshotError::ProjectNodeUnusable)?,
+            };
```
LPGhatguy: Does the comment listed below still hold now that we set `class_name` to something besides `Folder`?

https://github.com/rojo-rbx/rojo/pull/210/files#diff-6dbbd20807104f15267e0279a30e80beR259

PR #216, opened 2019-07-15
Change project file to be JSON5 instead of JSON

### LINKED ISSUE #207: Allow comments in project file
Project files in Rojo are complicated. We can make them nicer with comments.

We could:
- use an ad hoc superset of JSON with comments
- use JSON 5, which has poor library support, sadly
- use YAML, which allows deep nesting like TOML, but has a huge spec

LPGhatguy:
Closes #207.

This PR changes the JSON implementation used for project files from serde_json to json5, which enables comments, trailing commas, unquoted keys, and more!

It makes the project file more ergonomic, but incompatible with standard JSON tools, which is a downside.


LPGhatguy:
This branch has drifted significantly from master since it was opened. Since it's so minor, I think we can re-do this work eventually.

PR #217, opened 2019-07-16
Reconciler Mk5

LPGhatguy:
This approach is the successor to #198. In this branch, the goal is to be less pedantic about child ordering than the reconciler-mk4 branch while still trying to accomplish the same overall goal.

This means that Rojo will not attempt to preserve child order across mutations. We should be able to alleviate issues that come out of this with build determinism through other means.


server/src/snapshot_middleware/rbxmx.rs
```diff
@@ -42,15 +42,13 @@ impl SnapshotMiddleware for SnapshotRbxmx {
...
-                Ok(Some(snapshot))
-            }
-            _ => panic!("Rojo doesn't have support for model files with multiple roots yet"),
+        if children.len() == 1 {
+            let mut snapshot = InstanceSnapshot::from_tree(&temp_tree, children[0]);
+            snapshot.name = Cow::Owned(instance_name);
+
+            Ok(Some(snapshot))
+        } else {
+            panic!("Rojo doesn't have support for model files with zero or more than one top-level instances yet.");
```
Kampfkarren: `unimplemented!`?

LPGhatguy: This isn't something that is going to be implemented here, at least not like this. I guess we could leave a comment to turn it into an Err.

Kampfkarren: It'd make sense to me this would be an `Err` then, yeah.

server/src/imfs/imfs.rs
```diff
@@ -0,0 +1,280 @@
+use std::{
+    io,
+    path::{Path, PathBuf},
+};
+
+use crate::path_map::PathMap;
+
+use super::{
+    snapshot::ImfsSnapshot,
```
Kampfkarren: Out of alphabetical order?

server/src/imfs/imfs.rs
```diff
@@ -0,0 +1,280 @@
...
+    pub(super) children_enumerated: bool,
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    use maplit::hashmap;
+
+    use super::super::noop_fetcher::NoopFetcher;
```
Kampfkarren: > `use super::super`

Is this something common? O_o

LPGhatguy: It's pretty awkward. It's referring to the `imfs` parent module, which is more difficult because all the tests are in a module module even deeper than the code it's testing...

server/src/snapshot_middleware/rbxm.rs
```diff
@@ -0,0 +1,96 @@
...
+    ) -> SnapshotInstanceResult<'static> {
+        if entry.is_directory() {
+            return Ok(None);
+        }
+
+        let file_name = entry.path()
+            .file_name().unwrap().to_string_lossy();
+
+        if !file_name.ends_with(".rbxm") {
+            return  Ok(None);
```
Kampfkarren: Double space. `cargo fmt` :eyes:

LPGhatguy: Yeaah, seems like it'd be a good idea to `cargo fmt` the whole codebase consistently after this PR is merged.

Kampfkarren: That'll fix the double space too, I s'pose.

Issue #33, opened 2018-02-04
Partition-level ignore field


LPGhatguy:
It'd be nice to be able to ignore instances by name per-partition.

For example, if I want to ignore my `StarterCharacter` inside of `StarterPlayer` because I don't want Rojo to manage it, I could use:

```json
{
	"partitions": {
		"StarterPlayer": {
			"path": "StarterPlayer",
			"target": "StarterPlayer",
			"ignore": [
				"StarterCharacter"
			]
		}
	}
}
```

I don't think this is a feature that supports the long-term vision of Rojo (using `rbxl` files as output binaries, and avoiding putting them into version control) but I do think it would ease adoption.

LPGhatguy:
I think this is actually a super useful feature even just to ignore all of the `.spec.lua` files that come with all of Roblox's open source projects right now.

LPGhatguy:
Since I made the original issue, this idea has been stewing in my head a little bit.

Now, I suspect that the feature would look sort of like:

```json
{
  "partitions": {
    "StarterPlayer": {
      "path": "StarterPlayer",
      "target": "StarterPlayer",
      "ignoreInstances": [
        "StarterCharacter",
        "StarterPlayerScripts.Chat"
      ],
      "ignoreFiles": [
        "*.spec.lua",
        "start-{tests,benchmarks}.server.lua"
      ]
    }
  }
}
```

LPGhatguy:
This came up as a related feature request, which is ignoring things only in live sync mode.

Rojo can handle a lot more things producing place files than it can in live syncing mode right now, like terrain, so it'd be useful to ignore it. Terrain can also be very large (on the order of tens of megabytes when base64 encoded), so letting the Rojo plugin skip work there would be huge.

Kampfkarren:
> Is there a way we can clarify that this setting only affects live-sync? 
$dontSync?

PR #223, opened 2019-08-06
Update Reconciler.lua

oniich-n:
Prevent PackageLink instances from being destroyed


LPGhatguy:
This is a pretty specialized fix!

It feels like we should aim to solve this problem (and others like it) through #33. You'd be able to keep an `init.meta.json` (or entry in your project file) in the spots where you're expecting `PackageLink` instances, which would tell Rojo to ignore them.

PR #242, opened 2019-08-30
Update Theme for Dark Mode

### LINKED ISSUE #241: Plugin Dark Theme
It would be nice for the plugin to match Studio's theme. Currently the light theme sticks out quite a bit in Studio. I know this isn't a super high priority, but would be a "nice to have" feature in my opinion.

Screenshot example:

rojo_theme

Jaguar-515:
This PR closes #241. 

The only problem I have with this is to change the background color (white) of the plugin, and I'm a little unsure on how to do that.

### DIFF
--- a/plugin/src/Theme.lua
+++ b/plugin/src/Theme.lua
@@ -11,9 +11,9 @@ local Theme = {
 	LightTextColor = Color3.fromRGB(160, 160, 160),
 }
 
---[[ local DarkTheme = {
+local DarkTheme = {
 
-	Little unsure on how to change the color of the background on the plugin for dark mode...
+	-- Little unsure on how to change the color of the background on the plugin for dark mode...
 
 	ButtonFont = Enum.Font.GothamSemibold,
 	InputFont = Enum.Font.Code,
@@ -25,7 +25,7 @@ local Theme = {
 	PrimaryColor = Color3.fromRGB(53, 53, 53),
 	SecondaryColor = Color3.fromRGB(103, 103, 103),
 	LightTextColor = Color3.fromRGB(103, 103, 103)
-} ]]--
+} 
 
 if Enum.UITheme == Enum.UITheme.Light then
 	setmetatable(Theme, {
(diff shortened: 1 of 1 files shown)


plugin/src/Theme.lua
```diff
@@ -11,10 +11,34 @@ local Theme = {
...
+	MainFont = Enum.Font.Gotham,
+	
+	AccentColor = Color3.fromRGB(89, 88, 88),
+	AccentLightColor = Color3.fromRGB(255, 146, 145),
+	PrimaryColor = Color3.fromRGB(53, 53, 53),
+	SecondaryColor = Color3.fromRGB(103, 103, 103),
+	LightTextColor = Color3.fromRGB(103, 103, 103)
+} 
+
+if Enum.UITheme == Enum.UITheme.Light then
```
LPGhatguy: I think there's a different API for getting what the actual theme is, I imagine this will always return false?

plugin/src/Theme.lua
```diff
@@ -11,10 +11,34 @@ local Theme = {
 	LightTextColor = Color3.fromRGB(160, 160, 160),
 }
 
-setmetatable(Theme, {
+local DarkTheme = {
+
+	-- Little unsure on how to change the color of the background on the plugin for dark mode...
```
LPGhatguy: We'll probably need to put the remaining colors into our theme tables!

plugin/src/Theme.lua
Kampfkarren: You'd be right. `settings().Studio.Theme`

plugin/src/Theme.lua
Jaguar-515: Yeah, we should.

plugin/src/Theme.lua
Jaguar-515: Just replaced that in the latest commit.

Sleitnick:
I know this complicates things a little bit, but it would be cool if this utilized [ThemeChanged](https://developer.roblox.com/en-us/api-reference/event/Studio/ThemeChanged) so it dynamically switched the colors if the user switches. That way the user doesn't have to restart Studio to get the theme to match.

Jaguar-515:
Due to me being lazy and not paying attention to this and the plugin's code being rewritten since this pull request was first opened, I'll close this PR.

Issue #169, opened 2019-05-10
Use Built-in Plugin Injection for Improved Live-Sync


Kampfkarren:
Right now, Rojo runs on the same security level as every other plugin which has its limitations. Things like Terrain and Union importing are going to need to go through the internal pipeline to become features that normal plugins to use. At the same time, these features are available right now hidden behind CoreScript APIs like InsertService:LoadLocalAsset. Theoretically, Rojo could install a companion plugin into BuiltInPlugins and use the content folder for symlinks, but this is a notable shift in mindset on how the plugin is developed. The steps can all be condensed to a basic command (as part of a `rojo install` perhaps) but the major issue is whether this is worth pursuing.

Reshiram110:
If it is pursued, this would allow rojo to have almost full support for loading instances from the file system. This would be beneficial because it would not only save time when converting projects to rojo (no need to seperate instances from scripts), but it would also open the possibility of version control for instances.

With this we could theoretically save a map model file to a repository, thus allowing for version control on game assets as well as game source code. This is quite common outside of Roblox, games made with unreal/unity/some engine store their game assets on the filesystem, and typically use a VCS to manage asset versions.

Kampfkarren:
> With this we could theoretically save a map model file to a repository, thus allowing for version control on game assets as well as game source code. 

You already can, look at [roads](https://github.com/LPGhatguy/roads). rbx-xml has basically 100% type coverage from what I know and something like #165 would let Rojo sync things like Tags easily. I specifically mentioned Terrain and Unions as things that are completely infeasible to sync for Rojo.

Reshiram110:
At the moment, instance support in rojo is flakey from what I've experienced when trying to save models to the filesystem - it typically errors with ScreenGUIs and the likes.

> Terrain and Unions
Yeah that's definitely another reason to do this, a lot of maps are bound to have terrain, unions and meshes. Saving the to the filesystem is not possible (or is highly unfeasible) at the moment.

Kampfkarren:
> At the moment, instance support in rojo is flakey from what I've experienced when trying to save models to the filesystem - it typically errors with ScreenGUIs and the likes.
That's solved with something like #165.

Reshiram110:
> Saving them to the filesystem is easy. Getting them from the filesystem is impossible

Oh, I thought it was the other way around, my mistake.
[...]

PR #249, opened 2019-09-23
Automatically enable the HttpService if running from BuiltInPlugins folder.

CloneTrooper1019:
If the Rojo plugin is running locally in Roblox Studio's BuiltInPlugins folder, then it has access to enable the HttpService automatically. Roblox added methods for manipulating its value for the Game Settings plugin. This commit adds code to try and enable the HttpService if the module detects that its running from the BuiltInPlugin security context level.

### DIFF
--- a/plugin/src/Http.lua
+++ b/plugin/src/Http.lua
@@ -8,13 +8,28 @@ local HttpResponse = require(script.Parent.HttpResponse)
 
 local lastRequestId = 0
 
+local isBuiltInPlugin = pcall(function ()
+	-- This is a hack, but there is no explicit way to detect the security context of a thread.
+	-- If this routine runs without throwing, then we are running from the BuiltInPlugins folder!
+	local CorePackages = game:GetService("CorePackages")
+	tostring(CorePackages)
+end)
+
 -- TODO: Factor out into separate library, especially error handling
 local Http = {}
+
+local function tryEnableHttp()
+	if isBuiltInPlugin and not HttpService:GetHttpEnabled() then
+		Logging.warn("Automatically enabling the HttpService.")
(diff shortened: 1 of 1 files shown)


LPGhatguy:
We talked about this a bit on Discord and I mentioned this a little bit in [this comment](https://github.com/rojo-rbx/rojo/issues/169#issuecomment-533935571) after we talked about it.

I don't think it's a good idea for Rojo to promote being put into built-in plugins.

The security responsibility of running inside built-in plugins is very high, and it puts a big target on Rojo. Built-in plugins have significant API access, which is one reason that it's likely that they'll be locked down in the future from Studio.

Rojo silently doing privileged things in the background like enabling HTTP is not good, since enabling HTTP is a security risk. This is a minor quality-of-life fix for a very high cost that impacts user's games. There's no way for Rojo to turn on HTTP without also turning it on for the developer's game, which is not good!

I would be open to a PR that makes this flow nicer without privilege escalation, like automatically selecting `HttpService` if Rojo tries to use it and it's disabled.

CloneTrooper1019:
Fair enough :) 

I just wanted to propose this change in case it was of interest to anyone else. I was a bit annoyed having to manually enable the HttpService myself everytime I opened an empty place. I'll keep this patch on my own fork of the plugin for however long its available.

LPGhatguy:
If you want an empty place that has HTTP enabled, you can actually use Rojo to build it instead of using Roblox Studio!

With the default project given to you with `rojo init`, you can run:

```bash
rojo build -o Place.rbxlx
start Place.rbxlx
```

...and you'll get a Roblox Studio instance with a place that has all the fixins you expect from a new place, along with HTTP requests enabled!

PR #259, opened 2019-10-15
VFS Improvements

LPGhatguy:
Each commit in this PR can be looked at independently, I made these changes as an incremental refactor of the VFS implementation.

This PR refactors all of the methods on `Vfs` from accepting `&mut self` to accepting `&self` and keeping data wrapped in a mutex. This builds on previous changes to make reference count file contents and cleans up the last places where we're returning borrowed data out of the VFS interface.

Once this change lands, there are two possible directions we can go that I see:
* Conservative: Refactor all remaining `&mut Vfs` handles to `&Vfs`
* Interesting: Embrace ref counting by changing `Vfs` methods to accept `self: Arc<Self>`, which makes the `VfsEntry` API no longer need an explicit `Vfs` argument for its operations.

### DIFF
--- a/src/vfs/real_fetcher.rs
+++ b/src/vfs/real_fetcher.rs
@@ -5,7 +5,7 @@ use std::{
     collections::HashSet,
     fs, io,
     path::{Path, PathBuf},
-    sync::mpsc,
+    sync::{mpsc, Mutex},
     time::Duration,
 };
 
@@ -34,7 +34,7 @@ pub struct RealFetcher {
     //
     // `watcher` must be dropped before `_converter_thread` or else joining the
     // thread will cause a deadlock.
-    watcher: Option<RecommendedWatcher>,
+    watcher: Option<Mutex<RecommendedWatcher>>,
 
     /// Thread handle to convert notify's mpsc channel messages into
     /// crossbeam_channel messages.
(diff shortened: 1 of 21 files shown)


src/vfs/real_fetcher.rs
```diff
@@ -45,7 +45,7 @@ pub struct RealFetcher {
 
     /// All of the paths that the fetcher is watching, tracked here because
     /// notify does not expose this information.
-    watched_paths: HashSet<PathBuf>,
+    watched_paths: Mutex<HashSet<PathBuf>>,
```
MichaelMcDonnell: Was it one Mutex before and now it is two? Does that mean we have to do twice as much locking? Is that ok from a performance perspective?

Are there any tight loops where the performance might be affected by having the mutex on the inside vs outside? And does that matter?

src/vfs/real_fetcher.rs
```diff
@@ -176,13 +178,16 @@ impl VfsFetcher for RealFetcher {
         }
     }
 
-    fn watch(&mut self, path: &Path) {
+    fn watch(&self, path: &Path) {
         log::trace!("Watching path {}", path.display());
```
MichaelMcDonnell: This method locks twice. Does it matter from a performance perspective? I guess we could add a bulk watch if it matters?

src/vfs/real_fetcher.rs
LPGhatguy: This PR just adds the internal mutex, the second one, #260 actually removes the outer mutex and stops us from needing to double-lock.

I branched off for #260 since I was a little less certain about those changes.

src/vfs/real_fetcher.rs
LPGhatguy: It's locking two different things that should be held on a short period, so it should be ok?

PR #266, opened 2019-11-12
Add warning when serving a non-project when project files exist

Kampfkarren:
image

### DIFF
--- a/src/common_setup.rs
+++ b/src/common_setup.rs
@@ -1,7 +1,7 @@
 //! Initialization routines that are used by more than one Rojo command or
 //! utility.
 
-use std::path::Path;
+use std::{fs, path::Path};
 
 use rbx_dom_weak::RbxInstanceProperties;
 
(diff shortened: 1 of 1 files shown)


LPGhatguy:
I think a better solution to this might be to require that `serve`/`build`/`upload` have a project, but add a flag to allow them to operate without a project.

That would help with footguns like this while also making intent much more clear.

LPGhatguy:
After using Rojo 0.6.0 more and more, I'm not sure it's actually a good idea to let Rojo operate without a project file 😅

I've hit _this_ particular footgun a lot working on Rojo. Running `rojo build` from the root of the repo results in a bunch of weird error messages after churning for a bit, when it should probably just immediately stop.

LPGhatguy:
Going to close this PR in favor of some work I'm doing right now to make Rojo go back to requiring a project. I thought I was being clever with this feature originally, but now I see that it was likely a mistake.

PR #268, opened 2019-11-26
Refactor snapshot context and implement glob ignores

LPGhatguy:
This PR accomplishes two goals:

1. A necessary refactor to change how context is passed around in the instance snapshot functions. Previously, it was only used by the unfinished plugin system. Now, context is actually hierarchical and inherited, making it much more useful.
2. Added the `$ignorePaths` property to project nodes, allowing users to specify glob patterns of files to ignore.

Example of ignores:
```json
{
  "name": "example-project",
  "tree": {
    "$path": "src",
    "$ignorePaths": [
      "src/**/*.spec.lua"
    ]
  }
}
```

## To do
- [ ] Split apart PR:
    - [x] #269 - Added support for `InstanceMetadata` builder
    - [x] #270 - Added new `InstanceContext` type
    - [x] #271 - Remove existing plugin code and use `InstanceContext` instead of existing type.
    - [ ] #272 - Implemented glob functionality in project and snapshot functions
- [ ] Investigate performance impact
- [ ] Improve feature test coverage
- [ ] Fix up references to `InstanceMetadata`
- [ ] Update code to not compile glob regex for each item


LPGhatguy:
All the necessary work should now be in other PRs, closing this one.

PR #272, opened 2019-12-04
Implement glob ignores

LPGhatguy:
Implements #272 behind a feature flag.

## Usage
1. Build Rojo with the `unstable_glob_ignore_paths` feature enabled.
2. Add the `globIgnorePaths` key to your project. The value should be a list of glob strings, relative to the folder containing your project file.

Example project:

```lua
{
  "name": "ignore_glob_spec",
  "tree": {
    "$path": "src"
  },
  "globIgnorePaths": [
    "**/*.spec.lua"
  ]
}
```

## Origin
Split off of #268, which originally contained a version of this work along with a bunch of other refactors that have since been split off and merged.


src/snapshot_middleware/dir.rs
```diff
@@ -26,11 +26,19 @@ impl SnapshotMiddleware for SnapshotDir {
...
 
-        let children: Vec<VfsEntry> = entry.children(vfs)?;
+        let passes_filter_rules = |child: &VfsEntry| {
+            for rule in context.path_ignore_rules.iter() {
+                if !rule.passes(child.path()) {
+                    return false;
+                }
+            }
+
+            true
```
Kampfkarren: Potentially more readable as:

```rs
context.path_ignore_rules.iter().all(|rule| rule.passes(child.path()))
```

LPGhatguy: Ahah! I originally wrote this with iterators but chose `any` instead of `all` which did not work. I think you're right that this is cleaner.

src/glob.rs
```diff
@@ -0,0 +1,68 @@
...
+
+impl Serialize for Glob {
+    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        serializer.serialize_str(self.inner.glob())
+    }
+}
+
+impl<'de> Deserialize<'de> for Glob {
+    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        deserializer.deserialize_str(GlobVisitor)
```
Kampfkarren: You can probably replace this with:

```rs
Glob::new(String::deserialize(deserializer))
```

... and rip out the visitor completely.

src/snapshot/metadata.rs
```diff
@@ -99,11 +100,57 @@ impl Default for InstanceMetadata {
...
+}
+
+impl InstanceContext {
+    /// Extend the list of ignore rules in the context with the given new rules.
+    pub fn add_path_ignore_rules<I: IntoIterator<Item = PathIgnoreRule>>(&mut self, new_rules: I) {
+        let new_rules = new_rules.into_iter();
+
+        // If the iterator is empty, we can skip cloning our list of ignore
+        // rules and appending to it.
+        if let Some(upper_bound) = new_rules.size_hint().1 {
```
Kampfkarren: It probably won't make a difference in practice, but size_hint is not required to give accurate results. Every place this is used also implements ExactSizeIterator, so you can add that trait bound and check is_empty.

src/glob.rs
LPGhatguy: Because `Glob::new` (both ours and the one from the globset crate) accept `&str` and not `String`, I think I'd rather keep the possibility that we can avoid copying the glob string here.

The visitor implementation isn't _so_ bad! 😄

src/snapshot/metadata.rs
LPGhatguy: According to the docs, if `size_hint` provides a non-`None` upper bound and it's wrong, we can consider the iterator to be wrong.

Requiring `ExactSizeIterator` is a good idea though, I think it'll make this code a little clearer and catch any future misuses of this function.

src/glob.rs
Kampfkarren: str::deserialize or `<&str as Deserialize>::deserialize` probably work too then if you still want to try it ☺️

LPGhatguy: It ended up being a little weirder, but worked out!

```rust
let glob = <&str as Deserialize>::deserialize(deserializer)?;
Glob::new(glob).map_err(D::Error::custom)
```

Ended up having to `use serde::de::Error as _` as well to bring the `custom` method into scope.

Kampfkarren: I'm not confident one way or the other, other than to say I do things like this without issue.

Issue #247, opened 2019-09-19
LSP-like protocol


Kampfkarren:
This was briefly talked about on Discord, but it'd help for projects like roblox-lua-autofills if they could connect to the Rojo server and use it to receive useful information similar to an LSP.

Currently, roblox-lua-autofills attempts to recreate the entire file structure based on projects itself, and will sometimes make mistakes for multiple project files since it's not possible for it to figure out which project you're working on. However, Rojo always knows what files it has knowledge of and where, making these problems obsolete.

I'm not sure what in specific is needed, but it is a neat idea.

LPGhatguy:
We should be able to cover this by expanding and stabilizing the Rojo HTTP API!

Dekkonot:
Sourcemaps resolve this

PR #275, opened 2019-12-10
Implement sourcemaps

LPGhatguy:
This PR adds the `--sourcemap` argument to `rojo build`, enabling Rojo to output a map from each instance's full name back to the file that created it.

This can be expanded in the future to expose the same or similar sourcemaps through the Rojo web API, fulfilling #247 at least partially!

## TODO
- [ ] Tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -393,11 +393,6 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


LPGhatguy:
Will not be pursuing this feature for now.

PR #282, opened 2019-12-19
Two way sync V0

LPGhatguy:
The goal of this PR is to land an opt-in version of two-way sync for scripts behind a flag.

A lot of the foundational work for this is already done, but this is going to be tough, and feedback is welcome from anyone!

### DIFF
--- a/src/web/interface.rs
+++ b/src/web/interface.rs
@@ -100,6 +100,23 @@ pub struct ReadResponse<'a> {
     pub instances: HashMap<RbxId, Instance<'a>>,
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct WriteRequest {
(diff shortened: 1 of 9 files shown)


src/web/interface.rs
```diff
@@ -100,6 +100,23 @@ pub struct ReadResponse<'a> {
...
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct WriteRequest {
+    pub session_id: SessionId,
+    pub removed: Vec<RbxId>,
+
+    #[serde(default)]
+    pub added: HashMap<RbxId, ()>,
```
Kampfkarren: Why this and not HashSet?

LPGhatguy: The `()` is just a placeholder type since most of this struct is unused anyways. Good catch though!

PR #297, opened 2020-02-19
VFS in external crate

LPGhatguy:
This PR rips out a lot of the complexity currently in the VFS and replaces it with a crate that doesn't do nearly as much.

This means that our performance should be a bit worse in theory since we're reading files more often, but I think this is a more solid design that we can layer caching and memoization onto.

The crate needs a name but is a solid foundation.

## TODO
- [ ] Name for VFS crate
- [x] Rojo tests working again
- [x] Snapshot system for VFS testing
- [x] In-memory backend for VFS


LPGhatguy:
Let's do it!

PR #304, opened 2020-03-17
Install plugin from CLI

### LINKED ISSUE #200: Install plugin from CLI
Rojo should be able to install the plugin from the CLI. This would make #169 easier since there wouldn't need to be any special procedure for installing the built in plugin side (maybe even the whole plugin can be built in?). 

The weird part is how to actually accomplish this. There's many possibilities but none are that great. The ideas I can think of off the top of my head are:

- Downloading from online. This is the easiest to do but has complications with versions and local development. 

- Bundling the plugin somehow, like with `include_str!` or some build script mumbo jumbo. This has issued with `cargo install`, but maybe Rojo should release its own binaries anyway, asking people to install Rust to use Rojo is a bit clunky. The docs denounce other syncing tools for having to have the user install their language's environment: "I think that this tool needs to be built in a compiled language without a runtime, for easy distribution and good performance.".

jeparlefrancais:
Closes #200 

I added a new subcommand to manage rojo's plugin.

```
rojo plugin install
rojo plugin uninstall
```

It uses roblox_install to locate the plugins folder, and inserts a `rojo-$CARGO_PKG_VERSION.rbxmx` file.

When packing the plugin files, I filter files ending with `.spec.lua` to reduce the size of the plugin (and the cli itself too since it's included inside of it)

Let me know what you think of that!


src/cli/plugin.rs
```diff
@@ -0,0 +1,116 @@
+use crate::serve_session::ServeSession;
+
+use memofs::InMemoryFs;
+use memofs::{Vfs, VfsSnapshot};
```
LPGhatguy:
```suggestion
use memofs::{InMemoryFs, Vfs, VfsSnapshot};
```

We can merge these two lines of imports.

src/cli/plugin.rs
```diff
@@ -0,0 +1,116 @@
...
+use roblox_install::RobloxStudio;
+use snafu::{ResultExt, Snafu};
+use std::{
+    fs::{self, File},
+    io::{self, BufWriter},
+};
```
LPGhatguy: Can you move the `std` imports to the top of the file and the `crate` imports to below the imports from other crates?

Generally, I like to have imports in three groups:
1. std
2. external crates
3. my own crate

Within those import groups, I let rustfmt sort them.

src/cli/plugin.rs
```diff
@@ -0,0 +1,116 @@
...
+use structopt::StructOpt;
+
+static PLUGIN_FILE_NAME: &str = concat!("rojo-", env!("CARGO_PKG_VERSION"), ".rbxmx");
```
LPGhatguy: When a user upgrades their Rojo version, I think this will result in duplicate copies of the plugin getting loaded.

I'm not sure what a good solution to this is. I think it might be a good idea to install the plugin to the same path every time for now, and give it a name that's specific to the version the binary installs, like `RojoManagedPlugin.rbxm`.

src/cli/plugin.rs
```diff
@@ -0,0 +1,116 @@
...
+#[derive(Debug, StructOpt)]
+enum PluginSubcommand {
+    /// Install the plugin in Roblox plugins folder. If the plugin is already installed, installing
+    /// it again will overwrite the current plugin file.
+    Install,
+    /// Removes the plugin if installed.
+    Uninstall,
+}
+
+/// Install rojo's plugin.
```
LPGhatguy:
```suggestion
/// Install Rojo's plugin.
```

src/cli/plugin.rs
```diff
@@ -0,0 +1,116 @@
...
+pub struct PluginError(Error);
+
+#[derive(Debug, Snafu)]
+enum Error {
+    #[snafu(display("Could not locate Roblox Studio: {}", source.to_string()))]
+    CannotLocateRobloxStudio { source: roblox_install::Error },
+
+    #[snafu(display("{}", source))]
+    Io { source: io::Error },
+}
```
LPGhatguy: I recently refactored errors in master to switch to `anyhow` for CLI commands. I think it made managing errors like this a lot easier.

src/cli/plugin.rs
```diff
@@ -0,0 +1,116 @@
...
+    /// Removes the plugin if installed.
+    Uninstall,
+}
+
+/// Install rojo's plugin.
+#[derive(Debug, StructOpt)]
+pub struct PluginCommand {
+    #[structopt(subcommand)]
+    subcommand: PluginSubcommand,
+}
```
LPGhatguy: Can we move all of the `StructOpt` code into `src/cli/mod.rs`? I think it's nice to keep all of it contained in one spot.

src/cli/plugin.rs
```diff
@@ -0,0 +1,116 @@
...
+fn xml_encode_config() -> rbx_xml::EncodeOptions {
+    rbx_xml::EncodeOptions::new().property_behavior(rbx_xml::EncodePropertyBehavior::WriteUnknown)
+}
+
+pub fn plugin(options: PluginCommand) -> Result<(), PluginError> {
+    match options.subcommand {
+        PluginSubcommand::Install => install_plugin()?,
+        PluginSubcommand::Uninstall => uninstall_plugin()?,
+    }
+    Ok(())
```
LPGhatguy: Because both of these functions return `PluginError`, I think we can simplify this block to not need `?`:

```suggestion
    match options.subcommand {
        PluginSubcommand::Install => install_plugin(),
        PluginSubcommand::Uninstall => uninstall_plugin(),
    }
```

src/cli/plugin.rs
```diff
@@ -0,0 +1,116 @@
...
+}
+
+pub fn install_plugin() -> Result<(), PluginError> {
+    static PLUGIN_BINCODE: &[u8] = include_bytes!(concat!(env!("OUT_DIR"), "/plugin.bincode"));
+
+    let plugin_snapshot: VfsSnapshot = bincode::deserialize(PLUGIN_BINCODE)
+        .expect("rojo's plugin was not properly packed into rojo's binary");
```
LPGhatguy:
```suggestion
        .expect("Rojo's plugin was not properly packed into Rojo's binary");
```

build.rs
```diff
@@ -0,0 +1,85 @@
...
+            }),
+            "rbx-dom" => VfsSnapshot::dir(hashmap! {
+                "rbx_dom_lua" => VfsSnapshot::dir(hashmap! {
+                    "src" => snapshot_from_fs_path(&plugin_modules.join("rbx-dom").join("rbx_dom_lua").join("src")).unwrap()
+                })
+            }),
+        }),
+    });
+
+    println!("cargo:rerun-if-changed=build.rs");
```
LPGhatguy: Is this line required? I'm under the impression that Cargo should always rebuild whenever `build.rs` changes.

build.rs
```diff
@@ -0,0 +1,85 @@
...
+use maplit::hashmap;
+use memofs::VfsSnapshot;
+
+fn snapshot_from_fs_path(path: &PathBuf) -> Result<VfsSnapshot, PathBuf> {
+    if path.is_dir() {
+        let entries: Result<Vec<fs::DirEntry>, PathBuf> = fs::read_dir(path)
+            .map_err(|_| path.to_owned())?
+            .map(|entry| entry.map_err(|_| path.to_owned()))
+            .into_iter()
+            .collect();
```
LPGhatguy: We should be able to keep these items as an iterator instead of collecting them. If you want better error messages with paths, [fs-err](https://docs.rs/fs-err/2.3.0/fs_err/) is a crate we recently added a dependency on.

build.rs
```diff
@@ -0,0 +1,85 @@
...
+                    .file_name()
+                    .to_str()
+                    .ok_or(path.to_owned())
+                    .and_then(|file_name| {
+                        snapshot_from_fs_path(&path)
+                            .map(|snapshot| (file_name.to_owned(), snapshot))
+                    })
+            })
+            .into_iter()
+            .collect();
```
LPGhatguy: Two notes here:

1. The result of `map` is already an `Iterator`, so we don't need to call `into_iter` on it.
2. `VfsSnapshot::dir` accepts an iterator as an argument, so there's no need to collect here.

build.rs
```diff
@@ -0,0 +1,85 @@
...
+            .map_err(|_| path.to_owned())?
+            .map(|entry| entry.map_err(|_| path.to_owned()))
+            .into_iter()
+            .collect();
+
+        let vfs_entries: Result<Vec<(String, VfsSnapshot)>, PathBuf> = entries?
+            .iter()
+            .filter(|entry| {
+                let file_name = entry.file_name();
+                !file_name.to_string_lossy().ends_with(".spec.lua")
```
LPGhatguy: Since `to_string_lossy` can allocate and an invalid Unicode file will never end with `.spec.lua` how we want anyways, I like to write this as

```rust
                !file_name.to_str().map(|name| name.ends_with(".spec.lua")).unwrap_or(false)
```

We can also likely merge the line above us into this statement since we don't use `file_name` more than once.

build.rs
```diff
@@ -0,0 +1,85 @@
+use std::env;
+use std::fs;
+use std::path::{Path, PathBuf};
+
+use maplit::hashmap;
+use memofs::VfsSnapshot;
+
+fn snapshot_from_fs_path(path: &PathBuf) -> Result<VfsSnapshot, PathBuf> {
```
LPGhatguy: `PathBuf` is a strange error type! We should use `io::Error` (by returning `io::Result<VfsSnapshot>`) I imagine.

build.rs
```diff
@@ -0,0 +1,85 @@
...
+            "t" => VfsSnapshot::dir(hashmap! {
+                "lib" => snapshot_from_fs_path(&plugin_modules.join("t").join("lib")).unwrap()
+            }),
+            "rbx-dom" => VfsSnapshot::dir(hashmap! {
+                "rbx_dom_lua" => VfsSnapshot::dir(hashmap! {
+                    "src" => snapshot_from_fs_path(&plugin_modules.join("rbx-dom").join("rbx_dom_lua").join("src")).unwrap()
+                })
+            }),
+        }),
+    });
```
LPGhatguy: We talked in Discord about this block, and I've been thinking about it for awhile. I think it's okay to keep around for now, and we can revisit this problem later.

src/cli/plugin.rs
```diff
@@ -0,0 +1,116 @@
...
+    let vfs = Vfs::new(in_memory_fs);
+
+    let session = ServeSession::new(vfs, "plugin");
+
+    let tree = session.tree();
+
+    log::trace!(
+        "Writing plugin {} in {}",
+        PLUGIN_FILE_NAME,
+        plugins_folder_path.to_string_lossy()
```
LPGhatguy: For displaying paths, there's a dedicated `display` method that handles edge cases correctly and (probably) doesn't allocate as often:

```suggestion
        plugins_folder_path.display()
```

src/cli/plugin.rs
```diff
@@ -0,0 +1,116 @@
...
+    );
+    let file = File::create(plugins_folder_path.join(PLUGIN_FILE_NAME)).context(Io)?;
+
+    let mut file = BufWriter::new(file);
+
+    let root_id = tree.get_root_id();
+
+    rbx_xml::to_writer(&mut file, tree.inner(), &[root_id], xml_encode_config()).unwrap();
```
LPGhatguy: We can probably serialize the plugin with `rbx_binary`. It'll load a lot faster, and we've been releasing the plugin using the binary model format for awhile now.

src/cli/plugin.rs
jeparlefrancais: Yep I hesitated a lot with this too, I chose that in case some people would have different versions of rojo installed... like if they were trying an alpha release but wanted to keep their other rojo at the same time. I don't know if it's really a concern though.

It's probably more practical for the majority of useres to have a single name like you suggested.

build.rs
jeparlefrancais: I kept reading the sentence again and again, but I find it really confusing haha! Maybe you could help with that?

It's the third paragraph [here](https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorerun-if-changedpath).
> If the build script inherently does not need to re-run under any circumstance, then emitting `cargo:rerun-if-changed=build.rs` is a simple way to prevent it from being re-run. Cargo automatically handles whether or not the script itself needs to be recompiled, and of course the script will be re-run after it has been recompiled. Otherwise, specifying `build.rs` is redundant and unnecessary.

What I think I understand is that it, when printing it, it builds only when the file itself changes, instead of always?

Issue #205, opened 2019-06-28
Abuse game:GetObjects for live-sync


LPGhatguy:
This would solve #161 and be an alternative to #169.

There's a great API that loads models out of the content folder I didn't know about!

```lua
game:GetObjects("rbxasset://models/AvatarContextMenu/AvatarContextArrow.rbxm").Parent = workspace
```

Just like what we wanted to solve with a RobloxScript security API in #169, we should be able to use this to load in models that Rojo isn't able to live-sync. I'm not sure how we should try to detect models that fall into this scenario, and some exotic service properties like `Lighting.Technology` are still in a rough spot.

Kampfkarren:
Why not just always use this for rbxm/rbxmx files?

LPGhatguy:
Performance, and support for stuff not managed by Rojo. If we always save to a model and load from a model, delete the existing model, and replace it, changes will be super noticeable.

So while we _could_ use `GetObjects` for models or for everything, I think we can do a better job with individual property updates instead.

DBReinitialized:
You could also abuse the fact that Studio can run custom CoreScripts and utilize the VSCode extension to inject one into Studio, thus granting access to the RobloxScriptSecurity tagged APIs.

Kampfkarren:
@DBReinitialized That's basically the same as #169

Kampfkarren:
> I'm not sure how we should try to detect models that fall into this scenario
rbx_dom and check scriptability?

fredgig:
Any update on this? 

Not being able to sync MeshId properties with .rbxmx files from our Rojo project to Rblx Studio is a deal breaker for us, and we're limited to only syncing scripts. Are there currently any other ways to handle MeshParts?

LPGhatguy:
I recommend using fully-managed Rojo to work with MeshParts right now, which is currently the only way to have Rojo manage lots of things. This will likely mostly work for people to live sync things that they're already using `rojo build` for!

Kampfkarren:
> This will likely mostly work for people to live sync things that they're already using rojo build for!

Can confirm, it does.

Dekkonot:
This functionality will be implemented in #1028. I'm leaving this open for now, but if #1028 closes, this will close too.

PR #332, opened 2020-06-21
Break apart plugin reconciler

LPGhatguy:
The current reconciler is monolithic and hard to test. This PR breaks apart Rojo's reconciler into smaller pieces that are easier to test in isolation.

## Design
On initial sync, the Rojo plugin goes through this process:
* Grab the entire virtual DOM from the server
* `hydrate` the real DOM using IDs from the virtual DOM
* `diff` the real DOM and the virtual DOM, generating a patch
* `applyPatch` the patch to actually make these changes

Every time the server notifies Rojo of a change, the Rojo plugin goes through this process:
* `applyPatch` the patch sent from the server. That's it!

Almost all of these methods are fallible. They'll usually fail due to permission errors, but might also run into unknown or unexpected data that can't be predicted beforehand.

### Hydration - `hydrate`
Hydration is the process of collecting all existing instances, matching them up to virtual instances that the server knows about, and assigning them IDs. This process populates a type known as `InstanceMap`, which is a bidirectional map between Rojo's server-generated IDs and actual instances in Roblox Studio.

Anything that doesn't have an ID at this point is not present on the server and should potentially be deleted by the plugin.

### Diffing - `diff`
This step calculates a patch that would catch up the real DOM to match the virtual DOM. It depends on the DOM being hydrated, which drastically simplifies this process. Each property is measured and compared with the virtual equivalent, unknown virtual instances are marked for addition in the patch, and any unknown real instances are marked for deletion.

By returning a patch here instead of directly applying changes, I'm hoping we can use this for two way sync. On initial sync, we'll detect that there's some drift and let the user pick which side's changes should be kept. We'll need a way to invert a patch, which is fairly difficult, but this is a powerful technique.

### Patch Application - `applyPatch`
This is the only actual side-effectful function. This function takes a patch and applies it to the real DOM.

`applyPatch` will try to apply as much of a patch as it can, and return any parts that couldn't be applied. This is different than what Rojo does on the master branch currently! The goal of this behavior is to be able to fall back to other approaches for instances that couldn't be created or updated from Lua by normal means. The big one the trick outlined in #205, but any other form of content negotiation with the server could happen here.

`applyPatch` will sometimes call into another function that isn't exposed in the top-level API, `reify`. `reify` will attempt to create an instance, set all properties on it, and create all of its descendants. If an instance can't be created or a property can't be set, `applyPatch` will do as much as it can, and return a partial patch containing what couldn't happen.

## TODO
* [x] Finish applying ref properties in `reify`
* [x] Figure out long string issue
    * I pushed a change to try to spot-fix a bug I saw when manually testing, but I'm not sure it's right. `diff` should not pick up properties that didn't actually change, but it might be.


plugin/src/Reconciler/diff.spec.lua
```diff
@@ -0,0 +1,265 @@
...
+		local ok, patch = diff(knownInstances, virtualInstances, "ROOT")
+
+		assert(ok, tostring(patch))
+
+		assert(isEmpty(patch.updated))
+		assert(isEmpty(patch.removed))
+		expect(size(patch.added)).to.equal(1)
+		expect(patch.added["CHILD"]).to.equal(virtualInstances["CHILD"])
+	end)
+end
```
magnalite, line 265: This needs a test case to ensure the diff scans the entire virtual DOM/instance map. Currently this only tests the case where the root has an immediate child but doesn't check if changes to that child (and it's children) get included in the diff.

magnalite, line 265: An alternative test case where you have a cyclic reference may also be good. I doubt this can come up during normal use however it would be nice to ensure it doesn't blow up rojo in the event it does somehow happen.

plugin/src/Reconciler/hydrate.lua
```diff
@@ -0,0 +1,50 @@
...
+
+	-- For each existing child, we'll track whether it's been paired with an
+	-- instance that the Rojo server knows about.
+	local isExistingChildVisited = {}
+	for i = 1, #existingChildren do
+		isExistingChildVisited[i] = false
+	end
+
+	for _, childId in ipairs(virtualInstance.Children) do
+		local apiChild = virtualInstances[childId]
```
magnalite, line 27: Why is this called apiChild as opposed to something like virtualChild?

plugin/src/Reconciler/init.lua
```diff
@@ -0,0 +1,34 @@
...
+
+	return setmetatable(self, Reconciler)
+end
+
+function Reconciler:applyPatch(patch)
+	return applyPatch(self.__instanceMap, patch)
+end
+
+function Reconciler:hydrate(virtualInstances, rootId, rootInstance)
+	return hydrate(self.__instanceMap)
```
magnalite, line 27: Why are `virtualInstances`, `rootId` and `rootInstance` not passed to hydrate here?

plugin/src/Reconciler/setProperty.lua
```diff
@@ -0,0 +1,48 @@
...
+		return false, Error.new(Error.OtherPropertyError, {
+			className = instance.ClassName,
+			propertyName = propertyName,
+		})
+	end
+
+	return true
+end
+
+return setProperty
```
magnalite, line 48: `getProperty` and `setProperty` both share a lot of functionality, is there a way we could have more code reuse between them? I'm not sure if it's the best way but I tackled a similar problem by making a function like `propertyOperation("Write", instance, propertyName, value)`. What do you think about this kind of approach?

magnalite (commented):
This seems very cool! I left a few comments on things I'm not sure about.

plugin/src/Reconciler/setProperty.lua
LPGhatguy: I think the similarities are pretty much just the error handling boilerplate. Maybe cutting down on that would make these functions feel less similar?

plugin/src/Reconciler/init.lua
LPGhatguy: Good catch!

plugin/src/Reconciler/hydrate.lua
LPGhatguy: At one point, these objects were called ApiInstances instead of VirtualInstances. Looks like something I forgot to update :D

plugin/src/Reconciler/diff.lua
```diff
@@ -0,0 +1,137 @@
...
+				if not ok then
+					return false, err
+				end
+			end
+		end
+
+		for _, childId in ipairs(virtualInstance.Children) do
+			local childInstance = instanceMap.fromIds[childId]
+
+			if childInstance == nil then
```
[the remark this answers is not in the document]
LPGhatguy: If `childInstance` would be non-nil here, then we've handled it in the loop above this one. I'll add a couple comments to clarify both this check and both the loops overall.

plugin/src/Reconciler/diff.spec.lua
LPGhatguy: I'm not too worried about this case coming up. I think we can avoid adding the extra machinery to detect cycles unless it's something that could feasibly happen.

LPGhatguy:
Addressed and replied to most of the review feedback. I'll add more test cases tomorrow and see if I can get a baseline `applyPatch` implementation working tomorrow, where most of the fun code lives. Then we can hook everything back together in the plugin and start trying to find funny edge cases.

LPGhatguy:
Let's go forward with this. I think this is stable enough to promote to master, and it's a pretty big improvement in many ways.

PR #343, opened 2020-07-17
Add selene.toml file creation to init.rs

### LINKED ISSUE #342: Create selene.toml for Roblox projects when rojo init is run
Reasoning behind this: we already ``git init`` and create a repository, so I don't see why we shouldn't extend this utility to selene as the command is already not tooling agnostic.

Pain Point: I have to create a new selene.toml every time I ``rojo init``, and due to how the selene VSCode extension works (I'm going to make a PR to fix this on their side) I have to restart VSCode after that. It's annoying and I'm sure a lot of people who use selene have encountered the same issue.

I'll make a PR for this, but I'll link it back to here if anyone has any points to bring up about if this is the approach that would be most effective.

unix-system:
closes #342

### DIFF
--- a/assets/default-model-project/selene.toml
+++ b/assets/default-model-project/selene.toml
@@ -1,0 +1,3 @@
+# Selene is a linter for Lua that has support for Roblox Lua, when this file is present
+# Find out more at https://github.com/Kampfkarren/selene
+std = "roblox"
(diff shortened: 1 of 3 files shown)


unix-system:
I'm going to need to sign the CLA. Let's figure out how this works!

unix-system:
This PR is cursed, I'm going to close it and retry it...

PR #345, opened 2020-07-17
Add --no-git flag to rojo init

Jaguar-515:
As Kampfkarren said in #342, there should be an option to not initialize Git in `rojo init`.

This PR implements `--no-git` as an optional flag since there are some users who don't use Git as their VCS and use a different VCS or just don't want to use Git at all.

### DIFF
--- a/src/cli/init.rs
+++ b/src/cli/init.rs
@@ -51,7 +51,11 @@ pub fn init(options: InitCommand) -> Result<(), anyhow::Error> {
     }
 }
 
-fn init_place(base_path: &Path, project_params: ProjectParams, no_git: bool) -> Result<(), anyhow::Error> {
+fn init_place(
+    base_path: &Path,
+    project_params: ProjectParams,
+    no_git: bool,
+) -> Result<(), anyhow::Error> {
     eprintln!("Creating new place project '{}'", project_params.name);
 
     let project_file = project_params.render_template(PLACE_PROJECT);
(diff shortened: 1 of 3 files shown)


LPGhatguy:
Is this coming out of a specific need? Are there people initializing Rojo projects that do not want Git repositories today?

My understanding of the result of that conversation is that we don't need to change anything. We can wait and see if there is interest in supporting another VCS or if there's value in supporting _no_ VCS.

If we were going to make this change, I think it would be better to support a `--vcs` argument, defaulting to Git, but having an option for none. Something like:

```bash
# same as the default
rojo init --vcs git

# initialize no VCS at all
rojo init --vcs none

# a possible future?
rojo init --vcs pijul
```

That said, I want to be clear: I don't think we need to make a change like this right now!

src/cli/init.rs
```diff
@@ -43,13 +43,19 @@ pub fn init(options: InitCommand) -> Result<(), anyhow::Error> {
...
+        InitKind::Place => init_place(&base_path, project_params, no_git),
+        InitKind::Model => init_model(&base_path, project_params, no_git),
     }
 }
 
-fn init_place(base_path: &Path, project_params: ProjectParams) -> Result<(), anyhow::Error> {
+fn init_place(
+    base_path: &Path,
+    project_params: ProjectParams,
+    no_git: bool,
```
Kampfkarren, line 57: Using a bool parameter is going to get messy if we put more in. This should probably be its own struct of initialization options, with just `no_git: bool` in it for now.

Jaguar-515:
@LPGhatguy You actually do have a good point! I'll make another PR around the weekend that is based on your feedback on this PR.

PR #367, opened 2020-11-20
New UI

Reselim:
This aims to re-implement the Rojo UI with a somewhat new design, hopefully in time for the release of 6.0.0


Reselim:
Ready for review! It's a lot of code to go through, so good luck.

Reselim:
Oh, by the way: `sessionName` should be provided to `ServeSession:onStatusChanged`'s `details` argument for it to display in the UI.

LPGhatguy (commented):
This is a very pretty UI, and I'm very excited!

Can we rename the newly-introduced folders to `PascalCase`? The style I'm aiming for is to make them be Proper Things.

I'll do a more thorough review tomorrow, I've been bouncing between lots of things today.

.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+	"Lua.diagnostics.disable": [
+		"undefined-global",
+		"incorrect-call-arguments"
+	]
+}
```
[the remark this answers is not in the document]
Reselim: Whoops, totally didn't see that this was checked in. Meant to disable that globally, actually.

plugin/src/App/statusPages/Error.lua
```diff
@@ -0,0 +1,155 @@
...
+			PaddingBottom = UDim.new(0, 14),
+		}),
+	})
+end
+
+function ErrorPage.getDerivedStateFromProps(props)
+	return {
+		errorMessage = props.errorMessage,
+	}
+end
```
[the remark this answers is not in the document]
Reselim: This is so when/if `errorMessage` gets removed from its' props, it still remains as state so it can be animated out properly.

plugin/src/App/components/Throbber.lua
```diff
@@ -0,0 +1,65 @@
...
+
+local Roact = require(Rojo.Roact)
+
+local Theme = require(Plugin.App.Theme)
+
+local ROTATIONS_PER_SECOND = 1.75
+
+local e = Roact.createElement
+
+local Throbber = Roact.PureComponent:extend("Throbber")
```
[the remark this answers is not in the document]
Reselim: [Throbber](https://en.wikipedia.org/wiki/Throbber) is an accurate term, but spinner is far simpler for sure. Will update.

plugin/src/App/components/Throbber.lua
```diff
@@ -0,0 +1,65 @@
...
+
+function Throbber:init()
+	self.rotation, self.setRotation = Roact.createBinding(0)
+end
+
+function Throbber:render()
+	return Theme.with(function(theme)
+		return e("ImageLabel", {
+			Image = "rbxassetid://3222730627",
```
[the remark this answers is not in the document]
Reselim: Whoops -- I actually ported this code directly from Inklii. Thanks for catching that.

plugin/src/App/statusPages/Error.lua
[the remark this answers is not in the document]
LPGhatguy: Oh, that's funky! I hadn't thought about how this function would not update state if `props.errorMessage` is nil. Can you leave a comment describing that behavior? Will `errorMessage` _be_ removed and need to be animated out?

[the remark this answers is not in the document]
Reselim: At the moment, it's not necessary because `errorMessage` stays in `App` past when it navigates away. It could very much cause a difficult bug in the future though.

By the way, this stuff is also in the `Connected `component.

plugin/src/App/StatusPages/Connected.lua
```diff
@@ -0,0 +1,124 @@
...
+		Padding = e("UIPadding", {
+			PaddingLeft = UDim.new(0, 20),
+			PaddingRight = UDim.new(0, 20),
+		}),
+	})
+end
+
+function ConnectedPage.getDerivedStateFromProps(props)
+	-- If projectName or address ever get removed from props, make sure we still have the property!
+	-- This is because the component needs to be rendered for it to accurately render out.
```
LPGhatguy, line 116: What does this mean?

Reselim: The wording is pretty bad. I'll fix it.

plugin/src/App/StatusPages/Error.lua
```diff
@@ -0,0 +1,158 @@
...
+				local containerSize = object.AbsoluteSize - ERROR_PADDING * 2
+
+				local textBounds = TextService:GetTextSize(
+					self.props.errorMessage, 16, Enum.Font.Code,
+					Vector2.new(containerSize.X, math.huge)
+				)
+
+				self.setContentSize(Vector2.new(containerSize.X, textBounds.Y))
+			end,
+		}),
```
LPGhatguy, line 55: This is weird! Why not just pass through `[Roact.Change.AbsoluteSize]` in `ScrollingFrame`?

plugin/src/App/StatusPages/NotConnected.lua
```diff
@@ -0,0 +1,156 @@
...
+
+				BackgroundTransparency = 1,
+
+				[Roact.Ref] = props.portRef,
+
+				[Roact.Change.Text] = function(object)
+					local text = object.Text
+
+					text = text:gsub("%D", "")
+					text = text:sub(1, 5)
```
LPGhatguy, line 66: Even though ports don't go over 2¹⁶, truncating text like this is usually frustrating UX, and we should avoid doing it!

plugin/src/App/StatusPages/Settings.lua
```diff
@@ -0,0 +1,230 @@
...
+				}, {
+					Name = e("TextLabel", {
+						Text = self.props.name,
+						Font = Enum.Font.GothamBold,
+						TextSize = 17,
+						TextColor3 = theme.Setting.NameColor,
+						TextXAlignment = Enum.TextXAlignment.Left,
+						TextTransparency = self.props.transparency,
+
+						Size = UDim2.new(1, 0,  0, 17),
```
LPGhatguy, line 111:
```suggestion
						Size = UDim2.new(1, 0, 0, 17),
```

plugin/src/App/StatusPages/Error.lua
Reselim: Roact creates a new PropMarker every time you index `Change` or `Event`, which doesn't allow them to be passed down easily.

Reselim: Ah, my bad. Was having issues and thought Roact created a new PropMarker every time it was indexed.

plugin/src/App/Components/Studio/StudioPluginGui.lua
```diff
@@ -0,0 +1,68 @@
...
+					if self.props.onClose then
+						self.props.onClose()
+					else
+						self.pluginGui.Enabled = false
+					end
+				end)
+			else
+				-- Make sure the initial state is preserved until something changes
+				self.pluginGui.Enabled = self.props.active
+			end
```
LPGhatguy, line 51: Can this happen in `init` instead of conditionally in `render`? We should be able to break this into two components, plus maybe a little function glue component to make this work out.

plugin/src/App/Components/Studio/StudioToggleButton.lua
```diff
@@ -0,0 +1,48 @@
...
+				)
+
+				self.button.Click:Connect(function()
+					if self.props.onClick then
+						self.props.onClick()
+					end
+				end)
+
+				self.button.ClickableWhenViewportHidden = true
+			end
```
LPGhatguy, line 34: Same thing here, I think this sort of logic belongs in `init`.

plugin/src/App/Components/Studio/StudioPluginGui.lua
Reselim: Alright, will do.

plugin/src/App/StatusPages/NotConnected.lua
Reselim: The UI isn't really designed to support more than 5 characters. Are you fine with it just overflowing?

LPGhatguy (approved):

LPGhatguy (approved):

PR #401, opened 2021-02-26
Use GetObjects as fallback for instances that can't be created normally

### LINKED ISSUE #205: Abuse game:GetObjects for live-sync

Kampfkarren:
Closes #205.

TODO:
- [x] Everything labelled TODO 😄 
- [x] Track and fix bug where sometimes children are duplicated - **I can't actually seem to reproduce this? I think I just did something wrong locally one time...**
- [x] Package DogeUpdate rbxm used in tests, and parent to content folder automatically
- [x] Remove isDebugBuild edit
- [x] Track and fix "Connect" button just, not showing anything? - **It's a Flipper bug.**
- [x] Test metadata
- [x] Fix crash when updating MeshId in *.project.json.\*
- [ ] Possibly some sort of GC for all the assets being created in filesystem.

\* - This is because the serve is trying to request an update with the service its under. It tries to insert the service as a child, and then crashes. GetObjects should not be trying to create new instances when its just metadta that is different.


plugin/src/tryGetObjects.lua
```diff
@@ -0,0 +1,88 @@
...
+		table.insert(assetsToRequest, update.id)
+		table.insert(unappliedPatch.updated, update)
+
+		receiveCallbacks[update.id] = function(newInstance)
+			table.remove(unappliedPatch.updated, table.find(unappliedPatch.updated, update))
+
+			local oldInstance = instanceMap.fromIds[update.id]
+
+			-- TODO: What if oldInstance is nil?
+			-- TODO: This will dupe children if the rbxm is not empty.
```
Kampfkarren, line 35: I don't think this TODO is actually accurate, the assets being created don't have any children. The bug is from something else 🤷

plugin/src/Config.lua
```diff
@@ -1,6 +1,7 @@
 local strict = require(script.Parent.strict)
 
-local isDevBuild = script.Parent.Parent:FindFirstChild("ROJO_DEV_BUILD") ~= nil
+-- local isDevBuild = script.Parent.Parent:FindFirstChild("ROJO_DEV_BUILD") ~= nil
+local isDevBuild = true
```
LPGhatguy, line 4: I recommend testing in a place where you've run `_G.ROJO_DEV_CREATE()` instead of hard-coding this constant one way or the other.

plugin/src/tryGetObjects.lua
```diff
@@ -0,0 +1,88 @@
...
+local Promise = require(script.Parent.Parent.Promise)
+
+local function tryGetObjects(instanceMap, apiContext, patch)
+	assert(PatchSet.validate(patch))
+
+	local unappliedPatch = PatchSet.newEmpty()
+
+	-- GetObjects won't help with anything that failed to remove
+	unappliedPatch.removed = patch.removed
+	-- TODO: Implement this? Do we have to?
```
LPGhatguy, line 13: Yes! This is important for if a file is newly added on the filesystem

plugin/src/tryGetObjects.lua
```diff
@@ -0,0 +1,88 @@
...
+			Assuming we requested the following IDs:
+			- DOGE: A doge mesh named Doge
+			- ROCK: A rock mesh named MyPet
+
+			Root: Folder
+			* DOGE: Folder
+				* Doge: MeshPart
+			* ROCK: Folder
+				* MyPet: MeshPart
+		]]
```
LPGhatguy, line 63: Instead of creating objects inside a folder, create a model with multiple root objects. The items don't need special names -- they can just be in the same order the plugin requested them in!

plugin/src/tryGetObjects.spec.lua
```diff
@@ -0,0 +1,39 @@
+-- TODO: Have some documentation/script in order to force the rbxassets to be there
```
LPGhatguy, line 1: Does the API work with `rbxassetid://` URLs? What about a model that we know exists in the Studio distribution? We should rely on those if we can.

src/web/api.rs
```diff
@@ -279,6 +285,112 @@ impl ApiService {
...
         })
     }
+
+    fn handle_api_create_assets(&self, request: Request<Body>) -> <Self as Service>::Future {
+        let session_id = self.serve_session.session_id();
+
+        let serve_tree = self.serve_session.tree_handle();
+        let vfs = self.serve_session.vfs_handle();
+
+        Box::new(request.into_body().concat2().and_then(move |body| {
```
LPGhatguy, line 295: Damn. We've gotta refactor this code to start using async/await someday, lol.

src/web/api.rs
```diff
@@ -279,6 +285,112 @@ impl ApiService {
...
+            if request.session_id != session_id {
+                return json(
+                    ErrorResponse::bad_request("Wrong session ID"),
+                    StatusCode::BAD_REQUEST,
+                );
+            }
+
+            let serve_dom = serve_tree.inner();
+
+            let packed_uuid = Uuid::new_v4().to_simple().to_string();
```
LPGhatguy, line 317: This implements `Display`, we don't need to `to_string` on it before using it with `format!` below.

src/web/api.rs
```diff
@@ -279,6 +285,112 @@ impl ApiService {
...
+                        ErrorResponse::bad_request(format!(
+                            "Couldn't find Roblox Studio: {}",
+                            error
+                        )),
+                        StatusCode::INTERNAL_SERVER_ERROR,
+                    );
+                }
+            };
+
+            // TODO: memofs does not have create_dir_all, but it is necessary here.
```
LPGhatguy, line 333: We don't really want to use memofs for stuff outside the file tree that we want cached, anyways.

src/web/api.rs
```diff
@@ -279,6 +285,112 @@ impl ApiService {
...
+                &export_tree,
+                &vec![export_tree.root_ref()],
+            ) {
+                return json(
+                    ErrorResponse::bad_request(format!("Couldn't write DOM to buffer: {}", error)),
+                    StatusCode::INTERNAL_SERVER_ERROR,
+                );
+            }
+
+            if let Err(error) = vfs.write(packed_path, writer) {
```
LPGhatguy, line 382: This shouldn't go through the VFS; we should use regular filesystem methods here.

test-projects/script-unwritable-instances/default.project.json
```diff
@@ -0,0 +1,11 @@
...
+	"name": "script-unwritable-instances",
+	"tree": {
+		"$className": "DataModel",
+
+    "ReplicatedStorage": {
+      "$className": "ReplicatedStorage",
+      "$path": "src"
+    }
+	}
+}
```
LPGhatguy, line 11: Indentation is kinda all over on this one

plugin/src/tryGetObjects.lua
Kampfkarren: Figured, just couldn't figure it out because I was using existing projects, and wasn't creating new files.

plugin/src/tryGetObjects.lua
Kampfkarren: What would the roots in this example be? `DOGE` and `ROCK` or `Doge` and `MyPet`?

With the latter, is it possible for a .rbxm/.rbxmx that's being imported by Rojo to be multi-rooted?

plugin/src/tryGetObjects.spec.lua
Kampfkarren: Looks like it does!

src/web/api.rs
Kampfkarren: The errors I got when writing this were so confusing, please :cry:

src/web/api.rs
Kampfkarren: I thought it was for testability, so just use the filesystem directly?

test-projects/script-unwritable-instances/default.project.json
Kampfkarren: It's because it's both tabs and spaces, and GitHub has a hillariously wide tab width.

I'll fix it by making it whatever the standard is in the other project.json files.

plugin/src/tryGetObjects.spec.lua
Kampfkarren: As for using something in the Studio distribution, it's possible if we use a multi-rooted result instead of the current one.

plugin/src/tryGetObjects.spec.lua
```diff
@@ -0,0 +1,39 @@
...
+return function()
+	local InstanceMap = require(script.Parent.InstanceMap)
+	local PatchSet = require(script.Parent.PatchSet)
+	local Promise = require(script.Parent.Parent.Promise)
+	local tryGetObjects = require(script.Parent.tryGetObjects)
+
+	local MESH_DOGE = "rbxassetid://4574885352"
+
+	itFOCUS("should apply updates to existing instances", function()
```
Kampfkarren, line 10: Need to remove this itFOCUS.

Kampfkarren:
It's not just Connect that's not working...it's Settings too? Honestly clueless, will look further 🤷

Kampfkarren:
Oh...it's because Flipper is using PreRender now, which isn't actually enabled yet. CC @Reselim, might not want to do that yet!

Reselim:
Yeah, no worries. I realized that and I figured it would be okay to keep it as is as long as there isn't a new release... it seems Rojo is using the master branch though. I'll revert the change for now.

plugin/src/tryGetObjects.lua
Kampfkarren: I copied and pasted a file and...it worked perfectly? It was still under `updated`, not `added`.

Reselim:
Should be all good!

plugin/src/tryGetObjects.spec.lua
Kampfkarren: Though I'm a bit worried about relying on anything internal to Studio, honestly.

plugin/src/tryGetObjects.lua
Kampfkarren: Nevermind, added seems to be used when modifying .project.json.

Kampfkarren:
@LPGhatguy It appears that invalid rbxm files crash Roblox Studio, while invalid rbxmx files will simply error. I think it is better for Rojo to produce the latter for GetObjects, any thoughts?

plugin/src/tryGetObjects.lua
```diff
@@ -0,0 +1,92 @@
...
+
+		table.insert(assetsToRequest, update.id)
+		table.insert(unappliedPatch.updated, update)
+
+		receiveCallbacks[update.id] = function(newInstance)
+			table.remove(unappliedPatch.updated, table.find(unappliedPatch.updated, update))
+
+			local oldInstance = instanceMap.fromIds[update.id]
+
+			-- TODO: What if oldInstance is nil?
```
Kampfkarren, line 39: Is this possible? This is only in updated?

plugin/src/tryGetObjects.lua
```diff
@@ -0,0 +1,92 @@
...
+			for _, oldChild in ipairs(oldInstance:GetChildren()) do
+				oldChild.Parent = newInstance
+			end
+
+			local oldParent = oldInstance.Parent
+			instanceMap:destroyInstance(oldInstance)
+			newInstance.Parent = oldParent
```
Kampfkarren, line 46: This likely needs to be in a pcall.

plugin/src/tryGetObjects.lua
Kampfkarren: It's not.

Kampfkarren:
Code is all ready (except for the GC entry I added to the changelog).

I'm going to start battle testing this for my own games.

Kampfkarren:
Found the bug, when you create an instance in Studio then save to file, it duplicates. I don't think GetObjects has anything to do with this though, as nothing in this was GetObjects-worthy (maybe SourceAssetId?).


https://user-images.githubusercontent.com/3190756/109407981-40f82380-793a-11eb-9090-bec632f42853.mp4

Kampfkarren:
image

I have NO idea what is going on here. This is a MeshPart that was imported through GetObjects.

Kampfkarren:
I've been looking for a while and honestly still have no clue what's up with MeshParts being funky.

LPGhatguy:
Can you upload the model to the website and try to insert it with `InsertService`? Obviously this isn't quite a working solution for Rojo... probably, it might indicate whether there's a deeper issue or if this is just something funny about `game:GetObjects()`.

Kampfkarren:
It's already on the website, and I can import it just fine, it just imports really big (but the part size is accurate to the model).

Kampfkarren:
Looks like `game:GetObjects("rbxassetid://4716256451")[1].Parent = workspace` imports it correctly, weird!

Kampfkarren:
Looks like in this implementation, refs are not being reserved.

LPGhatguy:
Refs are going to be tricky! We'll need to use a hybrid approach to write ref properties that fall outside the bounds of the object we use `GetObjects` on.

CertifiedRice (approved):

unix-system:
So if I can understand the problem here, the Ref structure in the file format (which points to some other part of the file?) doesn't get preserved and so we lose this information. What approach can we take to solve this? 

This is the primary blocker for my team using Rojo 😔 so I'd love to help out if you'd like!

PR #403, opened 2021-03-06
change server bind address

4xMSAA:
127.0.0.1 is a loopback interface, and only works on the same host
binding to 0.0.0.0 will allow connections from other hosts and will actually give purpose to the IP field on the plugin

i have a laptop running linux (so cannot use studio) and another host running windows, where i RDP to for studio - i was wondering why it couldn't connect. rebuilt with these changes and it can now listen to the linux host

previously:
```
[Rojo-Warn] Disconnected from an error: Couldn't connect to the Rojo server.
[Rojo-Warn] Make sure the server is running — use 'rojo serve' to run it!
```
now:
image
*using stable, but change is universal*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Rojo Changelog
 
 ## Unreleased Changes
+* Added `--address` flag to `rojo serve` to allow for external connections. ([#403][pr-403])
+
+[pr-403]: https://github.com/rojo-rbx/rojo/pull/403
 
 ## [7.0.0-alpha.2][7.0.0-alpha.2] (February 19, 2021)
 * Fixed incorrect protocol version between the client and server.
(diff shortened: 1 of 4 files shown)


LPGhatguy:
As is, this change is very dangerous! Allowing access for all Rojo connections from outside the current machine opens up a lot of risk for leaking information, or worse, taking advantage of Rojo's two-way sync web API to write files onto the host machine.

Instead, I would accept a PR to make this configurable. Having support for `rojo serve --address 0.0.0.0` would let users opt into this behavior.

4xMSAA:
slight nitpick, but contribution page did not mention use of rustfmt - on the other hand, i could have probably been more careful myself

Deverium (approved):

LPGhatguy (approved):

LPGhatguy:
Thanks for putting this together!

I pushed a small handful of tweaks to the branch like using `SocketAddr`, displaying `localhost` in the console instead of the IP if the IP is loopback (`127.0.0.1` or `::1`), and updating the changelog.

PR #412, opened 2021-04-14
Reword top level model error message

### LINKED ISSUE #404: Confusing top level model error message
If you have a model with more than one top level instance the following error will pop up:
```
[ERROR] Rojo doesn't have support for model files with zero or more than one top-level instances yet.
Check the model file at path C:\blah\blah\blah.rbxm
```
This error is confusing as the wording is not great. It could be read as `zero or more top-level instances` (as I've accidently read it as multiple times) It should be reworded to make more sense and be less confusing.

Mixu78:
Closes #404


LPGhatguy (approved):
This is a good fix, thank you!

PR #413, opened 2021-04-17
Block usage of "Name" or "Parent" in $properties

### LINKED ISSUE #391: Prevent setting 'Name' or 'Parent' in $properties
Right now, it's possible to create a Rojo project with `Name` or `Parent` in the `$properties` map of a project file. This causes lots of problems when the Rojo plugin tries to apply patches for live syncing.

Mixu78:
Closes #391.

This pull makes the usage of "Name"/"Parent" in $properties ignored, with a warning to the user that said properties cannot be set. Maybe exiting with an error would be better?


LPGhatguy:
Thank you for the PR! I adjusted the code a little bit (using pattern matching instead of an array) and am merging it in!

LPGhatguy (approved):

Mixu78:
Cool! Was thinking about using a match at first but I don't quite know Rust best practices yet so I went with an array instead.

PR #420, opened 2021-05-02
Do not validate .model.json files with no content/whitespace only

### LINKED ISSUE #415: Only validate .model.json files after content has been added
When you create a new .model.json file Rojo immediately throws an error. This has been a longstanding cause for annoyance for me because when using the VS Code plugin it pops up an output window simply for creating a file.

It would be great if when a new .model.json file is added, Rojo only validates the contents once there's actually something in the file.

Relevant error:

```
[ERROR] Snapshot error: File is not a valid JSON model: c:\[...]/foo.model.json
        
        Caused by:
            EOF while parsing a value at line 1 column 0
[ERROR] Snapshot error: File is not a valid JSON model: c:\[...]/foo.model.json
        
        Caused by:
            EOF while parsing a value at line 1 column 0
```

Edit: This happens in Rojo v6.1.0, managed by the VS Code plugin.

Mixu78:
Closes #415

### DIFF
--- a/src/change_processor.rs
+++ b/src/change_processor.rs
@@ -13,7 +13,7 @@ use crate::{
     snapshot::{
         apply_patch_set, compute_patch_set, AppliedPatchSet, InstigatingSource, PatchSet, RojoTree,
     },
-    snapshot_middleware::{snapshot_from_vfs, snapshot_project_node},
+    snapshot_middleware::{snapshot_from_vfs, snapshot_project_node, util::match_file_name},
 };
 
 /// Owns the connection between Rojo's VFS and its DOM by holding onto another
@@ -283,6 +283,11 @@ fn compute_and_apply_changes(tree: &mut RojoTree, vfs: &Vfs, id: Ref) -> Option<
                 let snapshot = match snapshot_from_vfs(&metadata.context, &vfs, &path) {
                     Ok(Some(snapshot)) => snapshot,
                     Ok(None) => {
+                        // If the file that didn't return an instance is a model.json file
+                        // no return is expected as empty/whitespace mode.json files are ignored
+                        if let Some(_) = match_file_name(path, ".model.json") {
+                            return None;
+                        }
                         log::error!(
                             "Snapshot did not return an instance from path {}",
                             path.display()
--- a/src/change_processor.rs
+++ b/src/change_processor.rs
@@ -13,7 +13,7 @@ use crate::{
     snapshot::{
         apply_patch_set, compute_patch_set, AppliedPatchSet, InstigatingSource, PatchSet, RojoTree,
     },
-    snapshot_middleware::{snapshot_from_vfs, snapshot_project_node, util::match_file_name},
+    snapshot_middleware::{snapshot_from_vfs, snapshot_project_node},
 };
 
 /// Owns the connection between Rojo's VFS and its DOM by holding onto another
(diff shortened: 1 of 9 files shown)


src/snapshot_middleware/json_model.rs
```diff
@@ -18,6 +18,17 @@ pub fn snapshot_json_model(
     instance_name: &str,
 ) -> SnapshotInstanceResult {
     let contents = vfs.read(path)?;
+    if contents.is_empty() {
+        return Ok(None);
+    }
+
+    if match String::from_utf8((*contents).to_owned()) {
```
Kampfkarren, line 25: Why not str::from_utf8, rather than allocating an entirely new string here?

Mixu78: Didn't know about/look up str::from_utf8, I've changed it to use str::from_utf8 now

src/snapshot_middleware/json_model.rs
```diff
@@ -18,6 +18,17 @@ pub fn snapshot_json_model(
...
+
+    if match str::from_utf8(&contents) {
+        Ok(v) => v.is_empty() || v.trim().is_empty(),
+        _ => false,
+    } {
+        return Ok(None);
+    }
+
     let instance: JsonModel = serde_json::from_slice(&contents)
         .with_context(|| format!("File is not a valid JSON model: {}", path.display()))?;
```
LPGhatguy, line 33: We can simplify all of this code a lot by skipping redundant checks on the string. Additionally, we can switch to `serde_json::from_str` here to let Serde skip having to do UTF-8 processing again.

```suggestion
    let contents_str = str::from_utf8(&contents)
        .with_context(|| format!("File was not valid UTF-8: {}", path.display()))?;

    if contents_str.trim().is_empty() {
        return Ok(None);
    }

    let instance: JsonModel = serde_json::from_str(contents_str)
        .with_context(|| format!("File is not a valid JSON model: {}", path.display()))?;
```

src/change_processor.rs
```diff
@@ -283,6 +283,11 @@ fn compute_and_apply_changes(tree: &mut RojoTree, vfs: &Vfs, id: Ref) -> Option<
                 let snapshot = match snapshot_from_vfs(&metadata.context, &vfs, &path) {
                     Ok(Some(snapshot)) => snapshot,
                     Ok(None) => {
+                        // If the file that didn't return an instance is a model.json file
+                        // no return is expected as empty/whitespace mode.json files are ignored
+                        if let Some(_) = match_file_name(path, ".model.json") {
+                            return None;
+                        }
```
LPGhatguy, line 290: This does not feel like the correct solution here. I'd rather not special-case `.model.json` files and instead come back to this problem in another PR.

Alternatively, should we make `.model.json` files return a dummy instance, like an empty `Folder`? I think that returning `None` may cause Rojo to skip watching the file for changes, but I'm not sure.

src/snapshot_middleware/mod.rs
```diff
@@ -14,7 +14,7 @@ mod project;
 mod rbxm;
 mod rbxmx;
 mod txt;
-mod util;
+pub mod util;
```
LPGhatguy, line 17: I don't think we should be exporting utilities out of this module. See my other comment on the code using this export.

LPGhatguy (changes requested):
Thank you for the PR! Left some comments.

We should also add a test for this; you can see how tests are written [here](https://github.com/rojo-rbx/rojo/blob/532d1705854f7483d7f0956f7980c9907b27b684/tests/tests/serve.rs). Add a project with an empty `.model.json` file in it, assert that the tree looks correct, then write to the file with a new, valid `.model.json`. After the change registers, assert that the new tree looks correct.

LPGhatguy (approved):

PR #424, opened 2021-05-19
Add coverage reporting

matthargett:
This adds coverage reporting for PRs and merges to master, with a low-ball minimum coverage check.
A potential next step would be for an admin to add coveralls support to the repo, then add a coverage badge to the README.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,6 +17,14 @@ You'll want these tools to work on Rojo:
 * Latest stable [Rojo](https://github.com/rojo-rbx/rojo)
 * [Foreman](https://github.com/Roblox/foreman)
 
+To see the current way to build and run tests, read the [GitHub workflow files](.github/workflows/ci.yml) for the rojo Continuous Integration process.
+
+If you receive an error during the build process similar to
+```
+Error: failed to open file `/Users/alice/src/rojo/plugin/modules/xyz`
+```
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -17,6 +17,14 @@ You'll want these tools to work on Rojo:
 * Latest stable [Rojo](https://github.com/rojo-rbx/rojo)
 * [Foreman](https://github.com/Roblox/foreman)
 
+To see the current way to build and run tests, read the [GitHub workflow files](.github/workflows/ci.yml) for the rojo Continuous Integration process.
+
+If you receive an error during the build process similar to
+```
+Error: failed to open file `/Users/alice/src/rojo/plugin/modules/xyz`
+```
+, then you haven't initialized the git submodules correct. This can be fixed by running `git submodule update --init` within your existing rojo git clone.
```
Nezuo, line 26: The phrasing here is a little confusing to me. It might be better to say: "then the git submodules aren't initialized correctly."

matthargett: Updated. Let me know if that reads better.

LPGhatguy:
Since this PR was filed, Rust 1.60.0 was released with [official support for source-based coverage](https://blog.rust-lang.org/2022/04/07/Rust-1.60.0.html#source-based-code-coverage)! We should try this again, I think coverage is valuable and official support should make this much easier.

Issue #425, opened 2021-05-25
rojo upload doesn't appear to work


pracplayopen:
hello thanks for your rojo contribution.

am able to use rojo (v6 and v7) using rojo sync/serve.  thx

unable to get 'rojo upload' working, even though per your docs it's much simpler than sync.

attempted this using a project with one asset and one script.   always get error:

```
[ERROR] The Roblox API returned an unexpected error:
Bad Request
400 |There was a problem with your request
```

Since all that's being done is provide two roblox-supplied parameters via two args to 'roblox upload', this has to be rojo upload simply not working.    I've verified asset id and security cookie.  Tried many different interpretations of your docs, always get this error.

Then I looked at your source here: https://github.com/rojo-rbx/rojo/blob/c7ab6c435c3b66b7fa857ebb5553f262f9fbbe01/src/cli/upload.rs

I noticed you're using this url to upload:

```
let url = format!(        "https://data.roblox.com/Data/Upload.ashx?assetid={}",        asset_id    );
```

I tried to locate the roblox api docs and these show this being the correct asset upload url: https://publish.roblox.com/docs#!/Items/post_v1_assets_upload

When I rebuild rojo using this url, I get:

```
[ERROR] The Roblox API returned an unexpected error: {"errors":[{"code":0,"message":"The requested resource does not support http method 'POST'."}]}
```

Which is odd because the docs clearly show post as supported.

Will continue to look and post PR if fixing.

Also if something else is wrong please advise.

thx for sync.

LPGhatguy:
Thanks for the report!

This API is very temperamental. There are a few common issues:
1. The cookie has been invalidated. This can happen if you pressed "Log Out" on the website or if the cookie expired after ~6 months.
2. You are publishing to a universe/game/experience instead of a place. It's important to use the place ID.
3. You're using the wrong account. This happens a lot to me when I switch between accounts in Studio.

I recommend not specifying the upload cookie yourself if possible. On Windows, `rojo upload` can find your cookie from Roblox Studio for you.

pracplayopen:
thanks for follow up.

yes, based on the way they've implemented the login token... it's pretty clear api is a work in progress. :)

that said i'm certain i'm using right place id.

i was able to patch the upload command slightly so it outputs the status code of the error, which was helpful to fix one issue.
(previously it was spitting out response.text, which is basically a ton of useless javascript/css and no actual error code).

[...]

PR #431, opened 2021-06-14
better logging of upload errors, improved upload tracing

pracplayopen:
see open issue for more info:  https://github.com/rojo-rbx/rojo/issues/425#issuecomment-861015216

### DIFF
--- a/src/cli/upload.rs
+++ b/src/cli/upload.rs
@@ -37,6 +37,9 @@ impl UploadCommand {
             "Rojo could not find your Roblox auth cookie. Please pass one via --cookie.",
         )?;
 
+	let pathi = project_path.display();
+	log::trace!("project-path: {path} cookie: {cook}",path=pathi,cook=cookie);
+
         let vfs = Vfs::new_default();
 
         let session = ServeSession::new(vfs, project_path)?;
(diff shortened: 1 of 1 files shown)


src/cli/upload.rs
```diff
@@ -89,6 +92,7 @@ fn do_upload(buffer: Vec<u8>, asset_id: u64, cookie: &str) -> anyhow::Result<()>
         "https://data.roblox.com/Data/Upload.ashx?assetid={}",
         asset_id
     );
+	let urli = url.clone();
```
Kampfkarren, line 95: Why are you doing this? Looks like you can just `&url`?

pracplayopen: if you do that you'll get a borrow after move  compile error.

i'm not a rust expert but a generally programming practice that carries over well to passing rust's helpful compiler rules is to avoid re-using variables for different purposes.   

so to fix error and consistent w/general rule i cloned it to the url info variable.  it only lives for a few microseconds so should be fine for this application imho.

LPGhatguy: The correct solution here is not to clone the URL, it's just to move only a reference to the URL into the closure. I don't know what you mean about general programming practice, but it doesn't really apply here.

LPGhatguy:
I don't think this logging adds anything that's helpful for the user. Additionally, we really don't wanna printing out their authentication cookie into logs because it's really sensitive.

PR #446, opened 2021-07-14
Add Color3uint8 to shorthand property resolution

### LINKED ISSUE #443: Loading rojo built places loads baseplate without color
VERSION: 7.0.0-alpha.4

This is fixed upon syncing - even if the project.json is unchanged between the build and syncing. Not sure if this is an intended feature or not but it seems like unusual behaviour.

Before syncing
After syncing

Above images depict the issue discussed, first image is prior to syncing; second one is after. The texture *does* exist before syncing and is coloured appropriately (tried with a white colour and it loads in correctly). This seems to only affect Parts as the Texture's colour does translate to the build file correctly (tried 1,1,1 and 0,0,0 and both work).

This happens to all Parts that rojo manages in the project.json file.

kennethloeffler:
Depends on changes from rojo-rbx/rbx-dom#197 to close #443.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,9 @@ The shorthand property format that most users use is not impacted. For reference
 * Added the `fmt-project` subcommand for formatting Rojo project files.
 * Improved error output for many subcommands.
 * Updated to stable versions of rbx-dom libraries.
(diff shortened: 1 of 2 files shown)


kennethloeffler:
This has been superseded by rojo-rbx/rbx-dom#198.

PR #449, opened 2021-07-21
add optional tree node

### LINKED ISSUE #383: Support for Optional Tree Nodes
For [roblox-ts](https://www.roblox-ts.com), we'd like to have all of our npm packages provide Rojo `default.project.json` files. Unfortunately, there's a major issue blocking our support for this.

Here's an example `default.project.json` file for package:
```json
{
	"name": "my-package-name",
	"tree": {
		"$path": "out",
		"node_modules": {
			"$path": "node_modules/@rbxts"
		}
	}
}
```

The problem is that `node_modules` may not exist when this package is installed.
When a user installs the package, npm will flatten the dependencies so that every packages is under `/node_modules/*`. However, if any packages have conflicting versions of the same dependency, those packages will be given their own `node_modules` folder.

To solve this, we need some way or marking a Rojo tree node as "optional". Maybe as a `"$optional": true,` field?

MobiusCraftFlip:
closes #383 

I'm not very experienced with rust so I'm not that confident in this being merged.

### DIFF
--- a/rojo-test/build-tests/optional/default.project.json
+++ b/rojo-test/build-tests/optional/default.project.json
@@ -1,0 +1,14 @@
+{
+  "name": "optional",
+  "tree": {
+    "$className": "Folder",
+    "foo":{
+      "$path": "src/foo.txt",
+      "$optional": false
+    },
+    "bar":{
+      "$path": "src/bar.txt",
+      "$optional": true
+    }
+  }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Rojo Changelog
 
 ## Unreleased Changes
+
 In Rojo 6 and previous Rojo 7 alphas, an explicit Vector3 property would be written like this:
 
 ```json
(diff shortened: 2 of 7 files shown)


rojo-test/build-tests/optional/default.project.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "optional",
+  "tree": {
+    "$className": "Folder",
+    "foo":{
+      "$path": "src/foo.txt",
+      "$optional": false
```
sasial-dev, line 7:
```suggestion
```
shouldn't `"$optional": false` be omitted here?

sasial-dev (commented):
I'm not fluent in rust, so I just had a look through the rest of it.

Quenty:
The code here in `snapshot_middleware/project.rs:36` crashes when you have a blah.project.json that resolves to None. 

```rs
    // TODO: If this project node is a path to an instance that Rojo doesn't
    // understand, this may panic!
    let mut snapshot =
        snapshot_project_node(&context, path, &project.name, &project.tree, vfs, None)?.unwrap();
```

And needs to be changed to this
```rs
    let snapshot_option: Option<InstanceSnapshot> =
        snapshot_project_node(&context, path, &project.name, symlinks, &project.tree, vfs, None)?;

    if snapshot_option.is_none()
    {
        return Ok(None);
    }

    let mut snapshot = snapshot_option.unwrap();
```

PR #462, opened 2021-08-05
Fix Ref reification

kennethloeffler:
The reconciler's reify function was still using the old value format for Refs, causing Ref properties to fail to sync. This PR also updates the ApiValue type to use the new value format (this caused problems when type checking was enabled).


LPGhatguy (approved):
LGTM

PR #464, opened 2021-08-09
Support long file paths on Windows (past 256 limit)

Quenty:
This issue can occur when using symlinks deep in rojo such that very long paths can occur, among other scenarios.

Note while the original fix comes from here:
	https://gal.hagever.com/posts/windows-long-paths-in-rust/

The manifest had to be modified from this source:
	https://stackoverflow.com/questions/59816045/windows-10-1903-longpathaware-not-working


LPGhatguy (approved):
Good to go!

Issue #465, opened 2021-08-09
Output spam from plugin when syncing into place built by Rojo


ayuka:
I've fully migrated our project from a partially-managed setup to a fully-managed one now, and we're using the latest Rojo 7 alpha build due to the support of the binary format. Thank you for that!

However, when live-syncing into a place built by this it causes a huge lag spike and then major output spam. I think it spams well over 10,000 times but the studio output can only contain 5,000 at one time.

This is a large project and I'm unable to send a full build that reproduces this, however I am fully willing to go through any amount of steps necessary to help resolve this issue. I will try to create a sharable place file that reproduces this.

`[Rojo-Warn] Failed to decode property of type Ref (x4994)`

kennethloeffler:
We recently landed a fix related to `Ref` sync in #462. If you have cargo installed, you can try `cargo install --git https://github.com/rojo-rbx/rojo` to build Rojo off the master branch. It may or may not fix your issue but it's worth a shot!

ayuka:
I just installed from the repository and built the latest plugin from the master branch. I'm now receiving another message which is more obscure (?): `[Rojo-Warn] Failed to decode property of type nil (x4995)`

Since I have the source now, I'll see if I can debug a bit with it.

kennethloeffler:
Make sure to also reinstall Rojo's plugin using `rojo plugin install`

ayuka:
Yes, I built the plugin from source instead from the master branch and tried using the CLI install.

I printed off what was being passed and I'm getting this:
```
  11:26:26.428   ▼  {
                    ["Ref"] = "00000000000000000000000000000000"
                 }  -  Edit
  11:26:26.429  [Rojo-Warn] Failed to decode property of type nil  -  Edit
```

kennethloeffler:
It seems like you're still running the old plugin (I've experienced the same problem while working on it 😄). In previous versions of Rojo's protocol, property values had the fields `Type` and `Value`, but now they're of the form you see in the print. I think the plugin is still trying to access the `Type` field (which doesn't exist anymore) and is getting `nil`, hence the warning and failure to sync.

Uninstall or disable any other instances of the Rojo plugin in Roblox Studio, run `rojo plugin install`, then open a new place (or restart Studio) to make sure that the correct plugin is running.

ayuka:
I've managed to create a simple reproduction of this issue. It mostly seems that all object properties when nil cause this issue, on top of the old implementation from the start.

[...]

PR #466, opened 2021-08-09
Update nil Ref check and property decode warning to new Rojo protocol

ayuka:
I had ran into an issue with Refs in #465 on an older version of Rojo, and was guided to try out the latest version and still had issues. These parts of the plugin were relying on the old Rojo protocol and I've now updated them.

Regarding the comment I left on diff.lua#78, I've described one instance that can cause the table to be empty in #467.

*The unverified is because my GPG for Windows is kinda messed up. The next commit is verified me!*


LPGhatguy (approved):
Thank you!

PR #468, opened 2021-08-11
Play Solo Auto Connect

zoebasil:
Adds a setting to automatically re-connect to the rojo server when entering play solo


plugin/src/init.server.lua
```diff
@@ -20,6 +20,11 @@ local app = Roact.createElement(App, {
 })
 local tree = Roact.mount(app, nil, "Rojo UI")
 
+game.Close:Connect(function()
```
Kampfkarren, line 23: `game.Close` doesn't exist from my search, do you mean `OnClose` (deprecated) or `BindToClose`?

plugin/src/App/Components/PlaySoloAutoConnector.lua
```diff
@@ -0,0 +1,37 @@
...
+
+local Rojo = script:FindFirstAncestor("Rojo")
+
+local Roact = require(Rojo.Roact)
+
+local e = Roact.createElement
+
+local PlaySoloAutoConnector = Roact.Component:extend("PlaySoloAutoConnector")
+
+function PlaySoloAutoConnector:render()
```
Kampfkarren, line 11:
```suggestion
function PlaySoloAutoConnector:render()
	return nil
```

plugin/src/App/PluginSettings.lua
```diff
@@ -9,6 +9,8 @@ local Roact = require(Rojo.Roact)
 local defaultSettings = {
 	openScriptsExternally = false,
 	twoWaySync = false,
+	playSoloAutoConnect = false,
+	activeConnections = {},
```
Kampfkarren, line 13:
```suggestion
	activeConnections = {},
	openScriptsExternally = false,
	playSoloAutoConnect = false,
	twoWaySync = false,
```

plugin/src/init.server.lua
zoebasil: i guess its undocumented?
image

zoebasil: nope, its https://developer.roblox.com/en-us/api-reference/event/ServiceProvider/Close

Kampfkarren: Whack, wonder why it's not on Anaminus' site that I checked.

zoebasil: am i okay to mark this as resolved?

unix-system: Is there a reason we're using this instead of `BindToClose` ? As in, is there some behaviour we need from this obscure method. I think we should understand this thing before we put it into Rojo.

zoebasil: nah, ill just change it to BindToClose

sasial-dev:
Other than merge conflicts, is there anything else this PR needs to get merged?

LPGhatguy:
> Other than merge conflicts, is there anything else this PR needs to get merged?

In general, yeah! There's a bit of feature design work that I think we need to do, especially around making sure Rojo won't try to overwrite stuff that changes at runtime.

I've tagged PRs on the repo to try to indicate what their current status is.

jackTabsCode:
I'm curious what's blocking this feature from happening. This is something I am looking for! Would save me a decent chunk of time when testing.

Dekkonot:
I don't see a reason this is blocked beyond the fact that I've not reviewed this PR and it's old. I've asked @boatbomber to take a look since he's more in tune with Rojo's plugin source.

We could solve this in a few ways and this is just one of them, but I'd like his opinion since it'll inevitably be him that maintains any code for it we add (unless something critical breaks).

plugin/src/App/Components/PlaySoloAutoConnector.lua
```diff
@@ -0,0 +1,38 @@
...
+	return nil
+end
+
+function PlaySoloAutoConnector:didMount()
```
boatbomber, line 15: I'm not a big fan of this architecture, as it uses prop drilling and a component that just calls a function for the app. Instead, we can use the architecture that we now use for many other features- a call from `App:init()` that uses `self:startSession()`

plugin/src/App/init.lua
```diff
@@ -59,15 +63,57 @@ function App:startSession(host, port, sessionOptions)
 				appStatus = AppStatus.Connecting,
 			})
 		elseif status == ServeSession.Status.Connected then
+			if expectedSessionId and details.sessionId ~= expectedSessionId then
+				ServeSession:stop()
+			end
+
 			local address = ("%s:%s"):format(host, port)
+
+			if RunService:IsClient() and RunService:IsServer() and not RunService:IsRunning() then
```
boatbomber, line 72: Isn't this just `RunService:IsEdit()`?

plugin/src/App/init.lua
```diff
@@ -59,15 +63,57 @@ function App:startSession(host, port, sessionOptions)
...
+					port = port
+				})
+
+				if #activeConnections > 15 then
+					table.remove(activeConnections, 1)
+				end
+
+				settings:set("activeConnections", activeConnections)
+
+				workspace:SetAttribute("__RojoConnectionId", connectionId)
```
boatbomber, line 88: This is putting temporary data into a permanent location- could we use a non-Archivable Instance for this, or do they disappear when you press Play Solo? It might simplify the cleanup logic.

plugin/src/App/init.lua
```diff
@@ -59,15 +63,57 @@ function App:startSession(host, port, sessionOptions)
...
+					sessionId = details.sessionId,
+					host = host,
+					port = port
+				})
+
+				if #activeConnections > 15 then
+					table.remove(activeConnections, 1)
+				end
+
+				settings:set("activeConnections", activeConnections)
```
boatbomber, line 86: I don't understand why we have this array and write it to `settings.json`.

You're creating a GUID to key the host/port/sessionId within this array, then sharing the GUID via the Attribute, then linearly searching the array to find the host/port/sessionId with this GUID when play solo.
(Why are do a linear search? You can turn this array into a dict using the GUID as a key.)

I think this array and disc write is totally avoidable though, unless there's something I'm missing.
You can write the connection info to an easily parse-able string in the Attribute, so that the server window of the test can grab the host/port/sessionId without having to search through other connections from other projects in other windows.

IE:
```Lua
-- When connecting
workspace:SetAttribute("__RojoConnectionInfo", host .. "|" .. port .. "|" .. details.sessionId)

-- When in play solo
local connectionInfo = workspace:GetAttribute("__RojoConnectionInfo")
if not connectionInfo then return end
local host, port, sessionId = table.unpack(string.split(connectionInfo, "|"))
```

boatbomber (changes requested):
I have some concerns about implementation, and I'm not sure if we ever addressed LPG's note about avoiding overwriting dynamic content? We might need to force `ignoreUnknownInstances` on everything during a Play Solo connection.

This PR is also 3 years old and out of date, it might just be easier to start a new one.

plugin/src/App/init.lua
zoebasil: I agree that the disc write isn’t the way to go. It felt shaky even when I implemented it.

 As for the rest of your questions— I don’t have good answers for those, mostly because I can’t remember and I’d need to read back through the whole PR, which I’m not really interested in doing right now. (nor am I interested in continually addressing those issues in this PR)

I think it would be better for you to restart this feature from scratch

plugin/src/App/init.lua
zoebasil: TIL

plugin/src/App/init.lua
boatbomber: Yup, I opened #840 for that. Thanks for getting this started though!

Dekkonot:
Implemented in #840

PR #472, opened 2021-08-26
Add support for optional paths

### LINKED ISSUE #383: Support for Optional Tree Nodes

Quenty:
Closes #383 by adding the fields `"path": "src"` or `"path": { "optional": "src" }` which indicates that the path is optional. Also adds a unit test by @mobiuscraftflip adjusted for this format.

I recommend reviewing this PR commit-by-commit for clarity.

This pull request enables scenarios with `node_modules` and other tools that sometimes create folders and sometimes does not.


src/project.rs
```diff
@@ -168,6 +168,38 @@ impl Project {
...
+}
+
+impl PathNode {
+    pub fn as_path(&self) -> &PathBuf {
+        match self {
+            PathNode::Required(pathbuf) => &pathbuf,
+            PathNode::Optional(OptionalPathNode { optional }) => &optional,
+        }
+    }
+}
```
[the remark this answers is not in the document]
Quenty: This was very interesting, thank you for sharing!

src/snapshot/patch_compute.rs
```diff
@@ -10,16 +10,27 @@ use super::{
     InstanceSnapshot, InstanceWithMeta, RojoTree,
 };
 
-pub fn compute_patch_set(snapshot: &InstanceSnapshot, tree: &RojoTree, id: Ref) -> PatchSet {
+pub fn compute_patch_set(
+    snapshot: &Option<InstanceSnapshot>,
+    tree: &RojoTree,
+    id: Ref,
+) -> PatchSet {
```
LPGhatguy, line 17: I like this change!

Good opportunity for a Rust note: always prefer `Option<&T>` over `&Option<T>`. They're almost exactly the same, but `Option<&T>` is more flexible because it allows creating an Option of something you don't own. For other code, it just results in the ampersand moving a couple characters over.

LPGhatguy (changes requested):
Digging the latest round of changes!

It looks like there's a test failure, make sure you commit your updated Insta snapshot tests.

This should be the last feedback from me. If everything works from here, then I think we're good to go. I'd appreciate it if you would also file a PR on https://github.com/rojo-rbx/rojo.space with documentation updates! 😊

Quenty:
I'll be taking a look at this today, sorry for the delay.

src/snapshot/patch_compute.rs
Quenty: Super interesting, thank you so much!

ZeroParticle:
🥺

LPGhatguy (approved):
Going through with merging this, and opened an issue about the remaining issue.

PR #478, opened 2021-09-20
Add ChangeBatcher for two-way sync

### LINKED ISSUE #294: Debounce instance property changes

kennethloeffler:
Closes #294.

Depends on changes from rojo-rbx/rbx-dom#210.

This PR implements `ChangeBatcher`. Its job is to collect any instance changes and periodically try to turn them into a patch to send to the Rojo server. Right now, this happens every 200ms.

There is one small bump I hit during implementation. When a change is made to the `Source` property of a `LuaSourceContainer`, AND the change is made from a script or the command bar, AND the script is open in Studio, the `Changed` event fires three times for `Source`. The first event occurs right when the change is made (or, if deferred events are enabled, at the next invocation point), while the latter two events occur some time after the current `Heartbeat` cycle. 

If this is left unchecked, the `ChangeBatcher` detects a spurious change to `Source` after a patch application. To fix it, I added `InstanceMap:pauseInstanceForBatch`. It pauses an instance until the `ChangeBatcher` decides to unpause it (in this case, after the next `Heartbeat` cycle). I also made `InstanceMap:pauseInstance` no longer take a callback and added `InstanceMap:unpauseInstance` to avoid having to create a closure for every set of property updates during patch application.

### DIFF
--- a/plugin/src/ChangeBatcher.lua
+++ b/plugin/src/ChangeBatcher.lua
@@ -1,0 +1,155 @@
+--[[
+	The ChangeBatcher is responsible for collecting and dispatching changes made
+	to tracked instances during two-way sync.
+]]
+
+local RunService = game:GetService("RunService")
+
+local PatchSet = require(script.Parent.PatchSet)
+local RbxDom = require(script.Parent.Parent.RbxDom)
+local Log = require(script.Parent.Parent.Log)
+
+local ChangeBatcher = {}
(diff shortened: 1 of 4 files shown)


kennethloeffler:
It's more or less where I want it for now - any further revisions are welcome

kennethloeffler:
For some reason I had it in my head that `RunService.RenderStepped` throws when used on Roblox Studio's server mode - but I just checked and it does not!

PR #479, opened 2021-09-20
Track patch sources, make the plugin ignore its own changes

### LINKED ISSUE #295: Track "source" for PatchSet
To prevent changes ping-ponging between clients, patch sets should have a `source` field attached to them. Each endpoint, including servers and clients, should generate a GUID source for themselves so that they can ignore changes that come from their own changes.

In tandem with #293, we should be able to track change sources for two-way sync and prevent clients from getting into funny states with themselves at least.

kennethloeffler:
Closes #295.

This PR adds the newtype `PeerId` and adds a `source: PeerId` member to patch sets; this allows a subscriber to identify who was responsible for any given patch set. It uses this to make the plugin ignore changes that the plugin itself sent during two-way sync.

### DIFF
--- a/plugin/http/init.lua
+++ b/plugin/http/init.lua
@@ -63,4 +63,8 @@ function Http.jsonDecode(source)
 	return HttpService:JSONDecode(source)
 end
 
+function Http.generateGuid()
+	return HttpService:GenerateGUID(false):lower()
+end
+
 return Http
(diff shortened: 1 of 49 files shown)


LPGhatguy:
This is a great change. I need to find time to sit down and review it (and other changes!)

PR #495, opened 2021-10-26
Add git submodules info to CONTRIBUTING.md

blake-mealey:
I was pretty confused for a while about the missing submodules, so I figured I'd add this to help others in the future. I've never used submodules before so if I'm using the wrong command to load them, lmk and I can update.


LPGhatguy (approved):
Thank you for the docs fix! Hopefully we'll migrate these to Wally soon enough and stop having submodules at all.

PR #496, opened 2021-10-26
added caching to ci builds to reduce build times

blake-mealey:
This adds caching to the CI workflow which should improve build times quite a bit. I used the recommended cache parameters from here: https://github.com/actions/cache/blob/main/examples.md#rust---cargo

I tried this approach in my project and it took build times from ~1min to ~10s:
* before: https://github.com/blake-mealey/rocat/runs/3999515438?check_suite_focus=true
* after: https://github.com/blake-mealey/rocat/runs/4015757268?check_suite_focus=true

Note that I haven't added caching to the release workflow. We could do this as well but I don't think it'll give you any benefit since whenever you run that workflow your Cargo.toml has changed (bumped version number) and therefore it will always have a cache miss.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,11 +3,11 @@ name: CI
 on:
   push:
     branches:
-    - master
+      - master
 
   pull_request:
     branches:
-    - master
+      - master
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


LPGhatguy:
This is neat, but @Kampfkarren's concerns about cache spookiness resonate with me. I don't think our CI is quite slow enough for me to want to take this on yet!

I'd like to keep this PR open for future reference.

Dekkonot:
Already implemented in our CI workflow so I'm closing this to clean up our PRs.

PR #500, opened 2021-11-13
Remember sync connection settings.

### LINKED ISSUE #103: Preserve port number in connection dialog

GlowingUmbreon:
This pull request makes it so that the host and port are not cleared when connecting.
Closes #103


LPGhatguy:
This is great and very simple, thank you!

LPGhatguy (approved):

PR #504, opened 2021-12-06
Support the new Open Cloud API

### LINKED ISSUE #486: Support the new Open Cloud endpoint for uploading
Since the new [Open Cloud feature](https://developer.roblox.com/en-us/articles/open-cloud) is now in public beta, cookie authentication is close to being replaced for publishing places.

Rojo should eventually start supporting `--api-key` instead of `--cookie` for `rojo upload` using the [Place Publish endpoint](https://developer.roblox.com/en-us/articles/open-cloud#publishing-places) and any other upload endpoints that are added to open cloud, since this'll in theory be Roblox's official endpoint for tools like Rojo to use.

VernandoGames:
Closes #486 by providing a ``--api_key`` along with ``--universe_id`` in place of ``--cookie``. Universe ID is required in order to use the open cloud api.
So new upload commands will look like ``rojo upload --api_key <API_KEY> --asset_id <PLACE_ID> --universe_id <UNIVERSE_ID> [project file]``
Instead of ``rojo upload --cookie <COOKIE> --asset_id <PLACE_ID> [project file]``

This was made with an accumulative ~3 hours of experience with rust, so I'm sure things could be done better and I'm open to suggestions. Most notably lines 48-68.

### DIFF
--- a/src/cli/upload.rs
+++ b/src/cli/upload.rs
@@ -24,13 +24,13 @@ pub struct UploadCommand {
     #[structopt(long)]
     pub cookie: Option<String>,
 
-	/// Used for the new Open Cloud API. Still experimental.
-	#[structopt(long = "api_key")]
-	pub api_key: Option<String>,
+    /// Used for the new Open Cloud API. Still experimental.
+    #[structopt(long = "api_key")]
+    pub api_key: Option<String>,
 
-	/// Required when using the Open Cloud API.
-	#[structopt(long = "universe_id")]
-	pub universe_id: Option<u64>,
+    /// Required when using the Open Cloud API.
+    #[structopt(long = "universe_id")]
+    pub universe_id: Option<u64>,
 
     /// Asset ID to upload to.
     #[structopt(long = "asset_id")]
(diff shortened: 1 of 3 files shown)


src/cli/upload.rs
```diff
@@ -125,3 +159,41 @@ fn do_upload(buffer: Vec<u8>, asset_id: u64, cookie: &str) -> anyhow::Result<()>
...
+
+    let client = reqwest::Client::new();
+
+    let build_request = move || {
+        client
+            .post(&url)
+            .header("x-api-key", api_key)
+            .header(CONTENT_TYPE, "application/xml")
+            .header(ACCEPT, "application/json")
+            .body(buffer.clone())
```
Kampfkarren, line 181: Why clone the buffer? Looks like you don't use it again.

VernandoGames: That was actually copied and modified from the do_upload function that already exists. In the case of do_upload, this is required for the CSRF challenge. I've pushed a commit that will remove this from do_upload_open_cloud.

VernandoGames:
I believe this build is ready to be merged btw, The last commit was just to fix some formatting issues found with cargo fmt.

src/cli/upload.rs
```diff
@@ -33,9 +41,30 @@ impl UploadCommand {
...
+        } else {
+            0
+        };
+        let cookie = if use_open_cloud {
+            "undefined".to_string()
+        } else {
+            self.cookie.or_else(get_auth_cookie).context(
+                "Rojo could not find your Roblox auth cookie. Please pass one via --cookie.",
+            )?
+        };
```
LPGhatguy, line 67: This logic is pretty funky. I think we should move this logic down to where `do_upload` is called, and take advantage of Rust's match statement against all three of these values. You can use this to make some very detailed diagnostic messages for users, which can be very helpful!

I would write something like this:

```rust
match (self.cookie, self.api_key, self.universe_id) {
	(cookie, None, universe) => {
		if universe.is_some() {
			log::warn!("--universe_id was provided but is ignored when using legacy upload API");
		}

		let cookie = cookie.unwrap_or(get_auth_cookie).context("<good no cookie error>")?;
		do_upload(buffer, self.asset_id, &cookie)
	}

	(cookie, Some(api_key), Some(universe_id)) => {
		if cookie.is_some() {
			log::warn!("--cookie was provided but is ignored when using Open Cloud API");
		}

		do_upload_open_cloud(buffer, universe_id, self.asset_id, &api_key)
	}

	(_, Some(_), None) => {
		bail!("If --api_key is provided to upload, --universe_id must also be provided");
	}
}
```

src/cli/upload.rs
```diff
@@ -54,7 +83,11 @@ impl UploadCommand {
 
         log::trace!("Encoding binary model");
         rbx_binary::to_writer(&mut buffer, tree.inner(), &encode_ids)?;
-        do_upload(buffer, self.asset_id, &cookie)
+        if use_open_cloud {
+            return do_upload_open_cloud(buffer, universe_id, self.asset_id, &api_key);
+        } else {
+            return do_upload(buffer, self.asset_id, &cookie);
+        }
```
LPGhatguy, line 90: You don't need `return` in Rust, you can just remove the semicolons here. See above comment for how I think this can be integrated into the parameter reading code.

LPGhatguy (changes requested):

src/cli/upload.rs
VernandoGames: That's very close to what I was originally thinking (although I didn't know you could have multiple inputs in a match statement), but this would mean that it would have to build the buffer before it is able to verify if the information given is even correct. 

While this isn't technically an issue, I would like to avoid having that extra delay before you get to know if your information is correct. We may be able to keep the minimal delay by writing the binary file after it has verified the information but building the session and tree beforehand to minimize repeated code. (assuming writing the binary is actually what is taking the most amount of time)

I'll look into the best way to integrate this without adding unnecessary delay.

LPGhatguy (approved):
This looks excellent, sorry for the long turnaround!

Wingysam:
@VernandoGames @LPGhatguy Maybe use https://apis.roblox.com/universes/v1/places/:placeId/universe if the Universe ID isn't set to automatically fetch it?

PR #527, opened 2022-03-10
Support for Infinity,-Infinity, and NaN

### LINKED ISSUE #363: No support for infinity
When trying to sync a model that contains an instance with a property set to `inf` Rojo will set that property to `0`.

For example, if I have a `BodyPosition` and set the `MaxForce` property in Studio to `0, inf, 0` and then later on sync this with Rojo, the property will be changed to `0, 0, 0`.

st0nerhat:
Resolves #363

This PR aims to enable support for the Roblox dom notions of INF, -INF, and NAN which are represented in rbxl, rbxlx, and yaml but not in json. This proposed solution swaps the serde_json serializer with the json5 serializer which is also based on serde. Roblox can't read json5, so we replace the 3 special tokens used to encode IEEE 754 representations for non-numbers in json5 to string equivalents (Infinity -> "Infinity", -Infinity -> "-Infinity", NaN -> "NaN") by brute force (i.e. string replace).

This solution could be simplified if we had a json serializer that wrote the non-number representations as strings to begin with. But I couldn't identify a rust pattern that would accomplish that without re-implementing json serialization whole cloth. Looking for any leads on that front if they exist.

One disadvantage of this solution is that while you can write json project and model files in json5 format, a lot of tools are not setup to recognize or format json5 so will display visible warnings when entering these tokens in the file.

TODO:

- [x] Fix test failures:     tests::build::build_ignore_glob_inner, tests::build::build_ignore_glob_nested, tests::build::build_ignore_glob_spec


Kampfkarren:
Can you write "Resolves #363" or similar in the PR description so it automatically closes when this is merged? As well as putting a brief description of the issue in your title? This requires me to do extra reading in order to understand what this PR is doing.

plugin/http/Encode.lua
```diff
@@ -0,0 +1,23 @@
+local HttpService = game:GetService("HttpService")
+
+
+local Encode = {}
+
+function Encode.jsonEncode(object)
+	-- Encode json 5 IEEE 754 tokens
+	local body = HttpService:JSONEncode(object)
```
st0nerhat, line 8: I don't want to use regex here. But given the limited number of non-quoted tokens we need to handle it seems superior to alternatives

plugin/http/Response.lua
```diff
@@ -28,7 +28,12 @@ function Response:isSuccess()
 end
 
 function Response:json()
-	return HttpService:JSONDecode(self.body)
+	return Encode.jsonDecode(self.body)
 end
 
+function Response.setEncode(encode)
```
st0nerhat, line 34: Inject the encode functions here to avoid a circular dependency between Response and Http modules

plugin/rbx_dom_lua/EncodedValue.lua
```diff
@@ -11,13 +11,44 @@ local function unpackDecoder(f)
 end
 
 local function serializeFloat(value)
```
st0nerhat, line 13: Does this need to be contributed upstream? I'm not clear on whether the copy of rbx_dom_lua in rojo is fully synced with upstream.

src/project.rs
```diff
@@ -121,11 +122,14 @@ impl Project {
         contents: &[u8],
         project_file_location: &Path,
     ) -> Result<Self, ProjectError> {
-        let mut project: Self =
-            serde_json::from_slice(&contents).map_err(|source| Error::Json {
```
st0nerhat, line 125: json5 doesn't implement anything like from_slice

st0nerhat:
@Kampfkarren after sleeping on this, I feel that I'm ready for your feedback on whether this direction is plausible. I'm pretty certain I'll be able to get the remaining 3 failing tests passing today.

plugin/http/Encode.lua
st0nerhat: I just realized I can probably refine this to handle all non-quotes string tokens in one pattern with a group

st0nerhat:
I've done some exploratory testing with some real-world places and the regex is not going to hold water. I need to take a different approach here.

PR #530, opened 2022-04-04
Implement sourcemap CLI command

filiptibell:
### PR summary

This PR implements a new `sourcemap` command which produces an instance tree in JSON format in the following shape:

```json5
{
  // Equivalent to instance.Name in Roblox
  "name": "...",
  // Equivalent to instance.ClassName in Roblox
  "className": "...",
  // Array of existing file paths that contributed to the creation of this instance
  "filePaths": ["src/path/init.lua", "src/path/init.meta.json"],
  // Array of children with the same structure
  "children": []
}
```

This enables new & interesting use cases, as well as consistency in other tools/apps when using Rojo.
* Tools like [luau-analyze-rojo](https://github.com/JohnnyMorganz/luau-analyze-rojo) and [Roblox LSP](https://github.com/NightrainsRbx/RobloxLsp) should no longer need to do their own parsing of file structures and suffixes to get the same instance structure that Rojo provides
* Stack traces produced in Roblox can now be linked back to their original file paths, potentially enabling much simpler debugging workflows as well as additional features for services such as [Sentry](https://sentry.io/)

--------

### Initial implementation / design details

The command takes a project file as input or defaults to `default.project.json` just like other CLI commands.

The json data will either be written to a file specified by the `--output` parameter, similar to the `build` command, or if omitted will use `stdout` so that other tools do not have to write to and read a file as a proxy.

A flag `--include-non-scripts` determines if **all** instances should be included in the sourcemap, or just scripts. This can greatly reduce sourcemap size if the Rojo project is fully managed. If the flag is not provided, only scripts will be included in the source map. ***NOTE:** Might need a better name or default here?*

### DIFF
--- a/src/cli/sourcemap.rs
+++ b/src/cli/sourcemap.rs
@@ -33,7 +33,7 @@ struct SourcemapNode {
     children: Option<Vec<SourcemapNode>>,
 }
 
-/// Generates a model or place file from the Rojo project.
+/// Generates a sourcemap file from the Rojo project.
 #[derive(Debug, StructOpt)]
 pub struct SourcemapCommand {
     /// Path to the project to serve. Defaults to the current directory.
--- a/src/cli/sourcemap.rs
+++ b/src/cli/sourcemap.rs
@@ -36,7 +36,7 @@ struct SourcemapNode {
 /// Generates a sourcemap file from the Rojo project.
 #[derive(Debug, StructOpt)]
 pub struct SourcemapCommand {
-    /// Path to the project to serve. Defaults to the current directory.
+    /// Path to the project to use for the sourcemap. Defaults to the current directory.
     #[structopt(default_value = "")]
     pub project: PathBuf,
 
--- a/src/cli/sourcemap.rs
+++ b/src/cli/sourcemap.rs
@@ -1,20 +1,17 @@
 use std::{
-    sync::MutexGuard,
-    io::{BufWriter, Write, stdout},
-    path::PathBuf
+    io::{BufWriter, Write},
+    path::{Path, PathBuf},
 };
 
-use memofs::Vfs;
 use fs_err::File;
+use memofs::Vfs;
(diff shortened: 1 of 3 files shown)


src/cli/sourcemap.rs
```diff
@@ -0,0 +1,150 @@
...
+struct SourcemapNode {
+    name: String,
+    class_name: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    file_paths: Option<Vec<PathBuf>>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    children: Option<Vec<SourcemapNode>>,
+}
+
+/// Generates a model or place file from the Rojo project.
```
JohnnyMorganz, line 36:
```suggestion
/// Generates a sourcemap file from the Rojo project.
```

src/cli/sourcemap.rs
```diff
@@ -0,0 +1,150 @@
...
+#[derive(Debug, StructOpt)]
+pub struct SourcemapCommand {
+    /// Path to the project to serve. Defaults to the current directory.
```
JohnnyMorganz, line 39:
```suggestion
    /// Path to the project to use for the sourcemap. Defaults to the current directory.
```

JohnnyMorganz (commented):
Neat! It would be great if there were some snapshot tests highlighting the output

LPGhatguy (approved):
Thanks for the PR and sorry for the delay in getting it reviewed!

I pushed a couple changes onto the branch. Changes fell into a couple categories:
- Small interface changes: `recurse_create_node` now accepts a filter function instead of a bool and passes referents instead of instances. This cut down the code a little and made it a little easier to read.
- Rust-y things: Accepting `&Path` and `&RojoTree`, using an empty `Vec<T>` instead of `Option<Vec<T>>`. Nothing major here, functionally very similar.
- Minor formatting

With CI passing, this should be good to go. Great idea and submission!

PR #534, opened 2022-04-18
feat: Update reconciler / reify to use CreateMeshPartAsync to create mesh parts

### LINKED ISSUE #402: Use CreateMeshPartAsync to create mesh parts
GetObjects is weirdly unstable, and likely slower.

https://developer.roblox.com/en-us/api-reference/function/InsertService/CreateMeshPartAsync

unix-system:
Closes #402.

## Problem
Currently, Rojo does not support creation of some instances and properties via the Studio plugin - notably UnionOperations and MeshParts. MeshParts can in fact be created by Plugins through the [InsertService.CreateMeshPartAsync](https://developer.roblox.com/en-us/api-reference/function/InsertService/CreateMeshPartAsync) method at Plugin security.

This impacts developers using Rojo to work, as it prevents them from live importing any such instances (and can only be created from a file via `rojo build` or manual import).

## Solution
This pull request exclusively modifies the Lua plugin; no changes were necessary to the Rojo binary itself. The changes have a soft dependency on a [pull request to rbx-dom](https://github.com/rojo-rbx/rbx-dom/pull/216) to update the database files, although this is not a requirement.

The primary change was to modify the [reify function](https://github.com/rojo-rbx/rojo/pull/534/files#diff-a3cbd68f9ad15107f34668d0f42bd2cc53c44fa158cc9e1b5e8174d27369c109) to include a switch for custom instance creation logic which could be expanded in the future if necessary.

This change was coupled closely with the implementation of the `requiresRecreate` property of `ApiInstanceUpdate` objects, which is a partial replacement for the `changedClassName` property. This property triggers a condition in the [applyPatch function](https://github.com/rojo-rbx/rojo/pull/534/files#diff-921a0cda9152d2b6f49bade0b7322e5ad9fa6423f437aa0ce4fb7a22aab84c21) to recreate the instance using reify if the `requiresRecreate` property is present (previously done for the `changedClassName` property).

The `requiresRecreate` property is set during the creation of an update patch (either in batch, in [encodePatchUpdate](https://github.com/rojo-rbx/rojo/pull/534/files#diff-b31fd7f75bcd2afd22fc3346a8317f8530bfe1e5fc2d452885dde996a0a6e34e), or [during diff](https://github.com/rojo-rbx/rojo/pull/534/files#diff-59089ce4b718a7c79cacdf918e0bbeee07b702d7d8a18b3b2bc65e7df5809d5e)).

## Tests
Tests have been implemented which support the new feature being added. A dedicated test project (`test-projects/mesh_parts/default.project.json`) has been included with two MeshParts (one `rbxm` and one `rbxmx`), of the [Utah Teapot](https://en.wikipedia.org/wiki/Utah_teapot).

image

## Other
The development version of Rojo was bumped to `7.0.0 (latest)`.


plugin/src/ChangeBatcher/encodePatchUpdate.lua
```diff
@@ -7,28 +7,33 @@ return function(instance, instanceId, properties)
 	local update = {
 		id = instanceId,
 		changedProperties = {},
+		requiresRecreate = false
 	}
 
 	for propertyName in pairs(properties) do
 		if propertyName == "Name" then
 			update.changedName = instance.Name
-		else
```
unix-system, line 15: This change, whilst structural (it now encodes the **Name** field as well as setting the **changedName** property) has not changed the intended logic and is more concise.

plugin/src/Reconciler/reify.lua
```diff
@@ -40,6 +42,20 @@ local function addAllToPatch(patchSet, virtualInstances, id)
 	end
 end
 
+local function createInstance(instanceMap, className, properties)
+	-- Custom logic for MeshParts
+	-- see: https://github.com/rojo-rbx/rojo/issues/402
+	if className == "MeshPart" then
+		local okMeshId, meshId = decodeValue(properties.MeshId, instanceMap)
```
LPGhatguy, line 49: We need to handle `okMeshId`, because the second return value will not be a mesh ID if this failed to decode.

plugin/src/Types.lua
```diff
@@ -25,6 +25,7 @@ local ApiInstanceUpdate = t.interface({
 	id = RbxId,
 	changedName = t.optional(t.string),
 	changedClassName = t.optional(t.string),
+	requiresRecreate = t.optional(t.boolean),
```
LPGhatguy, line 28: I don't think we should update this interface at all. The server generates most of the patches that the Rojo plugin handles and it does not populate `requiresRecreate`. We can calculate this when we go to apply a patch instead.

plugin/src/Reconciler/applyPatch.lua
```diff
@@ -88,10 +88,10 @@ local function applyPatch(instanceMap, patch)
...
 		local partiallyApplied = false
 
-		-- If the instance's className changed, we have a bumpy ride ahead while
-		-- we recreate this instance and move all of its children into the new
-		-- version atomically...ish.
-		if update.changedClassName ~= nil then
+		-- If the instance's className changed, or there is custom recreate logic,
+		-- we have a bumpy ride ahead while we recreate this instance and move all 
+		-- of its children into the new version atomically...ish.
+		if update.requiresRecreate then
```
LPGhatguy, line 94: There are comments in this block of code about its caveats, notably not preserving existing properties if we end up needing to recreate an instance. It's probably not the end of the world, but some of the comments I left about this mostly occurring during Folder<->ModuleScript transitions are probably not accurate anymore.

plugin/src/ChangeBatcher/encodePatchUpdate.spec.lua
```diff
@@ -37,6 +37,18 @@ return function()
 		expect(update.changedName).to.equal("We'reGettingToTheCoolPart")
 	end)
 
+	it("should recreate instance in the update when the instance's ClassName changes", function()
+		local part = Instance.new("Part")
+		local properties = {
+			ClassName = true
```
LPGhatguy, line 43: We should set this value to a real `ClassName` so that future validation in the patch update function won't choke when it sees a bool instead.

plugin/src/Reconciler/applyPatch.spec.lua
```diff
@@ -156,7 +158,33 @@ return function()
 		expect(value.Value).to.equal("WORLD")
 	end)
 
-	it("should recreate instances when changedClassName is set, preserving children", function()
+	it("should recreate instances for MeshId updates", function()
```
LPGhatguy, line 161: This is a great test!

plugin/src/Reconciler/applyPatch.lua
LPGhatguy, line 94: Instead of having `requiresRecreate` as a property here, we should check the conditions that would cause us to recreate an instance directly. For now, that's just `changedClassName ~= nil or changedProperties["MeshId"] ~= nil`.

plugin/src/Reconciler/reify.lua
```diff
@@ -40,6 +42,20 @@ local function addAllToPatch(patchSet, virtualInstances, id)
...
+
+		-- It is okay to provide default properties here, because 
+		-- they will be overriden on the update cycle
+		return InsertService:CreateMeshPartAsync(meshId, Enum.CollisionFidelity.Default, Enum.RenderFidelity.Automatic)
```
LPGhatguy, line 53: This is the first time we'll be introducing an async function into Rojo's reconciler bits. This is a possible source of bugs that we should watch out for in the future.

plugin/src/Reconciler/reify.lua
```diff
@@ -40,6 +42,20 @@ local function addAllToPatch(patchSet, virtualInstances, id)
...
 end
 
+local function createInstance(instanceMap, className, properties)
+	-- Custom logic for MeshParts
+	-- see: https://github.com/rojo-rbx/rojo/issues/402
+	if className == "MeshPart" then
+		local okMeshId, meshId = decodeValue(properties.MeshId, instanceMap)
+
+		-- It is okay to provide default properties here, because 
+		-- they will be overriden on the update cycle
```
LPGhatguy, line 52:
```suggestion
		-- It is okay to provide default properties here, because 
		-- they will be set when properties are applied in reify.
```

LPGhatguy (changes requested):
PR looks good, thanks for the submission!

Can you back out the change to `database.json`? We should be able to land any updates to rbx_dom_lua in a separate PR and it is a frequent source of merge conflicts.

I mentioned in the review comments, but I think we should remove `requiresRecreate` from the patch type and instead calculate it during `applyPatch`. The Rojo server produces most of the patches that the plugin receives and doesn't have any knowledge of what would require an instance to be recreated. It's bets if we let the plugin decide for itself when it goes to apply a change.

Zodanim:
Not being able to sync in MeshParts is currently stopping me from being able to use Rojo efficiently, while this is pending is there any workaround that can be suggested?

sasial-dev:
It was closed in favour of #786.

PR #535, opened 2022-04-19
Add TestEZ std to selene

nezuo:
Almost all of the errors that are reported by selene relate to TestEZ. This PR reduces the error count from 175 down to 1.

### DIFF
--- a/selene.toml
+++ b/selene.toml
@@ -1,4 +1,4 @@
-std = "roblox"
+std = "roblox+testez"
 
 [config]
 unused_variable = { allow_unused_self = true }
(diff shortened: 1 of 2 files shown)


LPGhatguy (approved):
Nice!

Kampfkarren:
What's the 1? :eyes:

nezuo:
An `assert` doesn't have a 2nd argument:
https://github.com/rojo-rbx/rojo/blob/c7ab6c435c3b66b7fa857ebb5553f262f9fbbe01/plugin/src/ApiContext.lua#L88

PR #537, opened 2022-04-19
PluginActions for connecting/disconnecting a session

boatbomber:
Widget space is a precious resource on my 14 inch laptop, and the Rojo widget was wasting space for me most of the time. I simply pressed connect or disconnect and that was all. A simple keybinding would suffice for that common use case, but one didn't exist so I forked and added it.

I used a [PluginAction](https://developer.roblox.com/en-us/api-reference/class/PluginAction) so that users may go to `File > Advanced > Customize Shortcuts` in the Studio menu to set the keybinding. (We cannot set a default keybinding because Roblox still has not added that functionality, please support [this feature request](https://devforum.roblox.com/t/pluginactionassignedshortcut-and-4th-createpluginaction-parameter/126998/5?u=boatbomber).)

This PR adds a StudioPluginAction component, exposes the endSession function, and uses those to create plugin actions that start/end the session. Because it is done from within App, the keybinding will appropriately update the widget state.

There are three actions. One is a toggle- it will connect or disconnect, depending upon the current session state. The other two are connect and disconnect respectively. This allows the user to choose which to bind and decide if they prefer a single keybind for both actions or unique keybinds for each.
image

https://user-images.githubusercontent.com/40185666/164118642-c5cb56be-a215-4e00-a2d2-3166cf70f5df.mp4

--------

A few things before this PR is ready to merge:

- [x] Should respect settings
- [x] Should print/warn in the output upon successful connect/disconnect
- [x] Should connect to non-default host/port if user has typed into the widget (?)


unix-system:
Support for this! Would love to see merged.

As an extra convenience - perhaps an option in the Rojo settings that would create a [PluginToolbarButton](https://developer.roblox.com/en-us/api-reference/class/PluginToolbarButton) button on the Toolbar specifically for starting and stopping without having to open the full menu (?)

boatbomber:
> Support for this! Would love to see merged.
> 
> As an extra convenience - perhaps an option in the Rojo settings that would create a [PluginToolbarButton](https://developer.roblox.com/en-us/api-reference/class/PluginToolbarButton) button on the Toolbar specifically for starting and stopping without having to open the full menu (?)

@unix-system What would the button offer over the keybind? I personally almost never have the Plugin ribbon open on top, because I need to access the other buttons like Script and Test. Additionally, my Plugins ribbon is crowded enough as it is without adding another button that already has two alternatives.

unix-system:
@boatbomber Mostly a preference thing. For better or worse, I don't really use keybinds often in Studio. So assuming that's true for other users too, this would reduce the number of clicks required for me to start or stop Rojo.

(relevant)
[image](https://xkcd.com/1172/)

plugin/src/App/Components/Studio/StudioPluginAction.lua
```diff
@@ -0,0 +1,40 @@
...
+function StudioPluginAction:init()
+	self.pluginAction = self.props.plugin:CreatePluginAction(
+		self.props.name, self.props.title, self.props.description, self.props.icon, self.props.bindable
+	)
+
+	self.pluginAction.Triggered:Connect(self.props.onTriggered)
+end
+
+function StudioPluginAction:render()
+	return
```
LPGhatguy, line 23:
```suggestion
	return nil
```

We should explicitly return nil in Roact components that render nothing.

plugin/src/App/init.lua
```diff
@@ -138,11 +162,21 @@ function App:render()
...
+								local portText = self.portRef.current.Text
+
+								self:startSession(
+									if #hostText > 0 then hostText else Config.defaultHost,
+									if #portText > 0 then portText else Config.defaultPort,
+									{
+										openScriptsExternally = settings:get("openScriptsExternally"),
+										twoWaySync = settings:get("twoWaySync"),
+									}
+								)
```
LPGhatguy, line 179: This code is repeated a few times in this file and has gotten more complicated. Can we extract it somewhere else now that the logic has grown?

plugin/src/App/init.lua
```diff
@@ -65,6 +69,8 @@ function App:startSession(host, port, sessionOptions)
 				projectName = details,
 				address = address,
 			})
+
+			Log.warn(string.format("Connected to session '%s' at %s", details, address))
```
LPGhatguy, line 73: These warnings are helpful for users using the keybinds, but they probably should not be warnings.

I want to add a notification system for Rojo at some point. In the meantime, are the icons from #538 enough of an indicator? Can we remove these warnings?

LPGhatguy (changes requested):
Great PR, a couple comments and then we should be good to go!

plugin/src/App/init.lua
boatbomber: I'm not sure the best way to go about that given that it needs to consume `settings`.

plugin/src/App/init.lua
boatbomber: Removing them leads to zero indication for me, even with #538 because I don't leave the plugin ribbon open.

I could change them to Log.info instead of Log.warn, if that's better.

A notification system that puts popups into CoreGui would be nice as that would show regardless of widget or ribbon state.

plugin/src/App/init.lua
LPGhatguy: We can wrap this component in a function component that pulls out the plugin settings from context and passes them via props. `startSession` can be refactored to pull the values from the refs and settings directly, and we can just call `self:startSession()` in each of these cases.

plugin/src/App/init.lua
LPGhatguy: Dropping them to `Log.info` would be fine.

Agreed on notifications, I think that's where we would want to go eventually, but it's out of scope for this PR.

LPGhatguy (approved):
LGTM! Thank you!

PR #538, opened 2022-04-20
Add changing toolbar icon to indicate state

boatbomber:
The toolbar icon can act as a useful feedback mechanism about session status, especially when the widget is covered/closed.

I've added icon changeability to the StudioToggleButton component, and used it in the App.

image image image

The icons added to Assets.lua should preferably be reuploaded by @LPGhatguy for consistency before merging this.


LPGhatguy (approved):
This is a great, simple PR!

Can you put the images used for the icons into the `assets` folder as part of this PR?

boatbomber:
> Can you put the images used for the icons into the `assets` folder as part of this PR?

@LPGhatguy Even though those images aren't going to be used in the final result?

LPGhatguy:
> @LPGhatguy Even though those images aren't going to be used in the final result?

Yup! It's good to have assets available as files in case we need to reupload them for whatever reason. Eventually, maybe we'll use a tool to automatically manage our Roblox assets and keep them in sync with the repo.

PR #540, opened 2022-04-26
Add notification popups

boatbomber:
When #537 lands, users will be able to control Rojo without opening the widget or the toolbar. While this is excellent for developer experience, it leads to a total lack of feedback on their actions. As a stopgap, we added logs into the console for them, but that's not a great longterm solution.

I prototyped a notification system that aims to accomplish what @LPGhatguy and I discussed in #537. These show up in the viewport, so even if nothing Rojo-related is open the user will still receive these nice animated and themed popups.

Features:
- Rojo logo on the popup to make it clear where they came from
- Animated entry and exit
- Sound effect on entry
- Dynamic sizing to fit arbitrary notifications cleanly
- Timestamps on each notification, automatically handled
- Click to dismiss
- Timeouts to auto dismiss notifications a few seconds after being viewed to reduce friction
- Simple `App:addNotification(text: string, timeout: number?)` API so it can be easily integrated throughout
- Setting for enabling/disabling notifications (enabled by default)

https://user-images.githubusercontent.com/40185666/165400747-31543168-ef51-459d-8eb6-4be1a9be1e26.mp4


sasial-dev:
> * Won't notify if widget is enabled to avoid duplicate messages (except in the video below for ease of demo)
Consider adding a setting so you can still have notifications even if the widget is open!

plugin/src/Assets.lua
```diff
@@ -45,6 +45,9 @@ local Assets = {
 			[500] = "rbxassetid://2609138523"
 		},
 	},
+	Sounds = {
+		Notification = "rbxassetid://9119690001",
```
sasial-dev, line 49: Consider replacing with a [much nicer sound](https://www.roblox.com/library/255881176); that imo, makes more sense UX-wise!
```suggestion
		Notification = "rbxassetid://255881176",
```

boatbomber: It has to be something we have the right to use, not someone reuploading a Windows sound.

sasial-dev: Ahhh, far point!

boatbomber:
Added setting for notifs, should be nicer to give users that control

LPGhatguy (approved):
Let's do it!

PR #541, opened 2022-05-03
Bugfix: PluginAction spam causing errors

boatbomber:
Fixes the following issues that I missed in #537:
- [x] Actions checked session's existence rather than session's status, causing issues with actions during connecting status and other non-valid states
- [x] Actions during pages other than NotConnected would nil index fail because the hostRef/portRef weren't being rendered
- [x] In the refactor of the connect function, accidentally lost the host/port text being retained between page renders

My sincere apologies!

### DIFF
--- a/plugin/src/App/StatusPages/NotConnected.lua
+++ b/plugin/src/App/StatusPages/NotConnected.lua
@@ -14,8 +14,6 @@ local PORT_WIDTH = 74
 local DIVIDER_WIDTH = 1
 local HOST_OFFSET = 12
 
-local lastHost, lastPort
-
 local e = Roact.createElement
 
 local function AddressEntry(props)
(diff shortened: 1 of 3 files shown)


plugin/src/App/StatusPages/NotConnected.lua
```diff
@@ -90,6 +88,8 @@ function NotConnectedPage:render()
 		}),
 
 		AddressEntry = e(AddressEntry, {
+			hostText = self.props.hostText,
+			portText = self.props.portText,
```
LPGhatguy, line 92: This may cause the current value the user has typed in to be wiped out if Roact triggers a re-render here. We might want to switch this input to have a controlled value instead of using a ref.

LPGhatguy (approved):
Thanks for filing a follow-up fix! I put in one extra fix to preserve the address and port fields even if the user doesn't press connect.

PR #544, opened 2022-05-23
Fix sourcemap command not stripping paths correctly

filiptibell:
Simple fix to handle `rojo sourcemap` when a project file is not given correctly. Previous behavior was removing one level of directories too much inside the sourcemap if the project file path was not given.


LPGhatguy (approved):
Awesome, thank you for the timely fix!

PR #545, opened 2022-05-24
Bugfix: Settings not working

boatbomber:
When changing a setting, the UI updates (checkbox will fill/empty) but the **setting does not actually change** or take effect.
This is because the settings page and the app are consuming two different settings tables in memory and so the settings page doesn't impact the table that the app will actually use.
This is an slip during the refactor of App in #537, as we made the functional wrapper create settings and accidentally left another in stateful render.

A very simple fix- just remove the old unused duplicate, and we're good to go.

If the PR seems large, view the diff ignoring whitespace as I had to alter the indentation of a lot of components that are no longer nested in the provider.

### DIFF
--- a/plugin/src/App/init.lua
+++ b/plugin/src/App/init.lua
@@ -154,150 +154,146 @@ function App:render()
 		value = self.props.plugin,
 	}, {
 		e(Theme.StudioProvider, nil, {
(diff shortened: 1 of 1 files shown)


LPGhatguy (approved):
LGTM, thank you for the fix!

PR #548, opened 2022-05-28
Major Performance Improvements

LPGhatguy:
This PR relies on stuff currently unreleased in rbx-dom libraries. All of these fixes combined should result in a 3-4x speedup for large Rojo projects.

### DIFF
--- a/src/snapshot/patch_apply.rs
+++ b/src/snapshot/patch_apply.rs
@@ -1,6 +1,9 @@
 //! Defines the algorithm for applying generated patches.
 
-use std::collections::HashMap;
+use std::{
+    collections::{HashMap, HashSet},
+    mem::take,
+};
 
 use rbx_dom_weak::types::{Ref, Variant};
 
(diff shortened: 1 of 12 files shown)


src/snapshot/patch_apply.rs
```diff
@@ -117,24 +117,24 @@ fn apply_add_child(
...
-        .metadata(snapshot.metadata)
-        .snapshot_id(snapshot.snapshot_id);
-
-    let id = tree.insert_instance(parent_id, remaining_snapshot);
+    let children = take(&mut snapshot.children);
+
+    // If an object we're adding has a non-null referent, we'll note this
+    // instance down as needing to be revisited later.
+    let has_refs = snapshot.properties.values().any(|value| match value {
+        Variant::Ref(value) => value.is_some(),
```
Kampfkarren, line 128: `matches!(value, Variant::Ref(Some(_))` Clippy lints this, but it might not check is_some() checks

LPGhatguy: I think this is a clippy false positive. `value` is not an `Option`, it's actually just a `Ref`, which has `is_some` and `is_none` methods.

Kampfkarren: Oh, I meant that clippy *should* catch this, under my assumption it was an Option. My mistake.

LPGhatguy: That's probably an indication that these method names are bad, but I didn't know what else to call them, lol!

Issue #499, opened 2021-11-08
Support for .luau extension


GlowingUmbreon:
Recently the LuaU repository has changed the primary extension to .luau (see: Roblox/luau#97).
It may take a while for the .luau extension to pick up support in the development community but now since it is now the primary extension for the luau repository it would be ideal to add support for it in rojo.

LPGhatguy:
Let's plan to have this in right after 7.0 stable!

insyri:
Is there a tracked PR for this issue? If not, I can pick this up across the Rojo ecosystem.

Kampfkarren:
7.0.0 is out, so you should be able to just go for it.

ambers-careware:
Any updates? Writing a new library with heavy use of strict mode / type annotations, and was wanting to use .luau to make the file type distinct from .lua

imacodr:
Should be coming next update.

LPGhatguy:
Fixed by #552.

PR #552, opened 2022-06-12
Accept .luau files

imacodr:
This is really just a tracked PR to start accepting .luau files to the Rojo ecosystem as the language grows to separate itself a bit from Lua 5.

Solves #499


LPGhatguy:
Hello, thank you for the PR!

This PR looks like it adds support for the "Open Scripts in Studio" feature for .luau files only -- Rojo otherwise still will not recognize `.luau` files. I think we can put a fix together for that pretty easily, though!

imacodr:
I'm still figuring out a bit of the Rojo ecosystem for additional changes without breaking everything 😂 

I'll be looking out for new changes to add to it for Luau support.

imacodr (commented):
Serving unit tests are failing although I'm taking a closer look why. Tested in-studio and it seems to now sync as .luau

LPGhatguy (approved):

Issue #555, opened 2022-06-13
Not showing correct address on serve on the CLI


WatermelonArray:
I have been using Rojo for a network configuration. Using `rojo serve`, I can specify an address for Rojo to host to. In my case, my local machine's IPv4 address instead of hosting it on `localhost`.

However when successfully hosting to your specified address, the CLI displays incorrect information about where the host is sending data to - it still shows: `Visit http://localhost:34872 in your browser for more information`

This is does not effect my workflow at all but still is a **minor** bug.

Here is an image of the bug using Linux CLI binary Version 7.1.1:
![2022-06-13-123125_1920x1080_scrot](https://user-images.githubusercontent.com/34957838/173347286-db0cd78f-fac7-4ea0-8e1a-e1b8b264ccfc.png)

This should display:
`Visit http://192.168.1.226:34872 in your browser for more information`

---

NOTE:
The port number changes when you specify a port value on the command, but not the address.
![2022-06-13-125831_1920x1080_scrot](https://user-images.githubusercontent.com/34957838/173348040-3c0c914d-00bf-4688-a1d1-170e12f67932.png)

EDIT: Because you are hosting to an address, it is not hosting to localhost. If you try to connect to `localhost:348721`, you simply are unable to establish a connection.

LPGhatguy:
Thank you for the bug report, this is a great catch!

This should be a straightforward fix in `cli/serve.rs`.

PR #556, opened 2022-06-14
Added address reference to CLI output

WatermelonArray:
This pull request just fixes the CLI bug on rojo-rbx/rojo#555

Here is a screenshot of the changes on the output
2022-06-14-130402_1920x1080_scrot

### DIFF
--- a/src/cli/serve.rs
+++ b/src/cli/serve.rs
@@ -88,7 +88,14 @@ fn show_start_message(bind_address: IpAddr, port: u16, color: ColorChoice) -> io
     write!(&mut buffer, "Visit ")?;
 
     buffer.set_color(&green)?;
-    write!(&mut buffer, "http://localhost:{}/", port)?;
+    if bind_address.is_loopback() {
+        write!(&mut buffer, "http://localhost:")?;
+    } else {
+        write!(&mut buffer, "http://{}:", bind_address)?;
+    }
+
+    buffer.set_color(&green)?;
+    write!(&mut buffer, "{}/", port)?;
 
     buffer.set_color(&ColorSpec::new())?;
     writeln!(&mut buffer, " in your browser for more information.")?;
--- a/src/cli/serve.rs
+++ b/src/cli/serve.rs
@@ -66,8 +66,12 @@ fn show_start_message(bind_address: IpAddr, port: u16, color: ColorChoice) -> io
 
     let writer = BufferWriter::stdout(color);
     let mut buffer = writer.buffer();
-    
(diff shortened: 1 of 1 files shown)


src/cli/serve.rs
```diff
@@ -88,7 +88,14 @@ fn show_start_message(bind_address: IpAddr, port: u16, color: ColorChoice) -> io
     write!(&mut buffer, "Visit ")?;
 
     buffer.set_color(&green)?;
-    write!(&mut buffer, "http://localhost:{}/", port)?;
+    if bind_address.is_loopback() {
+        write!(&mut buffer, "http://localhost:")?;
```
Kampfkarren, line 92: The same loopback check is done to display the address earlier. You would be better off storing the address in a variable that is used for both displays IMO.

WatermelonArray:
@Kampfkarren Thanks for the feedback, all sorted!

src/cli/serve.rs
```diff
@@ -66,6 +66,8 @@ fn show_start_message(bind_address: IpAddr, port: u16, color: ColorChoice) -> io
 
     let writer = BufferWriter::stdout(color);
     let mut buffer = writer.buffer();
+    
+    let address_string = if bind_address.is_loopback() {"localhost"} else {stringify!(bind_address)};
```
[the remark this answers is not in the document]
WatermelonArray: Ah, I tried doing that before but my rust linter was saying that .to_string() wasn't a function in the struct.
Turns out my linter is slow and didn't catch up when I was trying to debug it! So I just tried using stringify!() and the problem went away - forgot that stringify!() does not do what I was trying to.

Anyway, sorted!

src/cli/serve.rs
```diff
@@ -88,7 +90,10 @@ fn show_start_message(bind_address: IpAddr, port: u16, color: ColorChoice) -> io
     write!(&mut buffer, "Visit ")?;
 
     buffer.set_color(&green)?;
-    write!(&mut buffer, "http://localhost:{}/", port)?;
+    write!(&mut buffer, "http://{}:", address_string)?;
+
+    buffer.set_color(&green)?;
+    write!(&mut buffer, "{}/", port)?;
```
LPGhatguy, line 96: Let's merge these calls together. You can pass multiple arguments into the same format string, something like this should work:

```rust
write!(&mut buffer, "http://{}:{}/", address_string, port)?;
```

LPGhatguy (changes requested):
Hello, thanks for submitting a PR! I just got back from vacation, submitted a little feedback and approved the CI workflows to run.

PR #558, opened 2022-06-20
Update rbx_binary to 0.6.5, fixes #557

### LINKED ISSUE #557: Rojo fails to sync serialized packages from remodel - "Unsupported property type: PackageLink.PackageId is of type Content"
**tl;dr** I think we can fix this issue by updating rbx_binary from 0.6.4 to 0.6.5. 

## The problem
I noticed that when saving a package in a Place using remodel, it saves fine but when building the place with rojo, it runs into an error:
```
[ERROR rojo] Unsupported property type: PackageLink.PackageId is of type Content
```
## rbx_binary fixed this back in April
FilamentGames fixed this back in April for rbx_binary with [this PR](https://github.com/rojo-rbx/rbx-dom/pull/215). That was pulled into remodel by [a commit updating remodel dependencies on June 13th](https://github.com/rojo-rbx/remodel/commit/4322a1a5d00da2966809665f53fb5430c2f555fe) in v0.10.0. Remodel now doesn't have this problem. 

## Rojo still has the problem due to old dependencies
However, rojo still has this problem. While #548 updated the contents of rbx_dom_weak from 2.3.0 to 2.4.0 on June 5th, rbx_binary still is on version 0.6.4, [having last been updated on October 11th 2021](https://github.com/rojo-rbx/rojo/commit/8d29b4315570f66d04236cd660c23ab79e927eac).  I believe it's likely that updating rbx_binary to 0.6.5 will fix this by pulling in [the fix PR](https://github.com/rojo-rbx/rbx-dom/pull/215).

## Updating the dependency seems to fix the problem. 
On a local development copy, updating the rbx_binary dependency of Rojo to 0.6.5 seems to fix the problem (or at least the error message doesn't show up). I can't get some of the tests running locally though on the base due to timeouts. I think this might just be a problem with my local setup though. 

I'm going to further verify the fixes locally. I can submit a PR, otherwise feel free to just bump the version.

schmatz:
Fixes #557. I can't seem to get the `cargo test` command passing on either the base or this commit. I suspect that issue is related to my local development setup, and if CI is green it should be fine. I also have tested the debug build in my workflow and am no longer seeing the same errors that I was in that issue.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1639,12 +1639,13 @@ dependencies = [
 
 [[package]]
 name = "rbx_binary"
-version = "0.6.4"
+version = "0.6.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "68f424adb7a0a24ab4bd153be141035f1404ae40affed902fd2721b42cca7f86"
+checksum = "28fe644ede286801c9bd59527b332d69449c6e9eaefa2277a0e776252097bc90"
 dependencies = [
  "log",
  "lz4",
+ "profiling",
  "rbx_dom_weak",
  "rbx_reflection",
  "rbx_reflection_database",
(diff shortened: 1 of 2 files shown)


LPGhatguy:
Thanks for the PR, this should be an easy merge and release!

LPGhatguy (approved):

PR #566, opened 2022-06-30
Change Notification sound to generic sound

imacodr:
The notification sound for the plugin notifications causes the game to summon an error due to no experience permissions with no way to grant permission. This is due to the new audio policy update.

### DIFF
--- a/plugin/src/Assets.lua
+++ b/plugin/src/Assets.lua
@@ -46,7 +46,7 @@ local Assets = {
 		},
 	},
 	Sounds = {
-		Notification = "rbxassetid://9716079936",
+		Notification = "rbxassetid://8747340426",
 	},
 	StartSession = "",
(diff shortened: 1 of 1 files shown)


boatbomber:
I think the sound file in assets/ should be replaced as well to create consistency between our repo assets and the used assets.

LPGhatguy:
I don't know if this sound fits. What other notification sounds are available that will work?

boatbomber:
How about [203785492](https://www.roblox.com/library/203785492/Pop-Close-Sound)?

LPGhatguy:
> How about [203785492](https://www.roblox.com/library/203785492/Pop-Close-Sound)?

Works for me!

imacodr:
Sounds good

LPGhatguy (approved):
LGTM.

PR #568, opened 2022-07-01
Add setting to toggle sound effects

boatbomber:
Title says it all, really. Adds a `playSounds` setting and makes the sound player object ingest & respect settings.
image


plugin/src/soundPlayer.lua
```diff
@@ -0,0 +1,39 @@
...
+	return setmetatable({}, SoundPlayer)
+end
+
+function SoundPlayer.withSettings(settings)
+	return setmetatable({
+		settings = settings,
+	}, SoundPlayer)
+end
+
+function SoundPlayer:Play(soundId)
```
LPGhatguy, line 25: We should be consistent about our casing, Rojo is generally `camelCase`.

plugin/src/soundPlayer.lua
```diff
@@ -0,0 +1,39 @@
...
+))
+widget.Name = "Rojo_soundPlayer"
+widget.Title = "Rojo Sound Player"
+
+local SoundPlayer = {}
+SoundPlayer.__index = SoundPlayer
+
+function SoundPlayer.new(settngs)
+	return setmetatable({}, SoundPlayer)
+end
```
LPGhatguy, line 17: This constructor is never used, so we can save a bit of complexity by removing this, and then assuming that we'll always have settings.

LPGhatguy (changes requested):
Thanks for the quick PR! Couple pieces of feedback, then LGTM.

LPGhatguy (approved):

Issue #418, opened 2021-04-27
Rojo Plugin should offer real-time status about sync details


BetterThanReal:
It's not terribly uncommon for developers to assume that Rojo is running, connected, and working properly, even when it's not.  A great example is when resuming Roblox Studio usage after the computer has been asleep for awhile - Rojo sometimes silently fails to sync, and the user continues to develop as though their latest changes are being synced to Studio.

I personally have lost a cumulative total of at least one hour troubleshooting my code with the assumption that Rojo was syncing properly, when it wasn't.

An easy way to help Rojo be more communicative about the state of the plugin is if the plugin reports, in real-time, the status of the latest file sync that resulted in changed files.

Suggested plugin status would read something like "5 files changed 32 seconds ago", where "32 seconds" would be a counter that updates once per second until at least one minute has passed, at which point the text would change to "5 files changed 1 minute ago", and again changing when the hour and day thresholds have been exceeded.

OverHash:
I think updating once per second would be a bit excessive and distracting. Maybe something like "A few seconds ago..." for <1m, and then you can increment once per minute after that? (Or maybe even less than once per minute).

Alternatively, maybe you could bind to window changes and refresh then? Not sure if that's possible with API exposed to plugins.

BetterThanReal:
> I think updating once per second would be a bit excessive and distracting.

The goal of updating once per second (during the first minute only) is to prevent circumstances where the developer makes a change, comes back to Studio to test, makes a change, comes back to Studio to test, and effectively wastes a minute of their time because Rojo is not syncing.

When I'm troubleshooting, sometimes it's not uncommon for me to make and test five changes per minute.  While I agree that wasting a whole minute is trivial compared to wasting an hour, it's still something I'd like to avoid.

FWIW, the Roblox Studio Diagnostics Bar shows updates 4 or 5 times per second, and that doesn't distract me at all.

PR #569, opened 2022-07-03
Real-time status about sync details

boatbomber:
Resolves https://github.com/rojo-rbx/rojo/issues/418.

image

Features:
- Human-readable timestamp text (just now, 3 seconds ago, 1 minute ago, 2 minute**s** ago)
- How many changes were applied by the latest patch
- Timestamp text live updates (and slows the update freq as it gets older)
- Combines patches that are within a second of each other, to more readably handle burst changes


OverHash:
Do users need to know how many changes have occurred in a given time period? I feel like just displaying "last synced just now" or "last synced 5 minutes ago" is enough.

boatbomber:
> Do users need to know how many changes have occurred in a given time period? I feel like just displaying "last synced just now" or "last synced 5 minutes ago" is enough.

@OverHash  The #418 issue specifically asked to know a count, and no one disagreed. That said, I do think that a simple last since readout would solve the most common use case put forth in the issue- it informs you that the plugin is still working. However, it's possible that there could be cases where it's only syncing partially, which wouldn't be exposed by just a timestamp.

plugin/src/App/init.lua
```diff
@@ -123,9 +124,15 @@ function App:startSession()
 		end
 
 		if count == 0 then return end
+
+		local old = self.patchInfo:getValue()
+		if now - old.timestamp < 2 then
+            count += old.updates
```
Kampfkarren, line 130: Weird indentation

boatbomber: Wrote it on mobile, sorry! I'll fix it when I get home later tonight.

Logic-Mastermind (approved):
I see nothing that could be improved on in terms of code, I think this should be merged as soon as possible.

plugin/src/App/StatusPages/Connected.lua
```diff
@@ -13,6 +13,26 @@ local BorderedContainer = require(Plugin.App.Components.BorderedContainer)
 
 local e = Roact.createElement
 
+local AGE_UNITS = { {31556909, "year"}, {2629743, "month"}, {604800, "week"}, {86400, "day"}, {3600, "hour"}, {60, "minute"}, }
```
LPGhatguy, line 16: "last synced one year ago" would be a pretty funny message to happen in practice

LPGhatguy (approved):
Very clean, thank you!

Issue #138, opened 2019-03-26
[Request] Option to let the Studio plugin display updates in console


Validark:
It would be nice to have a feature built-in to the plugin whereby a developer could see in the Roblox console how Rojo is updating files. I am on `Rojo 0.5.0-alpha.6`

LPGhatguy:
It should be pretty straightforward to expand/curate the log output to be more useful for users of Rojo. It'd also be great to have documentation for how to _get_ at that log output!

For the time being, to get the maximum amount of debug information, you can:

1. Run the Rojo server with `RUST_LOG=librojo=trace rojo serve`, which will dump tons of diagnostics to the terminal
2. Run `_G.ROJO_DEV_CREATE()` in your Studio command bar, which will create a Rojo developer settings object in your game. By default, it will tell the plugin to be as verbose as possible.

LPGhatguy:
Starting in Rojo 6, we've increased the verbosity of logging when `_G.ROJO_DEV_CREATE()` is called. From here, it's just a matter of documenting this feature. We should also expose this in the plugin settings,.

PR #572, opened 2022-07-06
Migrate DevSettings to PluginSettings for much better config flow

boatbomber:
Resolves https://github.com/rojo-rbx/rojo/issues/138.

The current method of setting config/debug settings (value instances through _G injection) is really clunky and 3 years old. We have a nicer setting interface now, so let's use it! Migrating DevSettings to PluginSettings for an improved UX.

dropdown

To Do:
- [x] Add config settings to plugin settings
- [x] Create dropdown component
    - [x] Basic dropdown functionality
    - [x] Animations when opening/closing
    - [x] Clear contrast and border on option container
    - [x] Static size to avoid jarring relayout
    - [x] Readable padding
- [x] Settings UI Refactor
    - [x] Decouple settings UI from checkboxes, allow for dropdowns
    - [x] Add reset to default option to settings UI
- [x] Add setting UI for new configs
- [x] Make the plugin actually respect the new settings
    - [x] Use logLevel
    - [x] Use typecheckingEnabled
    - [x] Redo the above using the settings method from #576
- [x] Update the test runners to use the new Settings
    - [x] Improve failing test output
    - [x] Fix failing test


plugin/src/App/StatusPages/Settings/Setting.lua
```diff
@@ -53,22 +56,48 @@ function Setting:render()
...
-				position = UDim2.new(1, 0, 0.5, 0),
-				anchorPoint = Vector2.new(1, 0.5),
-				onClick = function()
-					local currentValue = Settings:get(self.props.id)
-					Settings:set(self.props.id, not currentValue)
-				end,
-			}),
+				visible = self.props.showReset,
+
+				position = UDim2.new(1, -32 - (self.props.options ~= nil and 120 or 40), 0.5, 0),
```
Boegie19, line 96: 120 is a lot of padding between the text and the options I think 80 is better since then you have the minimum padding needed.
since if we where to try and fix #592 then the padding would become an issue

boatbomber: Padding is there to support the reset button and its hover circle.
image

Boegie19: 2 Question 
why is the hover circle needed would it not be better to auto update the log level on change of the options?

the other one is you could probebly remove a little bit of padding still we could also do that once we fix the size issues later

boatbomber: I think you misunderstand the point of that button. Of course we auto update the log level when you pick an option. When you have a non-default option selected, this button appears and clicking it resets the setting back to default for you.

LPGhatguy (approved):

PR #574, opened 2022-07-08
Simplify usage of attributes.

MaximumADHD:
# Summary

This commit deprecates using the `Attributes` property in `$properties`, in favor of a direct `$attributes` field in the JSON instances (or `attributes` in `*.model/meta.json` files). Unambiguous value types like `string`, `boolean`, and `number` can be written as fields of `$attributes` directly.

# The Problem

In trying to clean up the existing structure for writing attributes (i.e. adding unambiguous value types), it became apparent the following case would cause problems with how rojo handles explicit types:

```json
{
  "$properties": {
    "Attributes": {
      "String": "Test"
    }
  }
}
```

Rojo interprets the value of this object's attributes to be an explicit string `"Test"`, rather than a single attribute named `"String"` with `"Test"` as its value. Working around this requires the value of test to be encapsulated as so:

```json
{
  "$properties": {
    "Attributes": {
      "String": {
        "String": "Test"
      }
    }
  }
}
```

The solution is to explicitly make `$attributes` a separate field:

```json
{
  "$attributes": {
    "String": "Test"
  }
}
```

This not only looks cleaner, but it also clears up any ambiguity since `$attributes` is explicitly `HashMap<String, Variant>`.

# Compatibility

This change is forward compatible with the existing use of the `"Attributes"` property.
It just requires noisy explicit typing to work as expected.


LPGhatguy (changes requested):

src/resolution.rs
```diff
@@ -22,10 +23,17 @@ pub enum UnresolvedValue {
 }
 
 impl UnresolvedValue {
-    pub fn resolve(self, class_name: &str, prop_name: &str) -> anyhow::Result<Variant> {
+    pub fn resolve_explicit(self, class_name: &str, prop_name: &str) -> anyhow::Result<Variant> {
```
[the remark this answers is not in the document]
MaximumADHD: Maybe `resolve_unambiguous` or `resolve_primitive`?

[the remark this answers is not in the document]
MaximumADHD: `resolve_unambiguous` it is!

src/resolution.rs
```diff
@@ -130,7 +139,16 @@ impl AmbiguousValue {
                     Ok(CFrame::new(pos, orientation).into())
                 }
 
-                (VariantType::Attributes, AmbiguousValue::Attributes(value)) => Ok(value.into()),
+                (VariantType::Attributes, AmbiguousValue::UnresolvedValueMap(value)) => {
+                    let mut resolved = Attributes::new();
+
+                    for (name, unresolved) in value {
+                        let value = unresolved.resolve_implicit(&name);
+                        resolved.insert(name.into(), value.unwrap());
```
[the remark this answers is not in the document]
MaximumADHD: Seems to have been handled through `with_context` in the newest revision.

src/resolution.rs
```diff
@@ -148,6 +155,16 @@ impl AmbiguousValue {
         }
     }
 
+    pub fn resolve_unambiguous(self) -> anyhow::Result<Variant> {
+        match self {
+            AmbiguousValue::Bool(value) => Ok(value.into()),
+            AmbiguousValue::Number(value) => Ok(value.into()),
+            AmbiguousValue::String(value) => Ok(value.into()),
+
+            _ => bail!("Cannot unambiguously resolve type!"),
```
LPGhatguy, line 164:
```suggestion
            other => bail!("Cannot unambiguously resolve the value {other:?}"),
```

src/resolution.rs
```diff
@@ -218,12 +235,21 @@ mod test {
...
+    }
+
     #[test]
     fn bools() {
         assert_eq!(resolve("BoolValue", "Value", "false"), Variant::Bool(false));
 
         // Script.Disabled is inherited from BaseScript
         assert_eq!(resolve("Script", "Disabled", "true"), Variant::Bool(true));
+
+        // Check unambiguous boolean values
```
LPGhatguy, line 250:
```suggestion
```

This comment doesn't convey any new information, we can axe it.

src/resolution.rs
```diff
@@ -247,6 +273,12 @@ mod test {
         //     resolve("Folder", "Tags", "\"a\\u0000b\\u0000c\""),
         //     Variant::BinaryString(b"a\0b\0c".to_vec().into()),
         // );
+
+        // Check unambiguous string value
```
LPGhatguy, line 277:
```suggestion
```

src/resolution.rs
```diff
@@ -257,12 +289,14 @@ mod test {
...
         assert_eq!(
-            resolve("Folder", "SourceAssetId", "532413"),
+            resolve("IntValue", "Value", "532413"),
             Variant::Int64(532413),
         );
 
         assert_eq!(resolve("Part", "Transparency", "1"), Variant::Float32(1.0));
         assert_eq!(resolve("NumberValue", "Value", "1"), Variant::Float64(1.0));
+
+        assert_eq!(resolve_unambiguous("1.337"), Variant::Float64(1.337));
```
LPGhatguy, line 299:
```suggestion
        assert_eq!(resolve_unambiguous("12.5"), Variant::Float64(12.5));
```

When checking for exact floating point equality, we should use values that can definitely be resolved exactly to prevent futures issues related to rounding.

LPGhatguy (approved):

PR #576, opened 2022-07-09
Use singleton settings outside the Roact tree

boatbomber:
Our settings are currently tied into the Roact tree, which can cause pain points when trying to change settings from outside the tree (ex: test/dev configs, api settings, etc).

This PR removes PluginSettings and replaced it with Settings, a singleton with a few handy APIs:
- `Settings:get(id: string)` Gets the current value of a setting
- `Settings:set(id: string, value: any)` Sets the value of a setting
- `Settings:onChanged(id: string, callback: func)` Listens to changes of a setting

It will automatically save/load if run in a plugin environment, but won't break if run elsewhere (such as a test env).

Also, it makes use of Log.trace for a better debugging experience for maintainers.


boatbomber:
Running into a strange issue- only errors sometimes, and still functions correctly even when it errors?

`Roact.createSignal:52: Listeners can only be disconnected once.`, triggers from a :setState call in :init.
It still sets the state and renders+animates correctly so I'm confused by this.
```Lua
function Setting:init()
    self:setState({
        setting = Settings:get(self.props.id),
    })

    Settings:onChanged(self.props.id, function(value)
        self:setState({ -- This is where the error comes from
            setting = value,
        })
    end)
```

This is the only place that state gets used
```Lua
Checkbox = e(Checkbox, {
    active = self.state.setting,
```

boatbomber:
I think I figured out the "sometimes" aspect of it- seems to error the first two times a checkbox is clicked, then works fine for that one. Rinse and repeat for each checkbox.

boatbomber:
For some reason, it appears that :willUnmount is not being called on the setting component.

boatbomber:
I... wrote :willUnmount on the wrong table. Thanks, autocomplete. Fixed now.

plugin/src/Settings.lua
```diff
@@ -0,0 +1,100 @@
...
+	local bind, set = Roact.createBinding(self._values[name])
+	self._bindings[name] = {
+		bind = bind,
+		set = set,
+	}
+
+	Log.trace(string.format("Created binding for setting '%s'", name))
+
+	return bind
+end
```
chriscerie, line 98:
```suggestion
```

plugin/src/Settings.lua
```diff
@@ -0,0 +1,100 @@
...
+	twoWaySync = false,
+	showNotifications = true,
+	playSounds = true,
+}
+
+local Settings = {}
+
+Settings._values = table.clone(defaultSettings)
+Settings._updateListeners = {}
+Settings._bindings = {}
```
chriscerie, line 22:
```suggestion
```

plugin/src/Settings.lua
```diff
@@ -0,0 +1,100 @@
...
+	end
+
+	return self._values[name]
+end
+
+function Settings:set(name, value)
+	self._values[name] = value
+	if self._bindings[name] then
+		self._bindings[name].set(value)
+	end
```
chriscerie, line 51:
```suggestion
```

plugin/src/Settings.lua
```diff
@@ -0,0 +1,100 @@
+--[[
+	Persistent plugin settings that can be accessed via Roact bindings.
```
chriscerie, line 2:
```suggestion
	Persistent plugin settings.
```

chriscerie (changes requested):
It looks like we no longer need these since we moved bindings to state.

plugin/src/Settings.lua
boatbomber: They could be handy later, I don't see the harm in keeping this around. The only reason we went with state was for animation, but non-animated use cases later might be better off with bindings

chriscerie: The only reason you'd pick bindings over state is to optimize rapidly changing updates. Otherwise state is much less error prone.

boatbomber: Fair enough. I suppose if the need arises we can just add it back. Settings should hopefully not be changed that rapidly.

plugin/src/Settings.lua
```diff
@@ -0,0 +1,78 @@
...
+Settings._values = table.clone(defaultSettings)
+Settings._updateListeners = {}
+
+if plugin then
+	for name, defaultValue in pairs(Settings._values) do
+		local savedValue = plugin:GetSetting("Rojo_" .. name)
+
+		if savedValue == nil then
+			-- plugin:SetSetting hits disc instead of memory, so it can be slow. Spawn so we don't hang.
+			task.spawn(plugin.SetSetting, plugin, "Rojo_" .. name, defaultValue)
```
LPGhatguy, line 28: I don't think `plugin.SetSetting` yields. If it does blocking I/O, then spawning will help Rojo initialize a little quicker, but Studio will still pause while the settings are actually loaded.

LPGhatguy (approved):
Looks great to me, thank you for doing this work!

PR #580, opened 2022-07-16
Add References syntax to `.model.json` files

sasial-dev:
This PR will add references syntax to `.model.json` files.


sasial-dev:
@LPGhatguy, looking for your thoughts on this: is it worth using `snapshot_id` for this? or using a new field? As I noticed that it needs some fixes 🤔 

https://github.com/rojo-rbx/rojo/blob/5bf1f868863adac49a16ad32748a61d324ace592/src/snapshot/instance_snapshot.rs#L20-L22

sasial-dev:
Once you let me know what you'd rather, I'll finish up this PR 🙂

PR #581, opened 2022-07-17
Really default to the current directory in 'rojo fmt-project'

wackbyte:
I noticed `fmt-project` didn't actually default to the current directory like it said it would.


LPGhatguy (approved):

LPGhatguy:
Thank you!

Issue #579, opened 2022-07-11
Plugin throws warning even if `rojo serve` is stopped after disconnecting from plugin


Sleitnick:
While disconnecting Rojo from an ongoing connection, the expected steps would be to disconnect the plugin and then to stop serving. However, a connection failure warning is shown when doing this.

This appears to have no behavioral impact on Rojo.

**Repro steps:**
1. `rojo serve` (or VSCode extension)
2. Connect via Rojo plugin in Studio
3. Disconnect from the plugin
4. Stop the `rojo serve` process
5. After a few seconds, observe the warning shown in the plugin and Studio's output window

Warnings shown in Roblox Studio output window (usually one or the other):
```
[Rojo-Warn] Disconnected from an error: Unknown HTTP error: HttpError: NetFail - Log:52
```
```
[Rojo-Warn] Disconnected from an error: Couldn't connect to the Rojo server.
[Rojo-Warn] Make sure the server is running — use 'rojo serve' to run it!
```

Warnings shown in plugin:
![rojo_warning](https://user-images.githubusercontent.com/5747014/178333873-25ae037d-be3f-40c1-bd43-267811c5f76b.png) ![rojo_warning_2](https://user-images.githubusercontent.com/5747014/178334334-79e3a3f8-4c0b-41c4-87b9-a6f12c9690ad.png)

**Expected behavior:**
I assume the expected behavior is for total silence. No warnings/errors/messages should be shown, other than perhaps a disconnect notification.

boatbomber:
Fixed! This issue can be closed now.

PR #587, opened 2022-07-22
Fix errors after session already ended

boatbomber:
Fixes issue #579.

To debug, I put a traceback into `__stopInternal` and performed the repro steps from issue #579. I saw where the call came from, and then added an if statement there (the API promise) to prevent it from firing stop after the session is already stopped.
After this change, performing the repro steps leads to no errors or warnings as expected.


LPGhatguy (approved):
Thank you!

PR #590, opened 2022-07-24
Add sync locking for Team Create

boatbomber:
Rojo does not do well in Team Create. This is a reasonable limitation, but inexperienced users often do not realize and cause conflicts and headaches for themselves and their coworkers. This PR adds a simple locking system when using Rojo in Team Create- only one person is allowed to sync at any given time. This prevents users from accidentally shooting their foot.

image


plugin/src/App/init.lua
```diff
@@ -87,7 +90,65 @@ function App:getHostAndPort()
...
+		return true
+	end
+
+	local lock = ServerStorage:FindFirstChild("Rojo_SessionLock")
+	if not lock then
+		lock = Instance.new("ObjectValue")
+		lock.Name = "Rojo_SessionLock"
+		lock.Archivable = false
+		lock.Value = Players.LocalPlayer
+		lock.Parent = ServerStorage
```
sasial-dev, line 105: ServerStorage does not feel like the right place to put this...

sasial-dev: @boatbomber, if people sync into serverstorage, it'll automatically be overwritten by rojo - defeating the point?

boatbomber: We've been asking Roblox to give us PluginStorageService for years, but it still hasn't happened 🤷‍♂️

boatbomber:
Went back to draft for now because it needs more design work to account for the points @sasial-dev brought up.

boatbomber:
Requirements for the lock container:
- Must replicate in Team Create (not all do, surprisingly)
- Must be out of the user's way

Also, I'll need to special case it to never get removed by a sync as an "unknown instance".

boatbomber:
Okay,  it works in this state. Using ServerStorage is ugly (no denying that) but it is common practice (see: tag editor) for Roblox Studio plugins because we do not yet have a better alternative. Support this feature request please!
https://devforum.roblox.com/t/add-a-pluginstorage-service-for-plugins-to-store-instances-in/224012/

plugin/src/Reconciler/diff.lua
```diff
@@ -113,6 +113,13 @@ local function diff(instanceMap, virtualInstances, rootId)
 			local childId = instanceMap.fromInstances[childInstance]
 
 			if childId == nil then
+				if childInstance.Name == "__Rojo_SessionLock" then
```
LPGhatguy, line 116: I wonder if we should ignore everything that has `Archivable = false`. This seems like a fine workaround for now probably.

boatbomber: I can't think of a case where you'd want to have Rojo delete your temp instances, yeah. Good point! Should I make the switch?

LPGhatguy: Yeah, let's do that!

sasial-dev: This would then be considered a breaking change (for 8.0.0), right?

boatbomber: Hm. It's certainly a change in behavior, but what would it break?

sasial-dev: I don't suspect it would break anything, but I wonder if it still classes as one!

boatbomber: I'd say no, as it doesn't break existing workflows. If anything it enables them as Rojo will no longer kill your temp files.

LPGhatguy (approved):

Issue #577, opened 2022-07-10
rojo build --watch 


sasial-dev:
I'm using `rojo build --watch` and there's no debug way (even when I use `--verbose`!) to tell me what files have been changed. Either changes need to be debounced slightly to prevent overloading of output as shown below, or it is an issue with my .project.json file not registering the `globIgnorePaths` properly.

<details>
<summary>
My project.json file:
</summary>

```json
{
	"name": "roblox-ts-plugin",
	"globIgnorePaths": [
		"**/package.json",
		"**/tsconfig.json",
		"out/tsconfig.tsbuildinfo"
	],
	"tree": {
		"$path": "out",
		"include": {
			"$path": "include",
			"node_modules": {
				"$path": "node_modules/@rbxts"
			}
		}
	}
}
```

</details>

https://user-images.githubusercontent.com/44125644/178163297-02534b71-601b-4bbc-ab8d-5c77098cfde5.mp4

LPGhatguy:
I'm not sure what this video is showing. It looks like `rojo build` is triggering multiple times for a batch of changes.

I think that Rojo's file watching needs some work put into it. We do have a debounce set up, but it's possible that we need to consume all pending filesystem events at once and do a secondary debounce.

Boegie19:
I think I found the underlying issue
![image](https://user-images.githubusercontent.com/34578426/181483862-28353b02-7495-4301-beb9-9a1f9c84db0e.png)

changed does not respect globIgnorePaths aka if you update something what is in globIgnorePaths it will still see the change and fire it it will only say in the patch_set that noting changed

I see 3 options for how to fix this:

1 rewrite changed so that it does not react to globIgnorePaths files (will be the longest way to fix it but probably also the neatest still is it worth the time)

2 check patch_set  before push_messages to the message_queue (This could cause problems if other parts of rojo need to know of the changes even if it is not in the ignore path but will be a quick fix)

3 just make build ignore all messages where the patch_set is empty 

Fixes 2 and 3 I can do myself if you choose option 1 then I think someone who knows the rojo project better and  has more experience with rust needs to do it.

So my question now becomes what fix should we apply here?

Boegie19:
Only thing that still needs to be done is 

> I'm not sure what this video is showing. It looks like `rojo build` is triggering multiple times for a batch of changes.
> 
> I think that Rojo's file watching needs some work put into it. We do have a debounce set up, but it's possible that we need to consume all pending filesystem events at once and do a secondary debounce.

PR #600, opened 2022-07-31
fix empty applied patch sets

Boegie19:
half of the fix for #577 
this fix will make it so that empty patch sets no longer get send as an event.
this should fix multiple empty patch sends being send.


LPGhatguy:
Hey, thanks for the PR!

I landed a similar fix to this in 565c12405eb8fb5a9d33515562fe35cdcd37b748. I recommend checking it out. It was a quick fix, so I committed it directly to master instead of going through the PR workflow.

PR #602, opened 2022-07-31
Add `--watch` to sourcemap generation

JohnnyMorganz:
Adds support for `--watch` on sourcemap generation, to automatically regenerate sourcemap on changes.

I went ahead and did some drive-by fixes for some clippy warnings too. Also added a call to `forget` similar to how `rojo build` does it. Let me know if you want this reverted and in a separate PR though


LPGhatguy (approved):
LGTM!

PR #603, opened 2022-08-01
Add visual diffs to syncing

boatbomber:
When you initially connect to a serve, it displays a visual diff of all the changes it brings and asks the user to choose how to handle it. This helps prevent accidentally syncing the wrong place, and smooths the path towards proper two-way syncing.

You may choose "Accept" to bring in these changes or "Abort" to cancel and do nothing. For low friction, if the sync is perfect (no differences) it'll just auto accept and finish connecting. When two way sync is enabled there's a third option, "Reject", which syncs your Studio DOM back to your filesystem before connecting. (This is obviously pretty incomplete as two way sync is not quite there yet, but this PR ensures we have the infrastructure and some TODO comments in the right places.)

If your Rojo UI is not open (ie: syncing via keybind without Rojo actually open, or by closing Rojo before accepting/rejecting) it will open the diff in a popup window instead. This will toggle if you toggle the widget in middle so you never need more than one Rojo window open, reducing clutter.

Clicking on an edited instance in the diff will open a dropdown of specific property differences. Color3 values will show a blot of color along with their RGB values, and there is infrastructure to display other datatypes in special ways as well if we find others worth showing as more than text.

For further utility, the "Synced X changes T ago" text on the Connected page can now be clicked to view a visual diff of what those synced changes are. This makes Rojo less opaque and can additionally aid with diagnosing issues.

demo
screenshot


lankles:
phenomenal work as always

KingJesseDev:
love the icons and ui

LastTalon:
> When two way sync is enabled there's a third option, "Reject", which sends your Studio DOM back to your filesystem before connecting.

There's possibly room for improved wording of this in the future. I can see a lot of people getting confused about what "Reject" might do.

kennethloeffler:
> > When two way sync is enabled there's a third option, "Reject", which sends your Studio DOM back to your filesystem before connecting.
> 
> There's possibly room for improved wording of this in the future. I can see a lot of people getting confused about what "Reject" might do.

We have a similar functionality in our fork of Rojo and the meanings of "Reject" and "Accept" have been a source of confusion. Because misunderstandings here can cause users to lose work, it's quite important in my mind!

Maybe we ought to have some form of tooltips that detail what will happen when the user clicks on "Reject" or "Accept"?

boatbomber:
Added tooltips, wording is subject to change. They also have ZIndex issues sometimes because Sibling means that a tooltip inside a button can't overlay nearby buttons of higher sibling zindex.

boatbomber:
Moved the tooltips to #637, waiting on that before (fixing conflicts) and continuing this.

LPGhatguy (approved):

PR #607, opened 2022-08-04
Stylua adoption

boatbomber:
Adopts JohnnyMorganz Stylua for the plugin's Lua source code. Adds a check to the lint CI as well in order to enforce style guide on future PRs.


plugin/rbx_dom_lua/EncodedValue.lua
```diff
@@ -20,8 +20,8 @@ local function serializeFloat(value)
 	return value
 end
 
-local ALL_AXES = {"X", "Y", "Z"}
-local ALL_FACES = {"Right", "Top", "Back", "Left", "Bottom", "Front"}
+local ALL_AXES = { "X", "Y", "Z" }
```
LPGhatguy, line 23: We shouldn't format rbx_dom_lua since this code comes from another repository

boatbomber: Shouldn't it be a Package then? It's not even a submodule, it's literally treated as part of this repo.

LPGhatguy: I don't know if there's a good way to include it here. It's part of another repository that has lots of other related packages that we tend to update in near-lockstep.

A Git submodule would bring in lots of unrelated code that we're including via Cargo instead.

We could publish it to Wally, but it currently doesn't have a version number, stability guarantees, or many users as far as I know.

So far, including it manually has been mostly alright.

sasial-dev: You could use [`git sparse-checkout`](https://git-scm.com/docs/git-sparse-checkout) instead of submodules for it.

sasial-dev: While it is still _technically_ experimental, it's been around for > 2 years and is used in [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped)

LPGhatguy: Oh, that's an interesting command.

PR #615, opened 2022-08-07
Surgically update instances

Boegie19:
Surgically updates instances
also deals with init and meta files correctly.

Todo:

- [x] remove print statements
- [x] Fix tests
- [x] run Rustfmt 
- [x] write new tests

maby I need to add some more tests


Boegie19:
Closed because would add to much complexity to the codebase and to_instance_snapshot is to expensive

PR #616, opened 2022-08-08
fix_vfs_double_update

Boegie19:
fixes Dir write firing whole chain when a file inside the directory was added or removed.

### DIFF
--- a/src/change_processor.rs
+++ b/src/change_processor.rs
@@ -1,5 +1,6 @@
 use std::{
     fs,
+    path::PathBuf,
     sync::{Arc, Mutex},
 };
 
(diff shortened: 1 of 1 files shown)


src/change_processor.rs
```diff
@@ -261,6 +236,45 @@ impl JobThreadContext {
     }
 }
 
+// Find the nearest ancestor to this path that has
+// associated instances in the tree. This helps make sure
+// that we handle additions correctly, especially if we
+// receive events for descendants of a large tree being
+// created all at once.
```
LPGhatguy, line 243: Please move this comment so that it's attached to the code it's intended for. This is not documentation for a function!

src/change_processor.rs
```diff
@@ -124,46 +125,20 @@ impl JobThreadContext {
...
-                    for id in affected_ids {
-                        if let Some(patch) = compute_and_apply_changes(&mut tree, &self.vfs, id) {
-                            if !patch.is_empty() {
-                                applied_patches.push(patch);
-                            }
-                        }
-                    }
+        let applied_patches = match event {
+            VfsEvent::Write(path) => {
+                if path.is_dir() {
```
LPGhatguy, line 130: Are there any docs that talk about what circumstances will cause `Write` to be emitted for a directory?

Boegie19: I see no docs about but out of my testing it seems to happen every time a file or folder gets added or removed from the specific folder the parent folders are not effected.

LPGhatguy (approved):

LPGhatguy:
Thank you!

PR #618, opened 2022-08-09
Meta 2 way sync

Boegie19:
2 way sync support for meta files.

if you change properties on a instance that is synced with rojo it would create/update the meta.js file for that one. with what you changed in studio.


src/change_processor.rs
```diff
@@ -244,7 +245,45 @@ impl JobThreadContext {
                                 );
                             }
                         } else {
-                            log::warn!("Cannot change properties besides BaseScript.Source.");
+                            if instance.metadata().relevant_paths.len() >= 2 {
```
Boegie19, line 248: I personally not sure if this is good enough if we where to add a new path to relevant_paths for some reason this would break 

solution would be to add a special field to metadata for meta files. still this would make this change a lot lager.

PR #621, opened 2022-08-11
Add two way instance added

Boegie19:
requires both
https://github.com/rojo-rbx/rbx-dom/pull/233
https://github.com/rojo-rbx/rbx-dom/pull/232
to be accepted and to be pulled into rojo to work.

this PR adds instance added events to rojo so that we can later 2 way sync model.json. rbxm and rbxmx files.
because of that this pr also adds ids created instances and adds that idea in the metadata.

### DIFF
--- a/plugin/src/ChangeBatcher/encodePatchCreate.lua
+++ b/plugin/src/ChangeBatcher/encodePatchCreate.lua
@@ -1,0 +1,47 @@
+local HttpService = game:GetService("HttpService")
+
+
+local Packages = script.Parent.Parent.Parent.Packages
+local RbxDom = require(Packages.RbxDom)
+
+local  function createInstanceSnapshot(instanceMap,instance) 
+
+	local id = HttpService:GenerateGUID(false)
+
+	instanceMap.createdInstances[id] = instance
+
+	local children = {}
+	for i,child in ipairs(instance:GetChildren()) do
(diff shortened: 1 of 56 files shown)


plugin/src/ChangeBatcher/encodePatchCreate.lua
```diff
@@ -15,28 +15,30 @@ local  function createInstanceSnapshot(instanceMap,instance)
...
-	local attributes = nil
-	if #properties.Attributes ~= 0 then
-		attributes = properties.Attributes
-	end
-
-	properties.Attributes = nil
-	if #properties == 0 then
-		properties = nil
+	local success, properties = RbxDom.findAllNoneDefaultPropertiesEncoded(instance)
+	if success
```
Kampfkarren, line 19: `if success then`? Did you test this?

Boegie19: everyting exept that final change thanks for checking

imacodr (approved):
LGTM

Boegie19:
found a flaw in my implementartion if you have mutliple $path that have the same path then this would make rojo not update all paths only the path you changed that means that rojo will get into an state where in it is not correct.

Boegie19:
Closing this since this needs to be fully redone

PR #622, opened 2022-08-12
Implement `rojo open`

### LINKED ISSUE #305: Flag to open built place file in Roblox Studio
It would be great to have a flag to build a place file and immediately open it in Roblox Studio, like:

```bash
rojo build --output Place.rbxlx --open
```

Yumacide:
Closes #305.

- [x] Better support for Linux users
- [ ] Connect to the Rojo server from the plugin -- **how?**
- [ ] Refactor code that's shared with other subcommands

From #305:
> Inject a new plugin that puts sync details into _G or into the DataModel

Maybe the plugin could expose a function in _G to connect & the server would inject a new plugin that calls that function with the required address and port?

### DIFF
--- a/src/cli/open.rs
+++ b/src/cli/open.rs
@@ -25,11 +25,12 @@ use super::GlobalOptions;
 
 #[derive(Debug, Parser)]
 pub struct OpenCommand {
-    /// Path to the project file to serve from. Defaults to default.project.json
+    /// Path to the project file to serve from. Defaults to default.project.json.
     #[clap(long)]
     pub project: Option<PathBuf>,
 
-    // Path to an output place to build and serve to. Will be created automatically if it doesn't exist
+    // Path to an output place to build and serve to. Will be created automatically
+    /// if it doesn't exist.
     #[clap(long)]
     pub output: PathBuf,
 
--- a/src/cli/open.rs
+++ b/src/cli/open.rs
@@ -47,7 +47,6 @@ impl OpenCommand {
 
         log::trace!("Constructing in-memory filesystem");
         let vfs = Vfs::new_default();
-        vfs.set_watch_enabled(false);
         let session = ServeSession::new(vfs, &project)?;
 
         if !self.output.exists() {
--- a/src/cli/open.rs
+++ b/src/cli/open.rs
@@ -26,7 +26,7 @@ use super::GlobalOptions;
 #[derive(Debug, Parser)]
 pub struct OpenCommand {
     /// Path to the project file to serve from. Defaults to default.project.json.
-    #[clap(long)]
+    #[clap(value_parser)]
     pub project: Option<PathBuf>,
 
     // Path to an output place to build and serve to. Will be created automatically
--- a/src/cli/open.rs
+++ b/src/cli/open.rs
@@ -13,13 +13,7 @@ use crate::{
     web::LiveServer,
     PROJECT_FILENAME,
 };
-use std::{
-    env,
-    net::IpAddr,
-    path::{Path, PathBuf},
-    process::{Command, Stdio},
-    sync::Arc,
-};
+use std::{env, net::IpAddr, path::PathBuf, sync::Arc};
 
 use super::GlobalOptions;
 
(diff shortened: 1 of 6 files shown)


src/cli/open.rs
```diff
@@ -0,0 +1,102 @@
...
+
+        let session = ServeSession::new(vfs, &project)?;
+
+        let studio = RobloxStudio::locate()?;
+
+        if !self.output.exists() {
+            write_model(&session, &self.output, output_kind)?;
+        }
+
+        if !plugin_exists(&studio) {
```
nezuo, line 65: I think the plugin should install every time. The plugin might not have the correct version even if it exists.

src/cli/open.rs
```diff
@@ -0,0 +1,102 @@
...
+            .unwrap_or_else(|| DEFAULT_BIND_ADDRESS.into());
+
+        let port = self
+            .port
+            .or_else(|| session.project_port())
+            .unwrap_or(DEFAULT_PORT);
+
+        let server = LiveServer::new(Arc::new(session));
+
+        let _ = show_start_message(ip, port, global.color.into());
```
nezuo, line 83:
```suggestion
        show_start_message(ip, port, global.color.into());
```
I think the Result this function returns might need to be handled as well.

src/cli/open.rs
```diff
@@ -0,0 +1,102 @@
...
+
+        Ok(())
+    }
+}
+
+fn plugin_exists(studio: &RobloxStudio) -> bool {
+    studio.plugins_path().join("rojo.rbxm").exists()
+}
+
+fn open_place(studio: &RobloxStudio, place: &Path) -> anyhow::Result<()> {
```
nezuo, line 94: We might want to use the `opener` crate for this instead.

src/cli/open.rs
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        open_place(&studio, &self.output).expect("Could not open place in Roblox Studio");
+
+        let ip = self
+            .address
+            .or_else(|| session.serve_address())
+            .unwrap_or_else(|| DEFAULT_BIND_ADDRESS.into());
+
+        let port = self
```
nezuo, line 76: Should the port be randomized so it doesn't conflict with other open serve sessions?

src/cli/open.rs
```diff
@@ -0,0 +1,102 @@
...
+    sync::Arc,
+};
+
+use super::GlobalOptions;
+
+#[derive(Debug, Parser)]
+pub struct OpenCommand {
+    /// Path to the project file to serve from. Defaults to default.project.json
+    #[clap(long)]
+    pub project: Option<PathBuf>,
```
nezuo, line 30: I think this shouldn't be a flag so it matches commands like `rojo build`.

src/cli/open.rs
```diff
@@ -0,0 +1,102 @@
...
+    path::{Path, PathBuf},
+    process::{Command, Stdio},
+    sync::Arc,
+};
+
+use super::GlobalOptions;
+
+#[derive(Debug, Parser)]
+pub struct OpenCommand {
+    /// Path to the project file to serve from. Defaults to default.project.json
```
nezuo, line 28:
```suggestion
    /// Path to the project file to serve from. Defaults to default.project.json.
```

src/cli/open.rs
```diff
@@ -0,0 +1,102 @@
...
+    #[clap(long)]
+    pub project: Option<PathBuf>,
+
+    // Path to an output place to build and serve to. Will be created automatically if it doesn't exist
```
nezuo, line 32:
```suggestion
    // Path to an output place to build and serve to. Will be created automatically if it doesn't exist.
```

src/cli/open.rs
Yumacide: The user can provide a port to serve from, or we'll just use the default port. This is the same behavior as `rojo serve`, and changing it would be out of scope I think

Yumacide: Or I guess we /could/ randomize the port just because `rojo open` automatically connects anyway.

src/cli/open.rs
Yumacide: This would make things a bit slower. Ideally, we should check if the version matches but I'm not sure how.

src/cli/open.rs
nezuo: I see now that the serve command does this. I guess it was so to ignore the result. @LPGhatguy should the result be ignored here?

src/cli/open.rs
```diff
@@ -0,0 +1,75 @@
...
+impl OpenCommand {
+    pub fn run(self, global: GlobalOptions) -> anyhow::Result<()> {
+        let project = self
+            .project
+            .unwrap_or_else(|| env::current_dir().unwrap().join(PROJECT_FILENAME));
+        let output_kind = detect_output_kind(&self.output).context(UNKNOWN_OUTPUT_KIND_ERR)?;
+
+        log::trace!("Constructing in-memory filesystem");
+        let vfs = Vfs::new_default();
+        vfs.set_watch_enabled(false);
```
Boegie19, line 50: vfs.set_watch_enabled(false); means that rojo won't sync any changes made the file system. this seems to me like this is not intended since you create a live server later.

Yumacide:
Superseded by #896.

PR #623, opened 2022-08-12
2 way sync for rbxm and rbxmx files

Boegie19:
Adds a 2 way sync implementation that adds rbxm and rbxmx suport to 2 way sync.

node would only work properly if an change added gets implemented


PolyMonoStudios (approved):
Seems good to go!

Boegie19:
is old would have to be fully redone

PR #627, opened 2022-08-16
fix release action with aftman

Boegie19:
replaced forman with afterman for builds
to fix the issues with foreman not building specific builds.
I still use foreman for the creation of the plugin

node:
In some of my test the Build (aarch64-apple-darwin) failed since I exceeded the api limit Idk if this issue would reappear for you as well but if I rerun it it then works.

error:
Aftman error: Unexpected GitHub API response: {"message":"API rate limit exceeded for 199.7.166.17. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -113,7 +113,6 @@ jobs:
           profile: minimal
 
       - name: Setup Aftman
-        # when https://github.com/ok-nick/setup-aftman/pull/1 is accepted switch back to ok-nick version
         uses: Bok-nick/setup-aftman@v0.1.0
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -112,10 +112,13 @@ jobs:
           override: true
           profile: minimal
 
-      - name: Setup Foreman
-        uses: Roblox/setup-foreman@v1
+      - name: Setup Aftman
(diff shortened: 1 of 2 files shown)


LPGhatguy:
Why would switching to Aftman fix our build? I'm not sure Foreman, Aftman, or Wally run on Ubuntu 18.04 right now.

Boegie19:
It works I get green marks (Not sure if the builds work do except for windows)
https://github.com/Boegie19/rojo/actions/runs/2870196883

it works since wally has a Aftman config  and has
wally-0.3.1-linux.zip
wally-0.3.1-macos.zip
wally-0.3.1-win64.zip

also wally runs on linux on 18.04 fine.

and foreman would not work since a it does not have mac os suport.. and linux breaks since the last release we use in our forman config v7.2.1 is not buildable on linux.

LPGhatguy:
It looks like Wally only runs on Ubuntu 18.04 by accident. The last time they published a release, `ubuntu-latest` was an alias for `ubuntu-18.04`, but now it's an alias for `ubuntu-20.04`.

Wally hasn't done anything in particular to support Aftman. It should, though, since Aftman tries to work with any published tools.

Foreman supports macOS. I see a failure when trying to install Rojo with Foreman but I don't understand why.

LPGhatguy (approved):
I dunno what's goin on with Foreman. Let's try this out!

PR #630, opened 2022-08-27
Fix typo in .gitignore

BRUJOjr:
Fix a typo in .gitignore which caused it to exclude the Roblox Place file.

### DIFF
--- a/assets/default-place-project/gitignore.txt
+++ b/assets/default-place-project/gitignore.txt
@@ -1,5 +1,5 @@
 # Project place file
-/{project_name}.rbxlx
+/{project_name}.rbxl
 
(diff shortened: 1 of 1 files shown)


assets/default-place-project/gitignore.txt
```diff
@@ -1,6 +1,6 @@
 # Project place file
-/{project_name}.rbxlx
+/{project_name}.rbxl
```
BRUJOjr, line 2: Files with the .rbxlx are not created by Rojo, files with .rbxl however are.

LPGhatguy:
This is not a typo. `.rbxlx` files are also created by Rojo and used by the test suite.

Issue #634, opened 2022-09-06
Add support for syncing in a file as if it has a different extension


Corecii:
Sometimes we have files that we'd like to include in Roblox, but they're either tool-generated with non-standard extensions, or they would work fine synced in as text. It would be nice to include these anyway by telling Rojo, "treat this as if it's a .xyz".

Examples:
* Wally generates a `.lock` file, which is really a toml. This could be synced in as text or [as toml](https://github.com/rojo-rbx/rojo/pull/633) to ensure that the packages in the game match the lockfile.
* Git's files under `.git` are extensionless, but many are just text files. Some of these could be synced in to make sure the git branch matches the place (e.g. to ensure only `main` runs on production).
* Markdown files can be synced in as text files, while also maintaining editor support and without adding new extension handling code to Rojo.

LPGhatguy:
Thank you for filing this issue! I think it's similar in spirit to #562 and #55, but I like how specific the examples you've presented are.

What do you think about overriding existing extensions? I like PRs like #633 in spirit, but I'm worried about potential compatibility hazards, especially when users are including Rojo projects written by different people.

Do you think that it'd be okay if these mappings only applied within a single Rojo project? This would help in case you depend on something like a Wally package that relies on `toml` files being turned into `ModuleScript` objects, but your project wants them as plaintext.

Corecii:
> Do you think that it'd be okay if these mappings only applied within a single Rojo project? This would help in case you depend on something like a Wally package that relies on `toml` files being turned into `ModuleScript` objects, but your project wants them as plaintext.

I think having this per-file or per-project is ideal. There are a few ways I imagined this working, and they're all scoped, not global:
* Per-file, where you can specify the type as `.meta.json` files or with meta config in the project file
* Per-directory, where certain extensions or certain filenames are rewritten
* Project-wide, where certain extensions or certain filenames are rewritten

For the actual use-cases I have for this feature, specifying one-off files in `project.json` is enough. From an end-user perspective, it would make sense to be able to include any file deep in the tree by using `.meta.json`, or to include a whole directory of e.g. `.md` files as plantext with a `.meta.json` on the directory.

I looked into adding support for using `.meta.json` to configure this, but `.meta.json` is only read _after_ the filetype is determined. That could be worked around though if we decided that treating this like Rojo metadata makes sense!

---

[...]

PR #635, opened 2022-09-07
Add type override rules

Corecii:
This is an initial implementation of the settings described in this issue: https://github.com/rojo-rbx/rojo/issues/634#issuecomment-1238729110

It adds a setting to `.project.json` files to override types of files according to globs, like so:
```json
{
  "name": "example",
  "typeOverrideRules": [
    {
      "pattern": "wally.lock",
      "use": "rojo/plaintext"
    },
    {
      "pattern": "**/*.md",
      "use": "rojo/plaintext"
    },
    {
      "pattern": "**/*.spec.lua",
      "use": "rojo/ignore"
    },
  ],
  "tree": {
    "$path": "src"
  }
}
```

The following types are implemented:
```rs
"rojo/plaintext"
"rojo/luau"
"rojo/luauserver"
"rojo/luauclient"
"rojo/json"
"rojo/csv"
"rojo/project"
"rojo/rbxm"
"rojo/rbxmx"
"rojo/jsonmodel"
"rojo/ignore"
```
...and anything else is captured as an `Other(_)` value, which currently bails if encountered since plugins aren't supported yet.

---

I'm not really sure if this is the way you'd want this implemented, but I took a crack at it because:
1. I'm interested in this feature!
2. Even if this implementation isn't used, the familiarity it builds for me with the Rojo codebase is valuable!
3. Maybe this implementation can be a starting point and improved upon? No worries if not.


Corecii:
Some thoughts:
* do we want to call these "transformers"? I noticed that's the terminology you used, and I think it makes sense!
* is `transformerRules` a reasonable name for this project property?
* are we happy with refactoring snapshotting to be based on a `Transformer` type instead of deciding how to snapshot from the filename?
* are we happy with server, client, and module scripts being different types? These sync in as different objects so it makes sense. There are potential cases where you may need to specify something to sync in as a server script that isn't suffixed with `.server.lua` (or even `.lua` for that matter).

Dekkonot:
Sorry about the 10 month delay on someone getting back to you on this!

I'm not sure on the term 'transformer' being used in project files because I associate it with modifying the file itself somehow. Since this is just using existing Rojo transformers, I think it's more fair to say we're just adding or changing file type associations. I'm not sure what a better name is though, since `fileAssociations` is probably a bad one.

I'm perfectly fine with changing snapshots to use transformer associations rather than the filename. Our current method is fine but I don't feel attached to it in lieu of this solution.

Dekkonot:
As per our own discussions, I'm going to implement this myself.

PR #637, opened 2022-09-23
Add tooltips to buttons

boatbomber:
Adds tooltips to the buttons. This may seem a bit "extra" for now, but as the Rojo workflow gets more complex (two-way, etc) the single-word buttons aren't enough to convey the meaning behind them, necessitating tooltips. Superfluous tips will likely never been seen by users anyway (since they are unlikely to hover) so there's no detriment to adding them even to simple buttons.

Features:
- Easy usage via Roact component
- Auto size and position using available space
- Tail triangle points at your mouse
- Follows theme and style
- Supports multiple canvases and arbitrary tips

Tech breakdown:
`Tooltip.Provider` wraps the App, cascading the Context everywhere. `Tooltip.Canvas` renders any tip popups- it is done this way because Sibling ZIndex means the popups must be in a separate layer above all else. It also allows us to support multiple widgets by drawing Canvases in multiple places. Finally, and this is the part most used, we have `Tooltip.Trigger` which we put inside any element that we want a popup tip on.

Sample Usage:
```Lua
Sample = e(TextButton, {
	text = "Sample",
	style = "Bordered",
}, {
	Tip = e(Tooltip.Trigger, {
		text = "This is a sample tooltip"
	}),
})
```

Demo:
tooltips

### DIFF
--- a/plugin/src/App/Components/IconButton.lua
+++ b/plugin/src/App/Components/IconButton.lua
@@ -76,7 +76,7 @@ function IconButton:render()
 			BackgroundTransparency = 1,
 		}),
 
-		Children = e("Folder", {}, self.props[Roact.Children]),
+		Children = Roact.createFragment(self.props[Roact.Children]),
 	})
 end
 
--- a/plugin/src/App/Components/IconButton.lua
+++ b/plugin/src/App/Components/IconButton.lua
@@ -75,6 +75,8 @@ function IconButton:render()
 
 			BackgroundTransparency = 1,
 		}),
+
+		Children = e("Folder", {}, self.props[Roact.Children]),
 	})
 end
 
--- a/plugin/src/App/Components/Tooltip.lua
+++ b/plugin/src/App/Components/Tooltip.lua
@@ -22,24 +22,24 @@ local TooltipContext = Roact.createContext({})
 
 local function Popup(props)
 	local textSize = TextService:GetTextSize(
-		props.Text, 16, Enum.Font.GothamMedium, Vector2.new(math.min(props.canvasSize.X, 160), math.huge)
+		props.Text, 16, Enum.Font.GothamMedium, Vector2.new(math.min(props.parentSize.X, 160), math.huge)
 	) + TEXT_PADDING
 
 	local trigger = props.Trigger:getValue()
 
-	local spaceBelow = props.canvasSize.Y - (trigger.AbsolutePosition.Y + trigger.AbsoluteSize.Y - Y_OVERLAP + TAIL_SIZE)
+	local spaceBelow = props.parentSize.Y - (trigger.AbsolutePosition.Y + trigger.AbsoluteSize.Y - Y_OVERLAP + TAIL_SIZE)
 	local spaceAbove = trigger.AbsolutePosition.Y + Y_OVERLAP - TAIL_SIZE
 
 	-- If there's not enough space below, and there's more space above, then show the tooltip above the trigger
 	local displayAbove = spaceBelow < textSize.Y and spaceAbove > spaceBelow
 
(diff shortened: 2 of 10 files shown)


plugin/src/App/Components/Tooltip.lua
```diff
@@ -0,0 +1,221 @@
...
+
+function Canvas:init()
+	self:setState({
+		canvasSize = Vector2.new(200, 100),
+	})
+end
+
+function Canvas:render()
+	return Roact.createElement(TooltipContext.Consumer, {
+        render = function(context)
```
LPGhatguy, line 134: Mixed whitespace in this file: some lines are indented with spaces!

plugin/src/App/Components/Tooltip.lua
```diff
@@ -0,0 +1,221 @@
...
+		self.destroy()
+	end
+end
+
+function Trigger:render()
+	return Roact.createElement(TooltipContext.Consumer, {
+        render = function(context)
+			self.destroy = function()
+				context.removeTip(self.id)
+			end
```
LPGhatguy, line 186: In order to keep `render` pure, can we move the `TooltipContext` wrapping out of this component? Lazily initializing a destructor like this makes me a little uncomfortable. It shouldn't be too bad to split Trigger into an inner component and an outer component that injects the context value into the trigger's props.

plugin/src/App/Components/IconButton.lua
```diff
@@ -75,6 +75,8 @@ function IconButton:render()
 
 			BackgroundTransparency = 1,
 		}),
+
+		Children = e("Folder", {}, self.props[Roact.Children]),
```
LPGhatguy, line 79: Can we use a fragment here instead of a folder?

plugin/src/App/Components/Tooltip.lua
```diff
@@ -0,0 +1,221 @@
...
+				state.tips[id] = nil
+				return state
+			end)
+		end,
+	})
+end
+
+function Provider:render()
+	return Roact.createElement(TooltipContext.Provider, {
+        value = self.state,
```
LPGhatguy, line 120: Whitespace issues here

plugin/src/App/Components/Tooltip.lua
```diff
@@ -0,0 +1,221 @@
...
+    }, self.props[Roact.Children])
+end
+
+local Canvas = Roact.Component:extend("TooltipCanvas")
```
LPGhatguy, line 124: Can we rename this concept from "Canvas" to something else? Having a prop like `canvasSize` alludes to this being a `ScrollingFrame`, the main use of the word "canvas" in Roblox. If we ever get a canvas UI element for drawing arbitrary shapes, we'd have another collision!

Maybe this could be the "TooltipContainer"?

LPGhatguy (changes requested):
Nice work! Some review feedback attached.

It's kind of strange that we have to put tooltip triggers inside widgets instead of around widgets. Alas, we live with Roblox's lack of a UI layout model. 😅

plugin/src/App/Components/IconButton.lua
boatbomber: I chose Folder because I found it helpful in debugging as it creates a clear distinction between "components that are children" and "children that are just part of the button component".

LPGhatguy: Given that there aren't many instances or components involved, the extra folder usually doesn't help that much. We we don't spend that much time debugging the individual instances in buttons, so always adding a folder when we don't have to might not be a win.

There have been bugs historically from introducing folders in places like this. That might not be the case anymore, but also gives me pause. That was one reason for Roact gaining fragments when it did, a limitation specifically with ScreenGui at the time!

LPGhatguy (approved):

PR #639, opened 2022-09-25
Add Headless API

### LINKED ISSUE #638: Rojo Headless Plugin API Proposal
Exposing a headless API into `_G` will be very valuable to the Rojo ecosystem. Users can make companion plugins, and we can even utilize it ourselves for one-time plugin injections for `rojo open`. See issues #321 and #305.
Those are use cases and requests, and we've established there is demand and utility. Now, this issue is attempting to flesh out an execution of this idea. This is an ongoing discussion, not set in stone. Tell us about your use cases and desires so we can make this as valuable as possible!

**Proposed interface:**
```Lua
_G.Rojo
```
Plugins and command bar all share the same _G which allows us to put our interface in. Using the key `"Rojo"` is easy to remember and follows branding.

**Functions:**
```Lua
Rojo:Connect(host: string?, port: string?): (boolean, string?)
```
Attempts to connect Rojo to the given host & port. Same behavior as the user clicking the Connect button (sync locking, etc, all behave the same). Returns a boolean of whether it connected or not, with a second return of a message if necessary.

```Lua
Rojo:Disconnect(): void
```
Attempts to disconnect any active sync session.

```Lua
Rojo:GetSetting(setting: string): any
```
Returns the value of the given setting.

```Lua
Rojo:SetSetting(setting: string, value: any): void
```
Sets the value of the given setting to the given value.

```Lua
Rojo:Notify(msg: string): void
```
Sends a Rojo notification. Will indicate on it that it comes from a third party plugin.

**Properties:** *(All read-only)*

```Lua
Rojo.Version: {number}
```
The Rojo plugin version. (Example: `{7, 2, 1}`)

```Lua
Rojo.ProtocolVersion: number
```
The Rojo plugin's protocol version.

```Lua
Rojo.Connected: boolean
```
Whether Rojo is currently connected to a serve session.

boatbomber:
Closes #638.

You can read the API docs here: https://github.com/boatbomber/rojo.space/blob/headless-api-docs/docs/headless-api.md


plugin/src/App/Notifications.lua
```diff
@@ -129,46 +129,46 @@ function Notification:render()
...
+						Position = UDim2.fromOffset(38, 0),
 
 						LayoutOrder = 1,
 						BackgroundTransparency = 1,
 					}),
-					Time = e("TextLabel", {
-						Text = time:FormatLocalTime("LTS", "en-us"),
-						Font = Enum.Font.Code,
+					Info = e("TextLabel", {
+						Text = if self.props.source then "From: " .. self.props.source else time:FormatLocalTime("LTS", "en-us"),
```
sasial-dev, line 154: It's still worth keeping the time, imo:
```suggestion
						Text = (if self.props.source then "From: " .. self.props.source.. " •" else "").. time:FormatLocalTime("LTS", "en-us"),
```

sasial-dev (dismissed):
You might wanna change my impl. a bit but:

Kampfkarren:
FWIW I'm not sure this needs any kind of permissions system at all, what's the thought process of that?

boatbomber:
> FWIW I'm not sure this needs any kind of permissions system at all (me and lpg talked about it briefly), what's the thought process of that?

The API holds a lot of power over the user's project, and we don't want malicious plugins to be able to impact it without the user's knowledge and consent, much like Roblox has a permission system for plugins that attempt to write scripts or web requests.

Kampfkarren:
If any plugin can access it (since it's through _G) then I guess I get it, but would need to recheck what the permission scopes are

Kampfkarren:
I also wonder if there's an attack vector of changing `_G.Rojo` from another plugin and trying to hijack access from something else, making this whole process pointless

boatbomber:
> I also wonder if there's an attack vector of changing `_G.Rojo` from another plugin and trying to hijack access from something else, making this whole process pointless (if you have any other plugins)

(Sorry for late reply, missed the notification)

I don't think that attack vector would be effective, because the middle man would need to get permissions for all the APIs anyway. For example, let's say I write "EvilRojo" and make it grab `_G.Rojo` and overwrite it with my own, so all Rojo API traffic flows through EvilRojo before being passed to the original API so that behavior doesn't seem amiss. The user would see that EvilRojo is requesting all permissions (since otherwise it can't pass calls through) and additionally, all the actions would show EvilRojo as the source.

Boegie19:
No since everyone could edit _G.Rojo there is no protection agenst editing it or is there?

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Rojo Changelog
 
 ## Unreleased Changes
+* Added headless API for Studio companion plugins. ([#631])
```
Dekkonot, line 4:
```suggestion
* Added headless API for Studio companion plugins. ([#639])
```

CHANGELOG.md
```diff
@@ -51,6 +52,7 @@
 
     **All** sync rules are reset between project files, so they must be specified in each one when nesting them. This is to ensure that nothing can break other projects by changing how files are synced!
 
+[#631]: https://github.com/rojo-rbx/rojo/pull/631
```
Dekkonot, line 55:
```suggestion
[#639]: https://github.com/rojo-rbx/rojo/pull/639
```

plugin/src/HeadlessAPI.lua
```diff
@@ -0,0 +1,452 @@
...
+					actionId .. ".style must be 'Solid' or 'Bordered'"
+				)
+				assert(type(action.layoutOrder) == "number", actionId .. ".layoutOrder must be number")
+				assert(type(action.onClick) == "function", actionId .. ".onClick must be function")
+
+				sanitizedActions[id] = {
+					text = action.text,
+					style = action.style,
+					layoutOrder = action.layoutOrder,
+					onClick = action.onClick,
```
Dekkonot, line 395: It feels like a potential security risk to directly pass the function like this, since it means a malicious actor *could* get our internal environment.

I don't know if this is something worth worrying about due to the ModuleScript design this uses (you can use `getfenv(Rojo.API.Notify)` to a similar effect) but it's worth thinking about.

plugin/src/HeadlessAPI.lua
```diff
@@ -0,0 +1,452 @@
...
+					2
+				)
+			end
+
+			return Rojo[key]
+		end,
+		__newindex = function(_, key, value)
+			error(string.format("Attempted to set Rojo.%s to %q but it's a read-only value", key, value), 2)
+			return
+		end,
```
Dekkonot, line 446:
```suggestion
		end,
        __metatable = "This metatable is locked",
```

Dekkonot:
Overall, I like the design of this but I am concerned about the potential for abuse it gives third parties. As an example, there is nothing stopping me from going `rawset(require(game.Rojo), "API", {Notify = function() print("i am being malicious") end})`.

This isn't a concern for most people, since I don't imagine most devs are just installing random plugins, but it's still a real problem that I want to avoid. Do you think there's be consequences to exposing the API as a userdata with a metatable instead of a table?

plugin/src/HeadlessAPI.lua
```diff
@@ -0,0 +1,452 @@
...
+
+		-- Sanitize request
+		local sanitizedApis = {}
+		for _, api in apis do
+			if Rojo._apiDescriptions[api] ~= nil then
+				table.insert(sanitizedApis, api)
+			else
+				warn(string.format("Rojo.%s is not a valid API", tostring(api)))
+			end
+		end
```
nezuo, line 271: I looked at the code and I don't think it's an issue, but this should probably filter out duplicate valid APIs.

plugin/src/HeadlessAPI.lua
```diff
@@ -0,0 +1,452 @@
...
+
+	Rojo._apiDescriptions.SetSetting = "Sets a Rojo setting"
+	function Rojo:SetSetting(setting: string, value: any)
+		assert(type(setting) == "string", "Setting must be type `string`")
+
+		if Rojo:_checkRateLimit("SetSetting") then
+			return
+		end
+
+		return Settings:set(setting, value)
```
nezuo, line 358: I don't think `Settings:set` validates the `value`.

plugin/src/HeadlessAPI.lua
```diff
@@ -0,0 +1,452 @@
...
+	Rojo._apiDescriptions = {}
+
+	Rojo._apiDescriptions.Changed = "An event that fires when a headless API property changes"
+	Rojo.Changed = Rojo._changedEvent.Event
+
+	Rojo._apiDescriptions.Connected = "Whether or not the plugin is connected to a Rojo server"
+	Rojo.Connected = if app.serveSession then app.serveSession:getStatus() == "Connected" else false
+
+	Rojo._apiDescriptions.Address = "The address (host:port) that the plugin is connected to"
+	Rojo.Address = nil
```
nezuo, line 39: I don't understand why `GetPortAndHost` exists when you can get the address here. Or does that return what the user has input whether they are connected or not? That could be more clear in the docs if that's the case.

plugin/src/HeadlessAPI.lua
```diff
@@ -0,0 +1,452 @@
...
+	function Rojo:GetHostAndPort(): (string, string)
+		return app:getHostAndPort()
+	end
+
+	Rojo._apiDescriptions.CreateApiContext = "Creates a new API context"
+	function Rojo:CreateApiContext(baseUrl: string)
+		assert(type(baseUrl) == "string", "Base URL must be type `string`")
+
+		return ApiContext.new(baseUrl)
+	end
```
nezuo, line 413: Is there a usecase for creating an `ApiContext`? Users wouldn't know what `CreateApiContext` is for.

nezuo (commented):
I'm curious why `RequestAccess` lets the users pick out which APIs they want to allow. I can't think of a case where you would want to give a plugin only *some* of the permissions it's asking for and for a plugin to only use some of the permissions it wants.

boatbomber:
> Do you think there's be consequences to exposing the API as a userdata with a metatable instead of a table?

Will take a fair bit of rejiggering but that should work and close some potential attack vectors. Good call.

boatbomber:
> I'm curious why `RequestAccess` lets the users pick out which APIs they want to allow. I can't think of a case where you would want to give a plugin only _some_ of the permissions it's asking for and for a plugin to only use some of the permissions it wants.

Honestly, I can't remember my original reasoning so I'm gonna redo this to be all or nothing to create a simpler UX & DX. Thanks for bringing it up.

plugin/src/HeadlessAPI.lua
boatbomber: Address is the host and port that is currently connected and can be nil. GetHostAndPort returns whatever the plugin is set to, regardless of current connection. I will try to make this more clear, thanks for calling it out.

plugin/src/HeadlessAPI.lua
boatbomber: I added it because I needed it for Rojo Scanner, so yeah there's use cases. ApiContext lets you interact with Rojo server endpoints. Without being able to create an ApiContext through this, you'd need to reimplement and maintain Rojo endpoint usage in your plugins (which would suck). It's really only needed for more complex uses.

plugin/src/HeadlessAPI.lua
boatbomber: I can wrap it in an anon function that calls the third party's `onClick` through a `task.spawn` to orphan it a little more.

`onClick = function() print(debug.traceback()) end`
As is:
> user_TestingRojoApi.rbxmx.TestingRojoApi.Script:11 function onClick
> user_Rojo.rbxm.Rojo.Plugin.App.Notifications:100 function onClick

Wrapped with spawn:
> user_TestingRojoApi.rbxmx.TestingRojoApi.Script:11 function onClick

boatbomber:
Addressed all the feedback here and on the docs PR! 🚀

plugin/src/HeadlessAPI.lua
```diff
@@ -0,0 +1,495 @@
...
+		assert(type(setting) == "string", "Setting must be type `string`")
+
+		return Settings:get(setting)
+	end
+
+	Rojo._apiDescriptions.SetSetting = {
+		Type = "Method",
+		Description = "Sets a Rojo setting",
+	}
+	function Rojo:SetSetting(setting: string, value: any)
```
nezuo, line 381: Couldn't a plugin request permissions for `SetSetting` and then set their `apiPermissions`?

plugin/src/HeadlessAPI.lua
```diff
@@ -0,0 +1,495 @@
...
+		assert(type(setting) == "string", "Setting must be type `string`")
+		assert(type(value) == type(Settings:get(setting)), "Value must be the same type as the setting")
```
nezuo, line 383: This only checks the types but we should probably check the values as well. For example `confirmationBehavior` is a `string` but it only has a few possible values.

nezuo: The `t` library would be great for this and I think we already use it as a dependency.

plugin/src/HeadlessAPI.lua
boatbomber: ... How did we not realize this until now. Lol.

kennethloeffler: Do we have any compelling reasons to allow companion plugins to set Rojo's settings yet? Maybe we should just axe `Rojo.API:SetSetting` entirely.

boatbomber: Fair enough. `GetSetting` is clearly useful, but I don't really see `SetSetting` being strictly necessary and it seems to cause more trouble than it's worth.

Dekkonot:
Reviewing the docs made me realize I have a more logistical comment for this PR: it exposes an `ApiContext`, which means we need to care about it a lot more than we current do. This applies to the Rust side of things too.

This isn't a problem, but it means we need to consider it in versioning and should probably also document both sides.

plugin/src/HeadlessAPI.lua
```diff
@@ -0,0 +1,532 @@
...
+	end
+
+	Rojo._apiDescriptions.GetSetting = {
+		Type = "Method",
+		Description = "Gets a Rojo setting",
+	}
+	function Rojo:GetSetting(setting: string): any
+		assert(type(setting) == "string", "Setting must be type `string`")
+
+		return Settings:get(setting)
```
boatbomber, line 434: This can get `apiPermissions` to read existing permission grants. Technically harmless but totally should allowlist only specific setting keys to be valid here.

PR #642, opened 2022-09-29
Support explicitly specifying http or https protocol in plugin

fewkz:
When you put in the URL `https://example.com`, it would try to send a request to `http://https://example.com` instead, which would cause an "InvalidUrl" error. This fixes the error by only implicitly adding `http://` if a protocol is not otherwise specified.
This change is necessary for a project I'm working on that serves a Rojo project over https (via an SSL termination proxy) in order to prevent man in the middle attacks.

### DIFF
--- a/plugin/src/App/init.lua
+++ b/plugin/src/App/init.lua
@@ -203,8 +203,8 @@ function App:startSession()
 	}
 
 	local baseUrl = if string.find(host, "^https?://")
-		then string.format("%s:%d", host, port)
-		else string.format("http://%s:%d", host, port)
+		then string.format("%s:%s", host, port)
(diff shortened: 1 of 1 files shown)


boatbomber (approved):
LGTM!

LPGhatguy (approved):
Thanks for filing this PR!

It's important to note that because Rojo has no authentication support in the protocol, running it over the internet is discouraged. Having no authentication also means that TLS doesn't help all that much — you don't need to have a Man in the Middle, an attacker can just send their own requests to the same place!

That said, I'm approving this PR because it's small, tidy, and readable. If we want to support more interesting input, like paths in the URL, we should invest in making a URL parsing package at some point.

PR #649, opened 2023-01-05
Nightly release workflow

boatbomber:
This workflow runs each night, generating a new pre-release from the primary branch. To avoid these infinitely piling up, it also deletes nightly releases that are over one month old.

### DIFF
--- a/.github/workflows/nightly-release.yml
+++ b/.github/workflows/nightly-release.yml
@@ -17,9 +17,9 @@ jobs:
       - name: Delete old nightly releases
         uses: dev-drprasad/delete-older-releases@v0.2.0
         with:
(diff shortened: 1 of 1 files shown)


boatbomber:
We switched from wally packages to git submodules so this would need to be updated accordingly, if we decide that this workflow is something we want.

Dekkonot:
Given the amount of support we'd have to do for people running Nightly Build XYZ, I'm hesitant to okay this. Do you think that we'd get any benefits provided Rojo ends up with active maintenance or is this something we can close?

boatbomber:
I'm fine with closing this. The reason it was made was because releases took a long time and important fixes weren't available without forking. However, it looks like maintenance is back in swing and releases should be available again.

PR #663, opened 2023-04-18
Post-Build/Post-Sync Boilerplate Add-on Capability

lifemediallc:
Adding post-build & post-sync boilerplate capabilities has the potential to unlock a wide array of new functionalities.

### Example Use-Case:
A developer is utilizing (Non-Roblox) Lua's path string version of require like so:
```lua
local module = require("src/client/module")
-- Roblox no likey :(
```
By implementing post-build and post-sync boilerplate capabilities, this developer's code can be automatically prepended during the build/sync process with their own boilerplate, transforming their code into a format that is compatible with the Roblox platform:
```lua
local require = require(game:GetService("ReplicatedStorage"):WaitForChild("Framework"))
local module = require("src/client/module")
-- Hooray! :D
```

### Example "default.project.json" File:
```json
{
  "boilerplateText": "local require = require(game:GetService('ReplicatedStorage'):WaitForChild('Framework'))",
  "globBoilerplateIgnorePaths": ["**/*.client.lua", "**/*.server.lua"],
}
```

### DIFF
--- a/src/project.rs
+++ b/src/project.rs
@@ -73,6 +73,17 @@ pub struct Project {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub serve_address: Option<IpAddr>,
 
+    /// If specified, sets the boilerplate text that rojo should add to
+    /// the beginning of scripts when parsing.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub boilerplate_text: Option<String>,
+
+    /// A list of globs, relative to the folder the project file is in, that
+    /// match files that should be excluded when Rojo is applying boilerplate
+    /// text.
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub glob_boilerplate_ignore_paths: Vec<Glob>,
+
     /// A list of globs, relative to the folder the project file is in, that
     /// match files that should be excluded if Rojo encounters them.
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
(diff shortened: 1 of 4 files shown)


sasial-dev:
I'd consider changing the `project.json` structure to look like this:
```json
{
  "boilerplate": {
    "text": "local require = require(game:GetService('ReplicatedStorage'):WaitForChild('Framework'))"
    "globIgnorePaths": ["**/*.client.lua", "**/*.server.lua"]
  }
}
```

Kampfkarren:
This doesn't feel like a good implementation of a solution to this problem and I would much rather this be in the scope of plugins than to include it in Rojo itself. This is a hack that only *really* solves this very specific issue, rather than plugins which would solve this and more.

Kampfkarren:
> I had initially considered the plugin route, but that method doesn't allow for the output of "rojo build" to be formatted properly (which wouldn't allow for CI/CD, etc.).

I think it theoretically could (we could reasonably have plugin support for a post-build step?), but I also think this is also fairly straight forward with Remodel/Lune, or at least too straight forward to be justified in putting it in Rojo itself.

> Also with the boilerplate field allowing for any text, this could open up the ability for anyone to tweak Roblox globals & create faux globals behind the scenes, allowing for people to use their own proprietary module loaders without cluttering their code-base with excessive repeated boilerplate.

Again, this feels like something Rojo shouldn't be opinionated on. Whether or not it's useful is separate to whether or not it should be in Rojo itself. That's why a plugin is more interesting here.

> I do agree that it feels like it should belong elsewhere but considering that rojo is essentially a file-system - > Roblox converter, it might not be too out of place to optionally allow for extra configurability to customize that conversion process a bit.

Right, that's where plugins come in!

PR #668, opened 2023-04-27
Sourcemap performance improvements

filiptibell:
This PR brings two performance improvements to the `rojo sourcemap` command:

-   Use `rayon` with a small threadpool to parallelize sourcemap generation while still keeping startup cost very low
-   Remove conversions to owned strings and use lifetimes tied to the dom instead, which mostly improves performance with the `--include-non-scripts` flag enabled

From my personal testing on an M1 mac this decreases the sourcemap generation time of our games by 2x or more, from ~20ms to ~8ms on one project and ~30ms to ~15ms on another. Generation is pretty fast to begin with but since sourcemaps are heavily used in interactive tools (like luau-lsp) a difference of a couple frames can be great for ux.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Rojo Changelog
 
 ## Unreleased Changes
+* Significantly improved performance of `rojo sourcemap`. ([#668])
+
+[#668]: https://github.com/rojo-rbx/rojo/pull/668
 
 ## [7.3.0] - April 22, 2023
 * Added `$attributes` to project format. ([#574])
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1900,6 +1900,7 @@ dependencies = [
  "maplit",
  "memofs",
  "notify",
+ "num_cpus",
  "opener",
  "paste",
  "pretty_assertions",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1904,6 +1904,7 @@ dependencies = [
  "paste",
  "pretty_assertions",
  "profiling",
+ "rayon",
  "rbx_binary",
  "rbx_dom_weak",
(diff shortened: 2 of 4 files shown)


Dekkonot (changes requested):
After a glance, that clippy CI failure seems unrelated to this (which is worrying in its own right but that's a different issue). While I'm not sure what impact pulling `rayon` as a dependency will have on build time or executable size, I'm guessing it will be insignificant so that's whatever.

Can you add this to the [Changelog](https://github.com/rojo-rbx/rojo/blob/master/CHANGELOG.md)? After that, I'm not seeing a reason this couldn't be merged, though this is my first time reviewing a change to Rojo proper.

filiptibell:
> Can you add this to the [Changelog](https://github.com/rojo-rbx/rojo/blob/master/CHANGELOG.md)?

Sure thing, I tried to find a similar previous note in the changelog and https://github.com/rojo-rbx/rojo/pull/548 was the most similar so its more or less copy pasted from that.

> I'm not sure what impact pulling `rayon` as a dependency will have on build time or executable size

It's a pretty lightweight dependency afaik but pulls in some heavier things such as `crossbeam`. Lock file only shows `rayon` lib itself as a new dependency though, so seems we already depend on the rest.

---

I also went ahead and used `num_cpus` to cap the number of threads we create in the threadpool based on available cpu count so we don't create unnecessary threads for low-powered systems. This was another transitive dependency we already used so it won't increase build time or binary size.

Corecii (approved):
Looks good to me! Since this is my first approving review as a maintainer I'd like additional input from other common contributors. We already have Dekkonot's approval here. @Kampfkarren think this is good to merge?

Dekkonot (approved):

PR #669, opened 2023-04-28
Fix wrong version of Rojo displaying in studio.

Miizzuu:
Not entirely sure but I looked at some of the code and it seems like this should fix the wrong version showing up, "7.2.1" instead of "7.3.0", hopefully this works...


sasial-dev (changes requested):
There's a few other files to update:
* `plugin/wally.toml`
* `aftman.toml`

sasial-dev (approved):
LGTM!

Kampfkarren (approved):
All of this ought to be in CI, this is too much to change!

Issue #666, opened 2023-04-26
Rojo out of date on cargo


captalbator:
https://crates.io/crates/rojo shows 7.2.1

`cargo install rojo --version 7.3.0` also fails

medallyon:
Bump, would be great to get this updated on cargo ASAP.

OverHash:
The holdback is that publishing to Cargo is currently broken since the change to use Wally was made ([e864cf0c7df50bceb491ad7f3fcb1ea381201749](https://github.com/rojo-rbx/rojo/commit/e864cf0c7df50bceb491ad7f3fcb1ea381201749))

Zenthial:
Is there a plan to fix this, or is there a way that I could help fixing this, as cargo is the standard method of installing for me.

Dekkonot:
The blocking issue is resolved and has been for a while, though we're rapidly approaching a new release so it's not been a priority. It's a bit awkward because any release made now would have to incorporate at least one unreleased change or fork from 7.3.0 and cherrypick https://github.com/rojo-rbx/rojo/commit/d87c76a23e6c8e2bd8f67f398fbff2604b411351. 

You should be fine building from source if you're installing using `cargo install`. If that's not possible, let me know and I'll look into getting v7.3.0 on cargo.

Either way though, 7.3.1 is coming soon:tm: so it'll resolve itself.

PR #670, opened 2023-05-02
Include Cargo Publish step

medallyon:
This should automatically publish the package to cargo. Addresses #666

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   build:
-    name: Build and Test
+    name: Build, Test, and Publish
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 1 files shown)


Kampfkarren:
1. Rojo doesn't actually publish anymore ever since the switch from subdependencies to Wally
2. Unless this is already set from something else, current key holders don't actually have access to the crates.io package.

boatbomber:
We have switched back to submodules so that is unblocked, although we still don't have crate access.

Dekkonot:
@Kampfkarren Given that both items in your list are resolved now, what does the status of this look like?

Dekkonot:
We'll revisit this if it becomes wanted or needed, but right now I'm not prioritizing automating cargo releases.

Issue #512, opened 2021-12-18
Rojo marking scripts as "Unknown Instance" when syncing


yaelvalle:
Error Returned From Roblox:
![image](https://user-images.githubusercontent.com/36064214/146646520-c20781ec-3f3e-4783-a4f1-3df45c809a12.png)

Script In Question:
![image](https://user-images.githubusercontent.com/36064214/146646548-d95e06ee-f001-42f5-a200-3da73e8e58cb.png)

yaelvalle:
Is there any way this could be my own issue and if so is there a way i can fix it myself

LPGhatguy:
Hello, thank you for filing an issue. I think there are two problems here.

The first issue is with the diagnostic that Rojo is giving you. The instance should not be marked as an unknown instance. This is hopefully just a cosmetic issue, but hints at a potential bug in the plugin's internal instance tracking.

The second issue is that Rojo is not giving you details about _why_ it couldn't update `Source`. There are currently two known reasons why `Source` can fail to update:
- The script is too large. Roblox Studio plugins have a limitation of around 192 kB for adding scripts to the game. There isn't a way for us to work around this, but we can show a better error message.
- The Rojo plugin does not have the "update scripts" permission from Studio. Users don't run into this.

If you run into this issue, I would recommend looking for any very large scripts in your game. They are likely the cause, and I recommend splitting them into multiple modules not just for Rojo, but for general readability and performance.

PR #675, opened 2023-05-11
Fix disconnected session activity

boatbomber:
When a session is disconnected, the apiContext long-polling for messages continues until resolved/rejected. This means that even after a session is disconnected, a message can be received and handled.

This leads to bad behavior, as the session was already cleaned up and the message cannot be handled correctly. The instance map was cleaned up upon disconnect, so it will warn about unapplied changes to unknown instances. (Like #512)

It's very easy to repro:
Connect a session, disconnect it, then save a change.

https://github.com/rojo-rbx/rojo/assets/40185666/846a7269-7043-4727-9f9c-b3ac55a18a3a

-----------

This PR fixes that neatly by tracking all active requests in a map, and cancelling their promises when we disconnect.

### DIFF
--- a/plugin/src/ApiContext.lua
+++ b/plugin/src/ApiContext.lua
@@ -93,6 +93,7 @@ function ApiContext.new(baseUrl)
 		__sessionId = nil,
 		__messageCursor = -1,
 		__connected = true,
+		__activeRequests = {},
 	}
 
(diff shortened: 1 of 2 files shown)


kennethloeffler (approved):
Silly bug, bug your solution looks great - no complaints here.

PR #584, opened 2022-07-20
Switch git submodules to Wally packages

boatbomber:
Removes the git submodules and moves to a more elegant solution with [Wally](https://github.com/UpliftGames/wally/).
Also migrates to @evaera's Promise. No compatibility issues arose, and plugin tests all pass.

PR #678, opened 2023-05-22
Switch plugin packages back to git submodules

kennethloeffler:
Alright, so I hate to be the one to do this, but #584 broke crates.io publishing and also caused librojo to be unusable. I see that there was some discussion on Discord shortly after the problem was realized, but there was no action taken. 

I think keeping librojo and publishing working far, far outweigh any convenience added by Wally.

I've kept the same `Packages` naming convention to keep the diff minimal.


Dekkonot (approved):
Shame but yes this is long overdue

boatbomber (approved):
You're right, and I apologize for leaving it like this for so long. Thanks for rectifying my oversight.

kennethloeffler:
@Corecii @Kampfkarren you can write to this repo, right? What do we think, is this PR alright? It unblocks publishing releases (and indeed, last release was not published), so it seems kind of urgent to me!

Corecii:
I think doing this to unblock releases is reasonable.

I think we should look to use Wally in the future, but it's missing features that make it easy for us. It would be beneficial to make feature requests on the wally repo for what we need. I can see something like a wally crate with smarter download behavior working fine as part of a build script.

Corecii (approved):

kennethloeffler:
Hmm, I still can't merge this myself, but thanks for the review!
image

I guess we can wait for kampfkarren to spin through here as well

PR #689, opened 2023-05-31
Sync reminder notification & notification actions

### LINKED ISSUE #652: Prompt to resume syncing when entering place where you previously used Rojo
Rojo has a very real issue with communicating its sync status to users. Aside from the plugin icon being different somewhere in the Plugins tab, there is absolutely nothing to suggest that anything isn't syncing, especially if you save regularly to Roblox. This is because you're still able to play the place and everything *looks* like it's there, where in reality because Rojo is disabled, no new changes have synced. This has led to *hours* of wasted time on our development team hunting down bugs that were already fixed or wondering why debug prints were not showing.

The problem here is that, being a fully managed Rojo project, we expect syncing to be on all of the time. However, if syncing is off, we aren't told about it. Primarily this is a problem whenever Studio is first opened for the day, or restarted during the day (which is now commonplace due to recent stability issues).

Talking about this with my colleague, they note that the maintainers here have previously pushed back on ideas to allow Rojo to scan and automatically connect to existing Rojo sessions on the user's computer, which *does* still have the failure case of not starting Rojo in your external editor/terminal. I will **not** propose this. Instead, I propose an alternative.

When you have run Rojo in a place previously, it is likely you will run Rojo in that place next time you open it. With this insight, I propose that Rojo show a prompt as a notification in this scenario. The prompt's purpose is twofold:

- remind you that you have run Rojo in this place previously, so you should probably run it this time too
- give you the option to attempt to connect using the same IP/port settings as last time
- alternatively, allows you to dismiss the prompt, which will cause it to no longer show in future until you use Rojo there again

This serves multiple purposes:

- in the case you have completely forgotten about Rojo, it will remind you to start it in your external editor/terminal, then gives you a one-click option to resume syncing
- in the case you have started Rojo in your external editor/terminal, but have forgotten that the Studio half requires starting separately, it will give you a one-click option to resume syncing properly
- in the case you already remembered and intended to open Rojo to resume syncing, reduces the time it takes to invoke the action you already planned to do
- in the exceedingly rare case you did not want to resume using Rojo, the prompt can be dismissed right away

It would be a good idea to put this behind a setting of some kind, so on the off-chance this is annoying for someone's specific workflow, it can be turned off.

boatbomber:
Implements and closes #652.

image

image

image


kennethloeffler (commented):
Good addition - I'm happy there's a setting to turn it on and off. Can you just edit the PR title to say that it adds notification actions?

I also noticed some perhaps surprising behavior on unpublished Roblox template places (for example, the [Baseplate resulting from File>New](https://www.roblox.com/games/95206881/)): `DataModel.PlaceId` is equal to the original template's before publishing, so if the user has ever synced to an unpublished template, they will get a reminder to sync whenever they create one.

Something similar happens with local place files, which have a `DataModel.PlaceId` always equal to zero. Not sure what to do about these (maybe just avoid showing the reminder if we're on an unpublished local place file or an unpublished template?), but at the same time, it's probably way more common for the user to want to sync to places that are already published, and they can just turn the reminder off if they're not. Dunno, interested to hear your thoughts here.

boatbomber:
Ooh, I honestly forgot those have consistent place ids. I think(?) we can ignorelist them, since the reminder doesn't make sense in their context.

kennethloeffler:
> I think(?) we can ignorelist them

It's straightforward for local place files, but Roblox could change the template places without warning, and then we'd be right back to spurious notification land. Argh.... well, they don't change very often, do they? Maybe a hard coded check against template place IDs is okay.

boatbomber:
Fixed it. It was tedious to gather all those place ids initially but it should be literally one liner PRs when they add more.

plugin/src/ignorePlaceIds.lua
```diff
@@ -0,0 +1,21 @@
+return {
```
kennethloeffler, line 1: Add a comment explaining the purpose of this module (it could maybe use a more descriptive name too), then LGTM

kennethloeffler (approved):

plugin/src/App/StatusPages/Settings/Setting.lua
```diff
@@ -59,6 +59,7 @@ function Setting:render()
 			LayoutOrder = self.props.layoutOrder,
 			ZIndex = -self.props.layoutOrder,
 			BackgroundTransparency = 1,
+			Visible = if self.props.visible == nil then true else self.props.visible,
```
chriscerie, line 62:
```suggestion
			Visible = self.props.visible,
```

chriscerie (changes requested):
Small nitpick, otherwise looks good

plugin/src/App/StatusPages/Settings/Setting.lua
boatbomber: For some reason when writing this I must have thought nil being falsy meant that no prop passed would default to not visible. I was clearly very tired, since nil means there's no prop and we resort to the Roblox default of visible. Thanks for noticing!

plugin/src/App/Notifications.lua
```diff
@@ -122,22 +153,22 @@ function Notification:render()
 				transparency = transparency,
 				size = UDim2.new(1, 0, 1, 0),
 			}, {
-				TextContainer = e("Frame", {
-					Size = UDim2.new(0, 35+textBounds.X, 1, -20),
-					Position = UDim2.new(0, 0, 0, 10),
+				Contents = e("Frame", {
+					Size = UDim2.new(0, 35 + contentX, 1, -paddingY),
+					Position = UDim2.new(0, 0, 0, paddingY/2),
```
Dekkonot, line 158:
```suggestion
					Position = UDim2.new(0, 0, 0, paddingY / 2),
```

plugin/src/App/StatusPages/Settings/init.lua
```diff
@@ -111,20 +112,28 @@ function SettingsPage:render()
...
+				layoutOrder = 5,
 			}),
 
 			LogLevel = e(Setting, {
 				id = "logLevel",
 				name = "Log Level",
 				description = "Plugin output verbosity level",
 				transparency = self.props.transparency,
-				layoutOrder = 5,
+				layoutOrder = 100,
```
Dekkonot, line 136: Why is the layout order for this and the typechecking setting so high? I'm not intimately familiar with the plugin, so forgive me if this is an obvious question.

plugin/src/App/init.lua
```diff
@@ -470,7 +523,11 @@ function App:render()
 					}),
 				}),
 
-				RojoNotifications = e("ScreenGui", {}, {
+				RojoNotifications = e("ScreenGui", {
+					ZIndexBehavior = Enum.ZIndexBehavior.Sibling,
+					ResetOnSpawn = false,
+					DisplayOrder = 100,
```
Dekkonot, line 529: Same question as with the LayoutOrders; what's the reason this is so large?

plugin/src/App/StatusPages/Settings/init.lua
boatbomber: These are dev settings, not really used by consumers, so they're set really high so that they'll always be last in the list and out of the way of users without us having to manually update their layoutorder by one every time we add a setting.

plugin/src/App/init.lua
boatbomber: Rojo Notifications need to display on top for users to see. We don't know what else may be in their CoreGui, so we just use a high display order to increase our odds of having notifications viewable.

Dekkonot (approved):
LGTM

PR #691, opened 2023-06-02
Add protection against syncing non place projects

### LINKED ISSUE #113: Make plugin warn when syncing in something that isn't a place
The plugin tries to destroy the DataModel if you sync in a project that describes a model. That's no good, we should instead catch this case and warn.

Maybe the server should also issue a warning if you serve a non-DataModel project, since that generally doesn't make much sense!

boatbomber:
Closes #113.

Does what it says on the tin. When you try to sync a non-place project, it gives a clear message instead of crashing.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,10 @@
 
 ## Unreleased Changes
 * Significantly improved performance of `rojo sourcemap`. ([#668])
+* Added protection against syncing a model to a place. ([#691])
 
 [#668]: https://github.com/rojo-rbx/rojo/pull/668
+[#691]: https://github.com/rojo-rbx/rojo/pull/691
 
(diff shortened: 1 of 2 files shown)


kennethloeffler:
Nice, easy. My only suggestion is to more explicitly tell the user what they need to do to fix this, because they might not know what a "non-place project" is or what they should be looking for. Change the message to something like:
```
Cannot sync a model as a place. Ensure Rojo is serving a project file that has a DataModel at the root of its tree and try again.
```
Maybe even show them what the project file should look like?

boatbomber:
Good change. We can link to the project format documentation.

kennethloeffler (approved):

Dekkonot (approved):

PR #692, opened 2023-06-04
Improve sync info

boatbomber:
The line of sync info text on the Connected page needed some improvements.

- Fixed an edge case where disconnecting and then reconnecting would retain outdated info
- Fixed an issue where new patches wouldn't immediately update the change info text
- Removed extraneous changes count info, as it was not useful and could be checked in the visualizer anyway
- Added warning info for when some changes fail to apply
- Updates timestamp of last sync even if patch was empty, to have a more accurate signal on Rojo's uptime

image

image


kennethloeffler:
Code looks good, but I'll be back for a more thorough review later tonight. My only suggestion for now is to make it more obvious that the "failed to apply" text is clickable (maybe it can just be underlined, but if you want to get fancy with it it could underline on mouse over):
image

boatbomber:
> make it more obvious [...] is clickable

I added an underline on hover as well as a tooltip in order to make the user actions more clear.

image

kennethloeffler (approved):

Dekkonot (approved):

PR #695, opened 2023-06-10
Enable watch after server has started (M1 performance fix)

Boegie19:
This PR makes it so we only start watching the file system for changes after we are done creating the server.
This should be a performance improvement across the board mostly for M1.
M1 has gone from 1 minute to almost instant.


Boegie19:
After more looking into it wat makes M1 slow is the registering of the file watchers. this fix would not work

PR #696, opened 2023-06-11
Change gitmodules from ssh to https

Boegie19:
Why if you want to use ssh you need todo more setup aka add a ssh key into github while with https you don't have to do the extra work aka makes rojo easier to contribute to.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,15 +1,15 @@
 [submodule "plugin/Packages/Roact"]
 	path = plugin/Packages/Roact
-	url = https://github.com/roblox/roact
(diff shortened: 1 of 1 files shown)


kennethloeffler (approved):

boatbomber (changes requested):
I am pretty sure these urls need to end in `.git`, right? At the very least, that's convention. Like so: `https://github.com/Roblox/roact.git`

Boegie19:
They can but they don't have to but if you want me to change it sure

boatbomber (approved):
LGTM

PR #697, opened 2023-06-12
Add coverage reporting

Boegie19:
Adds codecov with the instrument-coverage

Also when I saved it got formatted so

closes https://github.com/rojo-rbx/rojo/pull/424


Dekkonot:
I don't believe we need code coverage for this project. Thank you for the attempt though.

PR #698, opened 2023-06-15
Integrated Mayhem

4812571:


Dekkonot:
I don't know what this means. Can you elaborate?

Dekkonot:
Unless this is elaborated upon, this can't be acted on.

PR #708, opened 2023-07-04
Fix Rojo breaking when users undo/redo in Studio

### LINKED ISSUE #705: Undoing in Studio may undo Rojo changes
Right now, it seems the Rojo plugin does not introduce ChangeHistoryService waypoints after a sync. This means if I later try to edit something in a partially managed project (e.g., move a part around) and undo that, it will also undo the changes that happened during the rojo sync.

Ideally, a waypoint should be created so that I don't accidently undo the Rojo changes as well

boatbomber:
Closes #705.

After a patch is applied by Rojo, a ChangeHistoryService waypoint is created. This allows users to undo the patch without also undoing previous actions, as well as allowing them to undo later actions without accidentally wiping a patch as well.

image

Frankly, it's a bit shocking we haven't had more complaints about this.

In addition to fixing this bad behavior, I've added protection against accidental undos. Users are very unlikely to want to undo a Rojo patch, so we have this notification and warning to make sure it was intended.

image

Redoing or undoing further will dismiss the notification automatically as it is no longer relevant.


plugin/src/App/init.lua
```diff
@@ -97,7 +156,16 @@ function App:init()
 	end
 end
 
-function App:addNotification(text: string, timeout: number?, actions: { [string]: {text: string, style: string, layoutOrder: number, onClick: (any) -> ()} }?)
+function App:didUnmount()
```
chriscerie, line 159:
```suggestion
function App:willUnmount()
```

Dekkonot: I'm curious what the motivation is here, why the name change suggestion?

chriscerie: `didUnmount` is not a valid Roact method. It should be `willUnmount`.

boatbomber: Whoops!

Dekkonot (approved):

PR #709, opened 2023-07-04
Select Instances from diff tree

### LINKED ISSUE #707: Click to select item in initial sync diff viewer
It would be nice if the instances in the diff viewer were clickable which would focus the selection onto that item.

For example, in the sync view below, I would like to click on "Part" to find out which part exactly it was added to
image

boatbomber:
Closes #707.

Double clicking selects the Instance, if the Instance exists. Single clicking opens the changes list, if changes exist.

Tooltips indicate which actions are available for the underlying DOM element.

https://github.com/rojo-rbx/rojo/assets/40185666/cf691083-d63e-43af-a1f7-6995ea18cdf5


chriscerie:
Great change. It's not immediately obvious that right click selects the instance. Especially when the behavior allows selection for left click without a changes table, I would just think that selecting is only possible for instances without changes. I'd suggest either:

- having double click select instances
- having select be the default activated behavior and including dropdown arrow to expand changes
- including tooltip when hovering to indicate right click to select

boatbomber:
Thanks for the feedback!
- Double click seems good, and ties in with the mental model of opening files on a computer so it should be intuitive. Good idea!
- Viewing the dropdown is the primary action, plus horizontal space is pretty tight so a dropdown arrow is a no go
- A tooltip is a good call, I keep forgetting to add those!

plugin/src/App/Components/PatchVisualizer/DomLabel.lua
```diff
@@ -106,21 +108,40 @@ function DomLabel:render()
...
+				[Roact.Event.Activated] = function(_rbx: Instance, _input: InputObject, clickCount: number)
+					if clickCount == 1 then
+						-- Double click opens the instance in explorer
+						if props.instance then
+							SelectionService:Set({ props.instance })
+						end
+					elseif clickCount == 0 then
+						-- Single click expands the changes
+						if props.changeList then
+							self.expanded = not self.expanded
```
chriscerie, line 124:
```suggestion
				[Roact.Event.Activated] = function(_rbx: Instance, _input: InputObject, clickCount: number)
					if clickCount == 1 then
						-- Double click opens the instance in explorer
						self.lastDoubleClickTime = os.clock()
						if props.instance then
							SelectionService:Set({ props.instance })
						end
					elseif clickCount == 0 then
						-- Single click expands the changes
						if props.instance then
							task.wait(0.2)
							if os.clock() - (self.lastDoubleClickTime or 0) <= 0.2 then
								return
							end
						end
						
						if props.changeList then
							self.expanded = not self.expanded
```

chriscerie (commented):
Selecting instances probably shouldn't also trigger dropdown expansion. Otherwise if a user already has it expanded and want to select the instance, they'd need to select it and open changes back up.

The delay here can be up to debate. Microsoft has 500ms as the default, but I find 200-300ms to be sufficient for this case.

chriscerie (approved):

Dekkonot (commented):
So, as a miner pet peeve, I've noticed that it this seems to only work for the top-level of a change's tree.

As an example, if a patch like this is present: 
image
you could double click on `Common` to select it in the explorer but not `Hello`. 

My expected behavior would be that it would work for both. Is that something that can be easily fixed?

chriscerie:
@Dekkonot It’s only able to select it in the explorer if it already exists. It looks like you’re adding the script than modifying it (assuming green + is instance addition)

Dekkonot:
> @Dekkonot It’s only able to select it in the explorer if it already exists. It looks like you’re adding the script than modifying it (assuming green + is instance addition)

This is on initial sync into a new place file and then I'm clicking through the patch visualizer after confirming that sync. I was surprised it worked with `Common` since I thought as you did that it wouldn't work for additions. Given that it worked though, my expectation would be that it worked on all additions.

boatbomber:
That's odd. If the instance exists (and the instanceMap is aware), it should be clickable.

chriscerie:
It looks like instance isn't being passed to addition nodes. Passing the instance here should fix it.

https://github.com/rojo-rbx/rojo/blob/be2109c06ca0cc36b60e5fe94665ca877591df68/plugin/src/App/Components/PatchVisualizer/init.lua#L341-L342

boatbomber:
Good catch, thanks guys!

chriscerie (approved):

Dekkonot (approved):

PR #719, opened 2023-07-08
(feat) Add Changelog Check Action

Barocena:
This pr adds a check for changelog file changes on pull request.

`no changelog` label for skip

###### you're welcome zack

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,15 @@ jobs:
       with:
         submodules: true
 
+    - name: Changelog check
+      uses: Zomzog/changelog-checker@v1.3.0
+      with:
+        fileName: CHANGELOG.md 
+        noChangelogLabel: no changelog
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -23,6 +23,15 @@ jobs:
       with:
         submodules: true
 
+    - name: Changelog check
```
boatbomber, line 26: I think we should make a separate job for this step, so that the feedback is more clearly exposed on the PR page and has a better separation of duties considering it has nothing to do with `Build and Test`

boatbomber (changes requested):
Thanks for making this! I'll create a `no changelog` label for us to use.

boatbomber:
I made the label but will not apply it yet because I want to see this run as another check before I disable it here

boatbomber (approved):
LGTM

PR #723, opened 2023-07-09
Improve tooltip behavior

boatbomber:
Tooltips only triggered based on mouse movement, but if the hover trigger moved out from under a stationary mouse then the tooltip remained. This was really irking me, so I redid the tooltip logic to solve this.

https://github.com/rojo-rbx/rojo/assets/40185666/433c748b-201a-4ed1-a460-383f05340d00

This doesn't technically support rotated triggers, but we have no need for that so I'm not going to over complicate the math when it won't provide benefit.

### DIFF
--- a/plugin/src/App/Components/Tooltip.lua
+++ b/plugin/src/App/Components/Tooltip.lua
@@ -182,6 +182,17 @@ function Trigger:willUnmount()
 	end
 end
 
+function Trigger:didUpdate(prevProps)
+	if prevProps.text ~= self.props.text then
+		-- Any existing popup is now invalid
+		self.props.context.removeTip(self.id)
+		self.showingPopup = false
+
+		-- Let the new text propagate
+		self:managePopup()
+	end
+end
+
 function Trigger:isHovering()
 	local rbx = self.ref.current
 	if rbx then
--- a/plugin/src/App/Components/Tooltip.lua
+++ b/plugin/src/App/Components/Tooltip.lua
@@ -165,9 +165,11 @@ function Trigger:init()
 	self.id = HttpService:GenerateGUID(false)
 	self.ref = Roact.createRef()
 	self.mousePos = Vector2.zero
+	self.showingPopup = false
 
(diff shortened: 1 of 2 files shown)


plugin/src/App/Components/Tooltip.lua
```diff
@@ -180,31 +182,77 @@ function Trigger:willUnmount()
...
+		local pos = rbx.AbsolutePosition
+		local size = rbx.AbsoluteSize
+		local mousePos = self.mousePos
+
+		return
+			mousePos.X >= pos.X and mousePos.X <= pos.X + size.X
+			and mousePos.Y >= pos.Y and mousePos.Y <= pos.Y + size.Y
+	end
+	return false
+end
```
chriscerie, line 208: This is simple to leave as a local function. The fewer methods we introduce now the easier it will be to migrate to hooks when the time comes.

plugin/src/App/Components/Tooltip.lua
```diff
@@ -180,31 +182,77 @@ function Trigger:willUnmount()
...
+
 function Trigger:render()
+	local function recalculate(rbx)
+		local widget = rbx:FindFirstAncestorOfClass("DockWidgetPluginGui")
+		if not widget then return end
+		self.mousePos = widget:GetRelativeMousePosition()
+
+		self:managePopup()
+	end
```
chriscerie, line 243: I really like this change. It's a lot easier to reason about

plugin/src/App/Components/Tooltip.lua
boatbomber: We'll cross that bridge when we get to it, if we ever do.

Dekkonot: This is a valid concern, but this feels reasonable for now and we'll have to handle a bunch of stuff like this when we migrate regardless. I think this is fine for now.

Dekkonot (approved):

Issue #715, opened 2023-07-08
Add `--plugin` flag to `build` command that outputs to the local plugins folder


nezuo:
Currently, building a plugin requires a command like `rojo build -o $LOCALAPPDATA/Roblox/Plugins/Name.rbxm --watch` that is hard to remember, is not cross-platform, and ignores configured plugins folder locations.

To solve these issues, I think we should add a flag to output your project to the plugins folder without needing to know its location. The updated command would look like this:

`rojo build --plugin Name.rbxm --watch`

PR #724, opened 2023-07-09
Upgrade to clap v4

nezuo:
I need clap v4 for my planned changes for #715 because it has a new feature called [argument relations](https://docs.rs/clap/latest/clap/_derive/_tutorial/index.html#argument-relations). I can use the feature to stop users from using `project` and `output` flags at the same time.

## New output
Clap changes the output from v3 to v4. Here's what the difference looks like:

### Clap v3
image

### Clap v4
image

Notice the lack of colors, the new casing, the bold, and the underlines.

## Changes
- I no longer display the author since rojo no longer has a sole author.
- For whatever reason, clap v4 does not allow a default of `""` for paths. Instead, I use the actual paths like `default.project.json`. Now it shows up in the help output: 
image


src/cli/mod.rs
```diff
@@ -25,13 +25,14 @@ pub use self::upload::UploadCommand;
 
 /// Command line options that Rojo accepts, defined using the clap crate.
 #[derive(Debug, Parser)]
-#[clap(name = "Rojo", version, about, author)]
+#[command(name = "Rojo", version, about, author)]
```
Dekkonot, line 28:
```suggestion
#[command(name = "Rojo", version, about)]
```

As far as I know, we don't need `author` here if we aren't displaying it to begin with.

Dekkonot (commented):
Personally, I don't care about the formatting changes and the loss of colors is... tragic, but acceptable in my opinion. Hesitantly, I think this is fine with me.

What does the alternative for not upgrading look like? I know it wouldn't display as exclusive in `help` but is that necessarily a deal breaker? We could always just raise an error if both fields are filled in.

nezuo:
Here's a few screenshots of what the output looks like with argument relations for reference:
image
image
image

Dekkonot:
That's kind of a funky syntax. I know it's not something you can control, but I don't like it. That said, it's not a deal breaker because most people will probably just be memorizing how to use a command anyway.

Would like to hear @Kampfkarren's and @kennethloeffler's inputs on this too, since it's a bit more of a design choice than normal and I want to make sure that multiple eyes okay it.

kennethloeffler:
> Notice the lack of colors

taking away my colors? Everything is already gray enough :sob:

Jokes aside... I agree with dekkonot, the changes to formatting/colors aren't a big problem, but it may be a smarter engineering decision to hold off on upgrading until we have a more compelling reason to do it (like maybe another mutually exclusive set of flags, or anywhere else we could use other new clap features)

nezuo:
I've decided to close this PR and implement the plugin flag without clap v4: #735. 

The removal of colors is a somewhat controversial change, and although the output in #735 isn't amazing, clap v4's output isn't either.

If clap ever decides to support colors again, it's easy to upgrade.

PR #725, opened 2023-07-09
Better settings control

boatbomber:
- Lock some settings during sync since they won't have effect if the session was already created
  - This was done generically enough to support locking any setting for any reason in the future
- Make experimental features more clearly labelled using a flag prop that adds warnings instead of some desc text no one read

image
image


plugin/src/App/Components/Dropdown.lua
```diff
@@ -30,6 +30,12 @@ function Dropdown:init()
 end
 
 function Dropdown:didUpdate()
+	if self.props.locked then
+		self:setState({
+			open = false,
+		})
+	end
```
chriscerie, line 37:
```suggestion
function Dropdown:didUpdate(prevProps)
	if self.props.locked and not prevProps.locked then
		self:setState({
			open = false,
		})
	end
```

chriscerie (changes requested):
This will infinitely rerender otherwise.

Dekkonot (approved):

PR #726, opened 2023-07-10
Rework patch visualizer with many fixes and improvements

### LINKED ISSUE #714: Patch visualizer cannot show Instances that were removed during sync
Repro:
- Sync a place
- Open the patch visualizer by clicking on the changes info text
- Delete a file while still syncing

You'll see the patch visuals totally empty, since that removal is for an instance that no longer exists so the visualizer cannot get information on it anymore.

image

boatbomber:
Closes #714.

The patch visualizer logic had a serious flaw: it was given a patch but the state of the world may be before or after that patch was applied, making it potentially impossible to build a tree from the patch data.

This PR totally reworks that, and makes some other improvements while we're at it.

- Reconciler now has precommit and postcommit hooks for patch applying
  - This is used to compute a patch tree snapshot precommit and update the tree metadata postcommit
  - Could be useful for more things down the line! Maybe patch preprocessing middleware? Who knows
- PatchVisualizer can now display Removes that happened during sync
  - It was previously missing because the removed objects no longer existed so it couldn't get any info on them (This is resolved because the info is gotten in precommit, before the instance was removed)
- PatchVisualizer now shows Old and New values instead of just Incoming during sync
  - (Still displays Current and Incoming during confirmation)
  - This is much more useful, since you now see what the changes were and not just which things were changed
- PatchVisualizer displays clarifying message when initial sync has no changes instead of just showing a blank box
- Objects in the tree UI no longer get stuck expanded when the next patch has the same instance but different info on it
- Objects in the tree UI correctly become selectable after their instance is added and unclickable when removed during sync

### DIFF
--- a/plugin/src/App/StatusPages/Connected.lua
+++ b/plugin/src/App/StatusPages/Connected.lua
@@ -83,7 +83,7 @@ function ChangesDrawer:render()
 				transparency = self.props.transparency,
 				layoutOrder = 3,
 
-				tree = self.props.patchTree,
+				patchTree = self.props.patchTree,
 			}),
 		})
 	end)
--- a/plugin/src/App/StatusPages/Connected.lua
+++ b/plugin/src/App/StatusPages/Connected.lua
@@ -83,10 +83,7 @@ function ChangesDrawer:render()
 				transparency = self.props.transparency,
 				layoutOrder = 3,
 
-				columnVisibility = { true, false, true },
-				patch = self.props.patch,
-				unappliedPatch = self.props.unappliedPatch,
-				instanceMap = self.serveSession.__instanceMap,
+				tree = self.props.patchTree,
 			}),
 		})
 	end)
@@ -365,8 +362,7 @@ function ConnectedPage:render()
 			ChangesDrawer = e(ChangesDrawer, {
 				rendered = self.state.renderChanges,
 				transparency = self.props.transparency,
-				patch = self.props.patchData.patch,
-				unappliedPatch = self.props.patchData.unapplied,
+				patchTree = self.props.patchTree,
 				serveSession = self.props.serveSession,
 				height = self.changeDrawerHeight,
 				layoutOrder = 5,
--- a/plugin/src/App/init.lua
+++ b/plugin/src/App/init.lua
@@ -554,8 +554,8 @@ function App:render()
 					initDockState = Enum.InitialDockState.Right,
 					initEnabled = false,
 					overridePreviousState = false,
-					floatingSize = Vector2.new(300, 200),
-					minimumSize = Vector2.new(300, 120),
+					floatingSize = Vector2.new(320, 210),
+					minimumSize = Vector2.new(300, 210),
 
 					zIndexBehavior = Enum.ZIndexBehavior.Sibling,
 
--- a/plugin/src/App/init.lua
+++ b/plugin/src/App/init.lua
@@ -367,9 +367,11 @@ function App:startSession()
 	end)
 	self.cleanupPostcommit = serveSession.__reconciler:hookPostcommit(function(patch, instanceMap, unappliedPatch)
 		-- Update tree with unapplied metadata
-		self:setState({
-			patchTree = PatchTree.updateMetadata(self.state.patchTree, patch, instanceMap, unappliedPatch),
-		})
+		self:setState(function(prevState)
+			return {
+				patchTree = PatchTree.updateMetadata(prevState.patchTree, patch, instanceMap, unappliedPatch),
+			}
+		end)
 	end)
 
 	serveSession:onPatchApplied(function(patch, unapplied)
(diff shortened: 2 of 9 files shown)


plugin/src/App/init.lua
```diff
@@ -358,6 +359,19 @@ function App:startSession()
...
+		-- Build new tree for patch
+		self:setState({
+			patchTree = PatchTree.build(patch, instanceMap, { "Property", "Old", "New" }),
+		})
+	end)
+	self.cleanupPostcommit = serveSession.__reconciler:hookPostcommit(function(patch, instanceMap, unappliedPatch)
+		-- Update tree with unapplied metadata
+		self:setState({
+			patchTree = PatchTree.updateMetadata(self.state.patchTree, patch, instanceMap, unappliedPatch),
+		})
```
chriscerie, line 372: Updating state based on previous state is not allowed without using setState callback.
```suggestion
		self:setState(function(prevState)
			return {
				patchTree = PatchTree.updateMetadata(prevState.patchTree, patch, instanceMap, unappliedPatch),
			}
		end)
```

plugin/src/App/StatusPages/Connected.lua
```diff
@@ -83,10 +83,7 @@ function ChangesDrawer:render()
 				transparency = self.props.transparency,
 				layoutOrder = 3,
 
-				columnVisibility = { true, false, true },
-				patch = self.props.patch,
-				unappliedPatch = self.props.unappliedPatch,
-				instanceMap = self.serveSession.__instanceMap,
+				tree = self.props.patchTree,
```
chriscerie, line 86: It would be ideal if the same prop passed to multiple components have consistent naming.

chriscerie (changes requested):
Good work!

Dekkonot (commented):
While this looks good, I'd like more comments. It might be obvious right now what is going on in `PatchTree` but if someone else comes along, they'd have to spend a little bit of time to figure out how to work with it and what it's used for. A few comments at the top of the file and above the methods describing their function would go a long way.

boatbomber:
True! I'll document this tonight.

plugin/src/PatchTree.lua
```diff
@@ -128,6 +137,8 @@ function Tree:addNode(parent, props)
     return node
 end
 
+-- Given a list of ancestor ids in descending order, builds the nodes for them
+-- using the patch and instanceMap info
 function Tree:buildAncestryNodes(ancestry, patch, instanceMap)
```
chriscerie: I would not have guessed `ancestry` meant a list of ancestor IDs. `ancestorIds: { number }` would be clearer and removes the need for the "what" comments.

plugin/src/PatchTree.lua
```diff
@@ -147,6 +158,8 @@ function Tree:buildAncestryNodes(ancestry, patch, instanceMap)
     end
 end
 
+-- Builds a new tree from a patch and instanceMap
+-- uses changeListHeaders in node.changeList
 local function build(patch, instanceMap, changeListHeaders)
```
chriscerie: If the function name was `buildTree` here, it would be obvious from the signature that it builds a new tree from a patch and instancemap without comments.

plugin/src/PatchTree.lua
```diff
@@ -341,6 +356,8 @@ local function build(patch, instanceMap, changeListHeaders)
 	return tree
 end
 
+-- Updates the metadata of a tree with the unapplied patch and currently existing instances
+-- Builds a new tree from the data if one does not exist
 local function updateMetadata(tree, patch, instanceMap, unappliedPatch)
```
chriscerie: Same thing here, if the function name was just `updateTreeMetadata` it's obvious just from the signature that it updates metatable of tree with the given args.

plugin/src/PatchTree.lua
```diff
@@ -83,6 +86,8 @@ function Tree:forEach(callback, node, depth)
 	end
 end
 
+-- Finds a node by id, depth first
+-- target is the node to search within, defaults to root
 function Tree:getNode(id, target)
```
chriscerie: If `target` is a clearer name like `nodeToSearchIn` it wouldn't need a comment saying it's the node to search in.

chriscerie (changes requested):

plugin/src/PatchTree.lua
boatbomber: I actually had that in previous commits but removed it cuz it was redundant- PatchTree.build() doesnt need to say its a tree twice.

plugin/src/PatchTree.lua
boatbomber: Same response as above- PatchTree already says tree in it so I removed the repeated word

plugin/src/PatchTree.lua
chriscerie: This looks like just a standalone build function instead of `PatchTree.build`, so it's not immediately obvious that it builds a tree or a component of a tree until you read how it's used.

Dekkonot (approved):
This changelog is getting out of hand, we're gonna have to condense it at release!

PR #730, opened 2023-07-11
Removed InstanceSnapshot snapshot_id's redudant Ref.

u-train:
Ref now is an optional inside, so it's redundant to have an option wrapping an option. The only snapshots that were changed were any that had a Ref within (from none to zeroed). Some also had some newlines added in the end.

Aside:
Since Ref is a nullable type, I'm worried about the change. For example, it's easy for someone to insert a null ref into any map where it's expected to be non-null, like in a PatchSet.


Dekkonot (approved):
I forgot to add the skip changelog label before running the workflows, sorry!

I'm not too worried about `Ref` being null-able because it's basically identical to `Option<T>`. It's something we have to be vigilant for, but no more so than before.

PR #731, opened 2023-07-12
Support ambiguous syntax for Font properties

Dekkonot:
Given how unambiguous `Font` as a datatype actually is, it makes not very much sense to require people use the fully qualified syntax for `FontFace` and any other properties that come up. So, this PR adds a new ambiguous syntax that supports it.

Long-term, I think we should reconsider our strategy for font families to make it more obvious how to set them (maybe use an enum instead of a raw string?) but that's an issue for another day.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 * Patch visualizer now indicates what changes failed to apply. ([#717])
 * Add buttons for navigation on the Connected page ([#722])
 * Improve tooltip behavior ([#723])
+* Added better support for `Font` properties ([#731])
 
 [#668]: https://github.com/rojo-rbx/rojo/pull/668
 [#674]: https://github.com/rojo-rbx/rojo/pull/674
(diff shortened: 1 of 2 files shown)


kennethloeffler:
> Long-term, I think we should reconsider our strategy for font families to make it more obvious how to set them (maybe use an enum instead of a raw string?) but that's an issue for another day.

The `Font` type in Roblox's Lua API also uses content URIs for font families, so I don't think it will be too surprising, as long as it's documented!

Dekkonot:
I don't think it's surprising necessarily, it's just rather inconvenient to have to specify a path to a JSON! Roblox offers `Font.fromEnum` and `Font.fromName` as convenience functions and I think it'd be nice to have something similar for Rojo users. That's out of scope for this PR though.

Dekkonot:
bruh.

Hedreon:
Would love for this to be merged.

kennethloeffler (approved):
Tested building an rbxm from
```json
{
  "name": "TextBox",
  "tree": {
    "$className": "TextBox",
    "$properties": {
      "FontFace": {
        "family": "rbxasset://fonts/families/RobotoMono.json",
        "weight": "Thin",
        "style": "Normal"
      }
    }
  }
}
```
and was successful, lgtm

boatbomber (approved):

PR #733, opened 2023-07-14
Update Notify to latest version

### LINKED ISSUE #732: Upgrade `notify` as a dependency of `memofs`
The version of [`notify`](https://crates.io/crates/notify) we are using as a backend for `memofs` is over 3 years out of date. In that time, Apple has released its own chips, MacOS has undergone 3 revisions, and Windows 11 released. Notify has gone from v4 to v6, and has changed dramatically. If for no other reason than tech debt, we should upgrade. However, there is also the potential for this to fix significant performance problems people are running into on M1 chips since it is both more up to date and supports an alternative watcher backend for MacOS.

However, upgrading is non-trivial. Starting with v5, Notify removed the debounce watcher we use. They did publish an upgrade guide [here](https://github.com/notify-rs/notify/blob/main/UPGRADING_V4_TO_V5.md) but it is slightly out of date (one of the examples it links to no longer exists) and more importantly states the following:
> The old `DebouncedEvent` is completely removed. [notify-debouncer-mini](https://crates.io/crates/notify-debouncer-mini) only reports an `Any` like event (named `DebouncedEvent` too), as relying on specific kinds (Write/Create/Remove) is very platform specific and can't be guaranteed to work, relying on a lot of assumptions. 

We rely on the specific kinds of events being reported, as you can see [here](https://github.com/rojo-rbx/rojo/blob/master/crates/memofs/src/std_backend.rs#L27-L40).

It's by no means impossible, but upgrading `notify` will require us to reconsider the way we process changes in `StdBackend` since it currently relies heavily on knowing the reported type of event from the watcher.

u-train:
There are a few open questions left to answer wrt to error handling from the watch itself. I also wonder, should an exact version be placed for notify now? It leaks quite a lot of information from each of its native implementation (not without reason); meaning that the current implementation can, just, break.

`memofs` API still stays the same. Only the internal `std_backend` has change implementation. 

closes #732

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -61,7 +61,7 @@ checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
 dependencies = [
  "hermit-abi 0.1.19",
  "libc",
- "winapi 0.3.9",
+ "winapi",
 ]
 
 [[package]]
@@ -437,7 +437,7 @@ checksum = "1b1d1d91c932ef41c0f2663aa8b0ca0342d444d842c06914aa0a7e352d0bada6"
 dependencies = [
  "libc",
  "redox_users",
- "winapi 0.3.9",
+ "winapi",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


Dekkonot:
Closing this for now since it's not going to be merged soon. Sorry, and thanks for the effort.

PR #735, opened 2023-07-14
Add plugin build flag

### LINKED ISSUE #715: Add `--plugin` flag to `build` command that outputs to the local plugins folder

nezuo:
Resolves #715. 

It will not output to the correct place if the user configures the location of their plugin folder. That functionality would have to be added to `roblox_install`. I haven't looked into whether it's possible or not yet.

Here is what the output looks like:
image
image
image

I wasn't able to get the error cases to say "rojo build" in the usage section.

### DIFF
--- a/src/cli/build.rs
+++ b/src/cli/build.rs
@@ -17,6 +17,8 @@ use super::resolve_path;
 
 const UNKNOWN_OUTPUT_KIND_ERR: &str = "Could not detect what kind of file to build. \
                                        Expected output file to end in .rbxl, .rbxlx, .rbxm, or .rbxmx.";
+const UNKNOWN_PLUGIN_KIND_ERR: &str = "Could not detect what kind of file to build. \
+                                       Expected plugin file to end in .rbxm or .rbxmx.";
 
 /// Generates a model or place file from the Rojo project.
 #[derive(Debug, Parser)]
@@ -33,7 +35,7 @@ pub struct BuildCommand {
 
     /// Alternative to the output flag that outputs the result in the local plugins folder.
     ///
-    /// Should end in .rbxm, .rbxl, .rbxmx, or .rbxlx.
+    /// Should end in .rbxm, .rbxl.
     #[clap(long, short)]
     pub plugin: Option<PathBuf>,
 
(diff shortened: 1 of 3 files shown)


Dekkonot:
We don't currently account for this with the `plugin` command and as far as I know nobody has ran into trouble. It might be worth opening an issue on `roblox_install` but I don't think we need to worry about it for this PR.

src/cli/build.rs
```diff
@@ -28,7 +29,13 @@ pub struct BuildCommand {
     ///
     /// Should end in .rbxm, .rbxl, .rbxmx, or .rbxlx.
     #[clap(long, short)]
-    pub output: PathBuf,
+    pub output: Option<PathBuf>,
+
+    /// Alternative to the output flag that outputs the result in the local plugins folder.
+    ///
+    /// Should end in .rbxm, .rbxl, .rbxmx, or .rbxlx.
```
Dekkonot, line 36:
```suggestion
    /// Should end in .rbxm or .rbxmx.
```
It seems like bad UX to suggest building a place file to the plugins folder. Preventing someone from building a place file as a model and vice versa can come in a different PR (since it impacts the `build` command as a whole) but I think we should at least discourage it here.

Dekkonot (changes requested):

Dekkonot:
Also we have a benchmark in [`benches/build.rs`](https://github.com/rojo-rbx/rojo/blob/master/benches/build.rs) that needs to have the `plugin` field added.

src/cli/build.rs
```diff
@@ -28,7 +31,13 @@ pub struct BuildCommand {
     ///
     /// Should end in .rbxm, .rbxl, .rbxmx, or .rbxlx.
     #[clap(long, short)]
-    pub output: PathBuf,
+    pub output: Option<PathBuf>,
+
+    /// Alternative to the output flag that outputs the result in the local plugins folder.
+    ///
+    /// Should end in .rbxm, .rbxl.
```
Dekkonot, line 38:
```suggestion
    /// Should end in .rbxm or .rbxl.
```

Dekkonot:
Minor nitpick with the punctuation and then looks good to me

Dekkonot (approved):
Looks good to me. Thank you!

PR #738, opened 2023-07-14
Add plugin template

nezuo:
Adds a new plugin template to the `init` command: `rojo init --kind plugin`.

The model template can be used for a plugin but it has a few disadvantages:
* The default script is a `ModuleScript` instead of a `Script`
* The default script's code is irrelevant to a plugin:
https://github.com/rojo-rbx/rojo/blob/fe874720aadda3bb43591d328a09fdffc396eccd/assets/default-model-project/src-init.lua#L1-L5
* It does not show the new `plugin` build flag in the README (#735)
* The `.gitignore` only ignores place lock files but it should ignore the plugin model file

In the README, I suggest using `.rbxm` instead of `.rbxmx` unlike the old templates. Does anyone disagree with this change? Should we update the old templates to suggest this as well?


boatbomber:
Yeah, .rbxm is stable now and the more efficient option. I haven't used .rbxmx in a while and we probably shouldn't still be recommending it as the default option.

src/cli/init.rs
```diff
@@ -67,6 +73,9 @@ pub enum InitKind {
 
     /// An empty model, suitable for a library or plugin.
```
boatbomber, line 74: Do we want to update this comment? Model isn't really suited for plugins, and we have one that truly is now.

Dekkonot (approved):

PR #739, opened 2023-07-15
Improve Rojo sync reliability

### LINKED ISSUE #734: Internal sync loop is recursive w/ promises
The internal syncing loop is implemented recursively, using promises as an abstraction. This is bad because it might be hiding bugs that cause the syncing loop to die without raising any trace, and it generally is not a good way to do syncing.

A better alternative could be a simple non-recursive while loop. This is better for readability, and more straightforward.

Any other alternatives would be appreciated!

boatbomber:
Closes #734.

Rojo's main sync loop was a recursive promise method, which may have been obscuring issues. Additionally, ApiContext would choose to silently hang forever where we wanted to silence an error after disconnect, which may have been happening erroneously and causing silent hangs.

I have rewritten these so that ServeSession is non recursive and also knows how to handle ApiContext errors correctly in any state so that the hanging promise solution is no longer needed. Also, it may have not been needed ever since I added promise cancelling to ApiContext, but either way it's certainly not needed now.

### DIFF
--- a/plugin/src/ApiContext.lua
+++ b/plugin/src/ApiContext.lua
@@ -11,13 +11,6 @@ local validateApiInfo = Types.ifEnabled(Types.ApiInfoResponse)
 local validateApiRead = Types.ifEnabled(Types.ApiReadResponse)
 local validateApiSubscribe = Types.ifEnabled(Types.ApiSubscribeResponse)
 
---[[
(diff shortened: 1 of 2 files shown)


Dekkonot (approved):
Looks good to me. Hopefully this resolves our silent death issue.

PR #746, opened 2023-07-24
Modernize old templates

nezuo:
This PR is meant to start a discussion on improvements we can make to the old templates. If anyone disagrees with a change I made, let me know!

Here are the changes I made:
* `.gitignore` updates:
  * Lock files in all directories are ignored now, instead of only the root directory.
  * The binary format is now ignored for place/model files.
  * The place/model files don't need a specific name to be ignored anymore.
* Added newlines to all the files (except the project files, since `rojo fmt-project` doesn't add a newline).
* The `README`s now recommend the binary format instead of XML.
* I ran `rojo fmt-project` on the place project file. I'm not sold on this, the output is a little strange and `rojo fmt-project` isn't even deterministic.
* I removed the `FilteringEnabled` property from the place project file since it's always on by default now.
* I renamed the `model` template to `library` to better reflect what it does.


Dekkonot (commented):
I think it's actually probably fine to leave the template as `model` instead of `library`. It's the goto alternative to `place`, so it makes sense to me.

boatbomber:
Should we update the place to match the new Baseplate of Studio?

nezuo:
It's a matter of preference but the current template we have has outdated Lighting properties. I think we should update it. Does anyone have any reservations about updating it?

PR #748, opened 2023-07-26
View rich diffs for Source property changes

boatbomber:
The property inspection on .Source changes is just not helpful in the current state:

image

To remedy this, I've implemented [Neil Fraser's diffing](https://github.com/google/diff-match-patch) in Luau and used it to create meaningful Source diff pop-out widgets. It uses my open source [Highlighter](https://github.com/boatbomber/Highlighter) module to apply syntax highlighting that matches the user's Studio preferences.

image
image


plugin/src/App/StatusPages/Connected.lua
```diff
@@ -367,13 +378,59 @@ function ConnectedPage:render()
...
 				end,
 			}),
+
+			SourceDiff = e(StudioPluginGui, {
+				id = "Rojo_ConnectedSourceDiff",
+				title = "Source diff",
+				active = self.state.showingSourceDiff,
+
+				initDockState = Enum.InitialDockState.Float,
+				initEnabled = true,
```
chriscerie, line 403: What does this prop do? This looks unused.

plugin/src/App/StatusPages/Confirming.lua
```diff
@@ -120,6 +135,44 @@ function ConfirmingPage:render()
...
 				PaddingRight = UDim.new(0, 20),
 			}),
+
+			SourceDiff = e(StudioPluginGui, {
+				id = "Rojo_ConfirmingSourceDiff",
+				title = "Source diff",
+				active = self.state.showingSourceDiff,
+
+				initDockState = Enum.InitialDockState.Float,
+				initEnabled = true,
```
chriscerie, line 145: Same thing here with unused prop

plugin/src/App/Components/StringDiffVisualizer/init.lua
```diff
@@ -0,0 +1,189 @@
...
+local e = Roact.createElement
+
+local StringDiffVisualizer = Roact.Component:extend("StringDiffVisualizer")
+
+function StringDiffVisualizer:init()
+	self.scriptBackground, self.setScriptBackground = Roact.createBinding(Color3.fromRGB(0, 0, 0))
+	self.contentSize, self.setContentSize = Roact.createBinding(Vector2.new(0, 0))
+
+	self:calculateContentSize()
+	self:getScriptBackground()
```
chriscerie, line 27: Can we rename this function to be something that better indicates it performs a side effect?

plugin/src/App/Components/StringDiffVisualizer/init.lua
```diff
@@ -0,0 +1,189 @@
...
+
+	self:setState({
+		add = {},
+		remove = {},
+	})
+end
+
+function StringDiffVisualizer:getScriptBackground()
+	self.setScriptBackground(Highlighter.getTokenColor("background"))
```
chriscerie, line 36: Can we get a comment here explaining why we are setting background color to the "background" token on every render despite it being the only place that sets the binding

chriscerie (commented):
Great work

plugin/src/App/StatusPages/Connected.lua
boatbomber: Huh. You're right. I copied it from App, I guess it did something once upon a time and was never cleaned up. Good eye!

plugin/src/App/Components/CodeLabel.lua
```diff
@@ -0,0 +1,62 @@
+local Rojo = script:FindFirstAncestor("Rojo")
+local Plugin = Rojo.Plugin
```
Dekkonot, line 2:
```suggestion
```

This variable isn't used

Dekkonot (changes requested):

Dekkonot (approved):

PR #749, opened 2023-07-27
Add initial ARCHITECTURE document

Dekkonot:
A while ago, I was reading a blog post by the creator of Rust Analyzer recommending an 'ARCHITECTURE.md' document for sufficiently large projects. That blog post is [here](https://matklad.github.io/2021/02/06/ARCHITECTURE.md.html) if you're curious.

While I don't think we need a lot of what they're suggesting, I think a document in the same style would be very helpful for Rojo. While it's fine for only a handful of people to know everything about the codebase, having a map of it and the mechanisms it uses would make directing new contributors significantly easier.

To that end, I've drafted an initial version of an 'ARCHITECTURE.md' document. It's incomplete, but I'm hoping that having a baseline will allow us to add to it over time and make the job less obnoxious; one person shouldn't have to write and maintain all documentation.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -10,7 +10,7 @@ Rojo is divided into two main pieces: the server and the plugin. The server is w
 
 When serving a project, the server gathers data on all of the files in that project, puts it into a nice format, and then sends it to the plugin. Then, when something changes on the file system, it does the same thing for only the changed files and sends them to the plugin.
 
-When it recieves a patch (whether it be the initial patch or any subsequent ones), the plugin reads through it and attempts to to apply it. Any sugar (the patch visualizer, as an example) happens on top of the patches received from the server.
+When it receives a patch (whether it be the initial patch or any subsequent ones), the plugin reads through it and attempts to to apply it. Any sugar (the patch visualizer, as an example) happens on top of the patches received from the server.
 
 ## Server
 
@@ -22,7 +22,7 @@ Rojo's server component is divided into a few distinct pieces:
 
 ### The CLI
 
-The Comand Line Interface (CLI) of Rojo is the only interface for the program. It's initialized in `main.rs` but is hosted in `src/cli`.
+The Command Line Interface (CLI) of Rojo is the only interface for the program. It's initialized in `main.rs` but is hosted in `src/cli`.
 
 Each command for the CLI is hosted in its own file, with the `mod.rs` file for the `cli` module handling parsing and running each command. The commands are mostly self-contained, though may also interface with Rojo's other code when necessary.
 
@@ -57,6 +57,8 @@ This section of the document is left incomplete.
 Rojo has many data structures and their purpose might not be immediately clear at a glance. To alleviate this, they are documented below.
 
 ### Vfs
+
+To learn more, read about [`memofs` architecture](crates/memofs/ARCHITECTURE.md).
 
 ### ServeSession
 
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -6,27 +6,25 @@ This document is a "what the heck is going on" level view of Rojo and the codeba
 
 ## Overarching
 
-Rojo is divided into two main pieces: the server and the Studio plugin. The server runs with access to your filesystem (whether it be via the terminal, the visual studio code extension, or a remote machine), with the Studio plugin acting as its main client.
+Rojo is divided into several components, each layering on top of each other to provide Rojo's functionality.
 
-When serving a project, the server gathers data on all of the files in that project, puts it into a nice format, and then sends it to the plugin. Then, when something changes on the file system, it does the same thing for only the changed files and sends them to the plugin.
+At the core of Rojo lies [`ServeSession`](#servesession). As the name implies, it contains all of the components to keep a persistent DOM, react to events to update the DOM, and serve the DOM to consumers.
 
-When the plugin receives a patch from the server (whether it be the initial patch or any subsequent ones), the plugin reads through the patch and attempts to to apply the changes described by it. Any sugar (the patch visualizer, as an example) happens on top of the patches received from the server.
+Most of Rojo's uses are built upon `ServeSession`! For example, the [`sourcemap` command](#sourcemapcommand) uses `ServeSession` to generate the DOM and read it to build the `sourcemap.json` file.
 
-## Server
+### The Serve Command
 
-Rojo's server component is divided into a few distinct pieces:
+There are two main pieces in play when serving: the server and the Studio plugin.
 
-- The web server
-- The CLI
-- The snapshotting system
+The server runs a local [`LiveServer`](#liveserver) with access to your filesystem (whether it be via the terminal, the visual studio code extension, or a remote machine). It consumes a `ServeSession` and attaches a web server on top. The web server itself is very basic, consisting of around half a dozen endpoints. Generally, [`LiveServer`](#liveserver) acts as a middleman with the bulk of the work is performed by either the underlying `ServeSession` or the plugin. 
 
-### The CLI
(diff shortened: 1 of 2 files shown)


u-train:
So, I have a different idea on how to start the document. I think we should start with `ServeSession` first instead. It's the core of Rojo, since the command line interface and web API depends on `ServeSession` and `ServeSession` uses a lot of the codebase (`ChangeProcessor`, `Vfs`, `RojoTree`).

Bringing it up first gives the context to the rest of the codebase and why it does the things it does. For example, it's impossible to explain `LiveServer` (which backs `serve`) without explaining `ServeSession` and its API.

ARCHITECTURE.md
```diff
@@ -0,0 +1,71 @@
+# Architecture
+
+Rojo is a rather large project with a bunch of moving parts. While it's not too complicated in practice, it tends to be overwhelming because it's a fair bit of Rust and not very clear where to begin.
+
+This document is a "what the heck is going on" level view of Rojo and the codebase that's written to make it more reasonable to jump into something. It won't go too into depth on *how* something is done, but it will go into depth on *what* is being done.
+
+## Overarching
+
+Rojo is divided into two main pieces: the server and the plugin. The server is what's ran on your computer (whether it be via the terminal or the visual studio code extension), with the plugin serving as its main client.
```
boatbomber, line 9:
```suggestion
Rojo is divided into two main pieces: the server and the plugin. The server is what's ran on your computer (whether it be via the terminal, the visual studio code extension, or a remote machine), with the plugin serving as its main client.
```
Perhaps worth mentioning that it can work over a remote connection too?

ARCHITECTURE.md
```diff
@@ -0,0 +1,71 @@
...
+
+When serving a project, the server gathers data on all of the files in that project, puts it into a nice format, and then sends it to the plugin. Then, when something changes on the file system, it does the same thing for only the changed files and sends them to the plugin.
+
+When it receives a patch (whether it be the initial patch or any subsequent ones), the plugin reads through it and attempts to to apply it. Any sugar (the patch visualizer, as an example) happens on top of the patches received from the server.
```
boatbomber, line 13:
```suggestion
When the plugin receives a patch from the server (whether it be the initial patch or any subsequent ones), the plugin reads through the patch and attempts to to apply the changes described by it. Any sugar (the patch visualizer, as an example) happens on top of the patches received from the server.
```
Too many ambiguous uses of "it".

ARCHITECTURE.md
```diff
@@ -0,0 +1,71 @@
...
+
+When generating snapshots, files are 'transformed' into Roblox objects through what's referred to as the `snapshot middleware`. As an example, this middleware takes files named `init.lua` and transforms them into a `ModuleScript` bearing the name of the parent folder. It's also responsible for things like JSON models and `.rbxm`/`.rbxmx` models being turned into snapshottable trees.
+
+Inquiring minds should look at `snapshot/mod.rs` and `snapshot_middleware` for a more detailed explanation.
+
+Because snapshots are designed to be translated into Instances anyway, this system is also used by the `build` command to turn a Rojo project into a complete file. The backend for serializing a snapshot into a file is provided by `rbx-dom`, which is a different project.
+
+### The Web Server
+
+Rojo uses a small web server to forward changes to the plugin. Once a patch is computed by the snapshot system, it's made available via the server's API. Then, the plugin requests regularly, and if a new patch exists, recieves it and applies it in Studio.
```
boatbomber, line 47:
```suggestion
Rojo uses a small web server to forward changes to the plugin. Once a patch is computed by the snapshot system, it's made available via the server's API. Then, the plugin requests patches regularly using a technique called [long polling](https://en.wikipedia.org/wiki/Push_technology#Long_polling) to receive the next patch from the server and then applies it to the Datamodel in Studio.
```

The plugin doesn't poll, it long polls. This means the patch always exists from the plugin's perspective since it just waits until the patch does exist.

ARCHITECTURE.md
```diff
@@ -0,0 +1,71 @@
...
+
+The web server itself is very basic, consisting of around half a dozen endpoints. The bulk of the work is performed by either the snapshot system or the plugin, with the web server acting as a middleman.
+
+## The Plugin
+
+This section of the document is left incomplete.
```
boatbomber, line 53: I would love some rough stubs of things you'd like me to document here. (Can be as replies here/discord if you don't want to muddy the doc.)
For example, I don't think documenting the props of every basic component is needed, but documenting the methods on App is needed. I'd appreciate input on where to draw these lines and what info/sections would be most valuable.

ARCHITECTURE.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Data Structures
+
+Rojo has many data structures and their purpose might not be immediately clear at a glance. To alleviate this, they are documented below.
```
boatbomber, line 57: We should also document the structures in the plugin, like PatchSet. It's particularly important because of how inconsistent things are with camelCase and PascalCase in our structures across the project.

boatbomber: We'll need to do this with Plugin and Server headers in this section due to the duplicate naming (ie: ServeSession) making the distinctions critical.

boatbomber (changes requested):

ARCHITECTURE.md
boatbomber: Following up
We discussed externally, and the plan is to only document the plugin behavior that directly relates to the core functionality of Rojo, such as the reconciler and servesession. Things like patch visualization, undo/redo protection, etc will be documented elsewhere.

ARCHITECTURE.md
boatbomber: Following up
Because this doc is just meant as a high level introduction to Rojo functionality, we don't need to document the details of every plugin structure. I do think PatchSet is still very important to document though, because its the main thing the plugin interacts with during core functionality.

ARCHITECTURE.md
boatbomber, line 53:
```suggestion
As mentioned above, the Studio plugin uses the web server to receive patches that contain the changes, additions, and removals to the Studio datamodel. The patch does not actually point at instances directly but rather uses string IDs. The plugin has an InstanceMap that is a bidirectional map between instance IDs and Roblox instances. It lets us keep track of every instance that Rojo knows about.

To actually apply the patch to the Studio datamodel, the plugin has a Reconciler. The Reconciler uses the InstanceMap to find the instances mentioned by the patch, and then it attempts to modify/create/destroy instances as specified in the patch. It will also create and return a new patch that contains all of the operations that failed to apply (effectively creating a patch that if applied, would fix the failures), which is used to display warnings to the end user.

The first time the plugin connects to the server is a bit of a special case. The server gives the plugin the current state, which is first used to hydrate the InstanceMap with all the initially known instances. Then, the Reconciler computes a patch by diffing the datamodel with the server state. This is known as the "catch up" patch, and that is what is shown in the Confirming page in the plugin. When that catch up patch is applied, the Studio datamodel should match the Rojo server.
```

ARCHITECTURE.md
boatbomber: Also, this section is mildly misleading, as the initial patch is actually computed by the plugin and not received from the server. That's a bit of complexity that is explained in the plugin section, so I think keeping it simple up here is fine. The application process is the same, and that's what the focus is here.

Dekkonot: > as the initial patch is actually computed by the plugin and not received from the server

I sort of assumed we were sending an initial patch file because we compute one when a `ServeSession` is initialized in the CLI as part of building the `RojoTree`. I imagine this is just a semantics difference though: regardless of what's sent, the initial patch would have to be computed by the plugin to bring it to parity with the server.

ARCHITECTURE.md
Dekkonot: I would prefer that this be left as a PR on [my branch](https://github.com/Dekkonot/rojo/tree/ted) rather than a suggestion here, since then it can be properly reviewed.

boatbomber: Moved it here: https://github.com/Dekkonot/rojo/pull/2

ARCHITECTURE.md
Dekkonot: We have a dedicated `Plugin` subsection that this can go in if need be, though maybe it's worth giving the Plugin its own document?

PR #751, opened 2023-07-28
Make CI only run in PRs that change relevant files

Dekkonot:
It only makes sense for us to run CI on pull requests that modify files that might change our builds. We should probably still run it on every commit though, just to be sure.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ on:
     - master
 
   pull_request:
-    paths: ["*.rs", "*.toml", "*.json", "*.lua", "*.luau"]
+    paths: ["*.rs", "*.toml", "*.json", "*.lua", "*.luau", "*.rbxm", "*.rbxl", "*.rbxmx", "*.rbxlx", "*.txt", "*.yml", "*.csv"]
     branches:
     - master
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,7 @@ on:
     - master
 
   pull_request:
+    paths: ["*.rs", "*.toml", "*.json", "*.lua", "*.luau"]
     branches:
     - master
 


kennethloeffler (changes requested):
If we're gonna do this, I think it should include more extensions: 
* `.rbxmx`
* `.rbxm`
* `.rbxlx`
* `.rbxl`
* `.txt`
* `.yml`
* `.csv`

This includes files that could change the results of tests or that could change build results, both now and the in the future.

There's an argument to be made that this list should also have extensions in `assets/` like `.png`, `.mp3`, and `.svg`, but since tmk that directory is currently only used for archival purposes, I chose not to include them.

Dekkonot:
That's completely fair, I forgot those. It might be more reasonable to just run CI when specific folders change, but that's not really future proof.

boatbomber (approved):

boatbomber:
Most PRs are likely going to hit this anyway, but at least markdown-only PRs will be saved.

Dekkonot:
um?

boatbomber:
I think that this PR disables that workflow on itself, but GitHub still expects it because it's on master. So it just never runs but is annoyingly still expected. Who has permission level high enough to hit the ignore button and merge anyway?

.github/workflows/ci.yml
```diff
@@ -6,6 +6,7 @@ on:
     - master
 
   pull_request:
+    paths: ["*.rs", "*.toml", "*.json", "*.lua", "*.luau", "*.rbxm", "*.rbxl", "*.rbxmx", "*.rbxlx", "*.txt", "*.yml", "*.csv"]
```
sasial-dev, line 9: CI should run when test files have changed:
```suggestion
    paths: ["*.rs", "*.toml", "*.json", "*.lua", "*.luau", "*.rbxm", "*.rbxl", "*.rbxmx", "*.rbxlx", "*.txt", "*.yml", "*.csv", "*.snap"]
```

sasial-dev (changes requested):

Dekkonot:
Given how many file types we actually need this to run for, I'm gonna just close this I think. Not worth it to shave time off of markdown edits!

PR #755, opened 2023-08-01
Fix PatchTree performance issues

boatbomber:
When building the tree, I've implemented a few improvements:

- We no longer traverse the full ancestry for every leaf node- we exit early when we find a node that already exists
- We no longer search the entire tree to see if a node id exists before creating one with that id, we just check if is in the map


Dekkonot (approved):
This is a substantial performance improvement! Great!

PR #760, opened 2023-08-01
Don't override initial state for source diff view

Dekkonot:
Due to an implementation detail with Roact (it calls `Destroy` on unmount), we have to recreate the source diff viewer plugin gui everytime a patch is visualized.

This ends up causing trouble because Studio doesn't like it if you create a plugin with the same ID. This was causing the plugin to die when trying to visualize a source diff. 

I have no idea why this is the case, but this fixes that from what my tests have showed. 🤷🏼


boatbomber (approved):
lgtm

PR #761, opened 2023-08-01
Update Theme.lua

u-train:
We now use `Color3.fromHex` now!

(why)


boatbomber (approved):
Thanks for this! Makes maintaining and extending theme easier as we'll no longer be dependent on an external color picker.

Issue #706, opened 2023-07-02
Support breaking changes to the project format with a `$version` field


Dekkonot:
It's been made abundantly clear through user reports that having `ignore_unknown_descendants` be `false` by default was a mistake. As a result, a consensus has formed that it should be changed.

This is, however, a breaking change. The obvious proposal is to add a `$version` field to a `.project.json` but that isn't a complete solution. Since we currently don't have that field, it would need to be optional in the next non-major Rojo release, which means that going forward we would require a `$version` field on project files. This is not the end of the world, but it's something we should consider.

Additionally, introducing multiple versions to the project format has a backwards compatibility concern. We would like old Rojo projects to be readable by subsequent versions of Rojo, which means we need to consider how we would like to handle old project formats. I see two paths: support each format individually and establish a conversion from one to the other internally, or having a migration tool. Both have their ups and downs, so it's unclear which we should go for.

boatbomber:
I support migration tooling over internal handlers.
For one, it makes our maintenance easier long term. Additionally, it helps keep our users up to date. This will help us with support as well as having more users benefit from the latest and greatest features.

hoontee:
+1 for the `$version` tag, assuming that value is for the project format version and not the version that Rojo is on.
Would be good to warn of potential compat breakage.

I think it's fine to introduce breaking changes with no backwards compatibility considerations. Users can always stick with the older version of Rojo until they're ready to correct the project json.

Dekkonot:
> I think it's fine to introduce breaking changes with no backwards compatibility considerations. Users can always stick with the older version of Rojo until they're ready to correct the project json.

My concern would be something like Wally, since we don't have a mechanism for automatically updating those projects so they'd be forever out of date.

hoontee:
> My concern would be something like Wally, since we don't have a mechanism for automatically updating those projects so they'd be forever out of date.

Fair point. In this case, automatic project json format conversion seems necessary.

steveprogramc:
Thank Note

Dekkonot:
A thought: how comfortable are we with being annoying to the end-user? I want to push people to include a `version` field for projects, but having a warning every time we parse one that doesn't have it specified would be... frustrating. As an example, here's me building the plugin:

![image](https://github.com/rojo-rbx/rojo/assets/4625546/d987a561-c4a3-4168-a7cc-ddbdcfc527df)

[...]

PR #765, opened 2023-08-03
Add `RunContext` support for script outputs

### LINKED ISSUE #667: RunContext does not work
When specifying the RunContext property it simply does not get applied. Here is my setup:

image

with the following in the .meta
```json
{
    "properties": {
      "RunContext": {
        "Enum": 2
      }
    }
}
```

Which results in a Script with the RunContext set to legacy instead of Client:
image

In the mean time I have made a work-around extra plugin which applies the RunContext instead. It uses the following code:
```lua
foundClientScript.RunContext = 2;
```
which makes me believe that it can not be a permissions issue because the above code works.

Is there any way I can properly do this with rojo or do I have to wait for the time being? Am I doing something wrong?

I am using Rojo 7.3.0 but it did not work in the version prior either.

sasial-dev:
Resolves #667

This PR:

- Introduces a new field in the project file: `scriptType` which has the default value of `Class` (in parity with previous versions), but can also be `RunContext`.
- This is then passed to `InstanceContext` from the `Project` struct.
- This then changes the RunContext in the lua `snapshot_middleware`


src/snapshot_middleware/lua.rs
```diff
@@ -20,36 +36,54 @@ pub fn snapshot_lua(
 ) -> anyhow::Result<Option<InstanceSnapshot>> {
     let file_name = path.file_name().unwrap().to_string_lossy();
 
-    let (class_name, instance_name) = if let Some(name) = match_trailing(&file_name, ".server.lua")
+    let is_runcontext_enabled = true; //project.script_type == ScriptType::RunContext;
```
Kampfkarren, line 39: Nit: Should be run_context

src/snapshot_middleware/lua.rs
```diff
@@ -20,36 +36,54 @@ pub fn snapshot_lua(
...
+
+    let run_context = &rbx_reflection_database::get().enums.get("RunContext").ok_or_else(|| format_err!("Unable to get RunContext enums!"))?.items;
```
Kampfkarren, line 41: If you're not formatting, just `.expect`

src/snapshot_middleware/mod.rs
```diff
@@ -39,6 +39,7 @@ use self::{
 };
 
 pub use self::project::snapshot_project_node;
+pub use self::lua::ScriptContextType;
```
Kampfkarren, line 42: Combine the two of these like you see above in the normal `use`

src/snapshot_middleware/lua.rs
```diff
@@ -12,6 +14,20 @@ use super::{
     util::match_trailing,
 };
 
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, Default)]
+pub enum ScriptContextType {
+    #[default]
+    Class,
+    RunContext
```
Kampfkarren, line 21:
```suggestion
    RunContext,
```

src/snapshot_middleware/lua.rs
```diff
@@ -1,8 +1,10 @@
 use std::{path::Path, str};
 
-use anyhow::Context;
+use anyhow::{Context, format_err};
```
Dekkonot, line 3:
```suggestion
use anyhow::{Context};
```

src/snapshot_middleware/lua.rs
```diff
@@ -20,36 +36,54 @@ pub fn snapshot_lua(
 ) -> anyhow::Result<Option<InstanceSnapshot>> {
     let file_name = path.file_name().unwrap().to_string_lossy();
 
-    let (class_name, instance_name) = if let Some(name) = match_trailing(&file_name, ".server.lua")
+    let is_run_context_enabled = true; //project.script_type == ScriptType::RunContext;
```
Dekkonot, line 39:
```suggestion
    // project.script_type == ScriptType::RunContext;
    let is_run_context_enabled = true;
```
We generally prefer to put comments above code

src/snapshot_middleware/lua.rs
```diff
@@ -20,36 +36,54 @@ pub fn snapshot_lua(
...
         .with_context(|| format!("File was not valid UTF-8: {}", path.display()))?
         .to_owned();
 
+    let mut properties = hashmap! {
+        "Source".to_owned() => contents_str.into(),
+    };
+
+    if let Some(run_context) = run_context {
+        properties.insert("RunContext".to_owned(), Enum::from_u32(run_context.to_owned()).into());
+    }
```
Dekkonot, line 79: This will force a reallocation every time when RunContext is enabled. We should probably swap away from the `hashmap!` macro as a result.

src/project.rs
```diff
@@ -73,6 +73,10 @@ pub struct Project {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub serve_address: Option<IpAddr>,
 
+    /// The mode to use when mapping scripts onto Roblox
```
Dekkonot, line 76:
```suggestion
    /// The mode to use when mapping scripts into Roblox.
    /// Can be either `Class` or `RunContext` and determines whether script
    /// behavior is set using the `RunContext` property or the script's `ClassName`.
```

Something like this is probably better, though I am by no means a documentation expert.

Dekkonot (commented):
What happened with the submodules?

Otherwise, I think this is a good idea. I'll think on the best way to pass project config to snapshots, though I imagine we'll end up passing arund a reference to a config object, so if you want to assume that's the appropriate course I wouldn't object.

src/snapshot_middleware/lua.rs
sasial-dev: Oops, forgot to remove this - thanks!

src/project.rs
sasial-dev: Will fix everywhere, thanks

sasial-dev:
Thanks @Dekkonot! I'm not sure what happened with the submodules, as they keep pushing up and I don't want to force push after each commit (like I did with the first after I saw the submodules) I'll remove them at the end.

Okay, I'll have a go passing around a reference.

src/snapshot_middleware/lua.rs
sasial-dev: The thing to keep in mind though, is that the RunContext could be "Server"/1, "Client"/2, or "Legacy"/0 so it has to be reassigned?

sasial-dev:
Okay, this is ready for a review. Please do let me know where I should write tests for this change, as I'm not sure how far I should extend them to.

src/snapshot/metadata.rs
```diff
@@ -55,15 +57,6 @@ pub struct InstanceMetadata {
...
 impl InstanceMetadata {
-    pub fn new() -> Self {
-        Self {
-            ignore_unknown_instances: false,
-            instigating_source: None,
-            relevant_paths: Vec::new(),
-            context: InstanceContext::default(),
-        }
-    }
-
```
Dekkonot, line 66: Standard Rust convention is to use `new` and also implement `Default`. It's something Clippy suggests and is generally preferred to use `::new` over `::default`.

Dekkonot (changes requested):
I want tests for RunContext! Specifically, tests that check it works at all, if it builds properly, and how it works with nested projects.

Anything else you can think of would be good too. It doesn't need to be too extensive since it's not touching a lot of the code, but we want to have some checks just to be sure it works as expected.

src/snapshot/metadata.rs
sasial-dev: Interesting, I didn't get a clippy warn!
Thanks for this - I'll do so :)

src/snapshot/metadata.rs
```diff
@@ -93,16 +86,33 @@ impl InstanceMetadata {
...
+impl From<&InstanceContext> for InstanceMetadata {
+    fn from(context: &InstanceContext) -> Self {
+        Self {
+            ignore_unknown_instances: false,
+            instigating_source: None,
+            relevant_paths: Vec::new(),
+            context: context.to_owned(),
+        }
+    }
+}
```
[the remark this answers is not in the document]
sasial-dev: Will do! I'd be curious to know why it's not prefered, does it produce more boilerplate?

sasial-dev:
@Dekkonot, I went ahead and switched back from `::default` to `::new` for this, which removes the diff for the snapshot_middleware, but some sections (IE: `instance_snapshot`) used default in the past. How do I determine what should use `::default` and what should use `::new`? Should I change it back in places that were originally that?

Dekkonot:
> How do I determine what should use `::default` and what should use `::new`? Should I change it back in places that were originally that?

As a general rule, you should be using `::new` whenever possible. There are a variety of standard library functions that work if you have `Default` implemented, which is why it's suggested if you have `::new` anyway (it's basically free in that case), but `::new` is the constructor you want almost every time.

src/snapshot/metadata.rs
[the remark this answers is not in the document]
Dekkonot: It's not idiomatic; generally, implementations of `From` should be used to convert cleanly between types that are equivalent representations. Or it's used to eliminate common boilerplate for types that take ownership of another. You can see that if you read through the [implementations in the standard library](https://doc.rust-lang.org/std/convert/trait.From.html#implementors).

In this case, it would be doing neither, since `InstanceContext` and `InstanceMetadata` aren't equivalent types and `InstanceMetadata` isn't just a wrapper for `InstanceContext`.

sasial-dev:
Sounds good - thanks for that! I'll change back to `::default` in places where it already was there :)

Dekkonot (commented):
Like I mentioned before, tests for building and syncing are what we'll need for this, but otherwise it looks good.

kennethloeffler:
I'm not sure I understand the goal here. RunContext can already be used in Rojo via property entries in instance descriptions or meta files, and I don't see what a special `$scriptType` key does for users besides add another piece of project format to learn. Unless I'm missing something, it's just as much work to change RunContext in this PR, but the syntax is just a bit different

sasial-dev:
Hey Ken,
Here's my justifications:

RunContext is basically the "newer" way to using classes to determine the script type. If you're creating a model (like I am!) it allows for all your scripts (server, client etc) to all be bundled in - let's say - ReplicatedStorage (it's loaded in at runtime). While it is possible atm with rojo, it's a bit of an anitpattern to create a `.server.lua` file and then have a `.meta.json` with the `RunContext` as `Client`. As this is an OSS project as well, it gets really confusing for beginners.

Personally for my games, I plan to continue using classes and not switch to RunContext, but for models imo, it makes tons of sense and a first-class soulution is best.

Now you might ask, why not switch everything to RunContext? This can have unintended consequences as well. I learnt recently that it **breaks existing behavior** - it is a _breaking change_ to switch all scripts to RunContext. The first example that comes to mind (although not a good one) is that a server script will run in ReplicatedStorage, which I believe doesn't happen now (this is off the top of my head - may not be accurate?). Basically, runcontext will run the server/client scripts anywhere it can be accessed instead of specific containers.

Therefore, I still believe that a `project.json` field is the best soulution, with the default maintaing "Class" to prevent breaking changes.

kennethloeffler (commented):
At first glance, I missed that the RunContext mode applies to an entire project - it makes a lot of sense now!

src/snapshot/metadata.rs
```diff
@@ -125,11 +135,29 @@ impl InstanceContext {
...
+
+impl From<Option<ScriptContextType>> for InstanceContext {
+    fn from(script_type: Option<ScriptContextType>) -> Self {
+        Self {
+            script_type: script_type.unwrap_or_default(),
+            ..Self::new()
+        }
+    }
+}
+
```
sasial-dev, line 157: I wasn't sure if I should keep this in for the tests. Let me know your thoughts.

src/snapshot/metadata.rs
```diff
@@ -144,6 +146,15 @@ impl From<ScriptContextType> for InstanceContext {
...
 
+impl From<Option<ScriptContextType>> for InstanceContext {
+    fn from(script_type: Option<ScriptContextType>) -> Self {
+        Self {
+            script_type: script_type.unwrap_or_default(),
+            ..Self::new()
+        }
+    }
+}
+
```
sasial-dev, line 157: I wasn't sure if I should keep this in for the tests. Let me know your thoughts.

sasial-dev:
Github was having a spazz - apologies. If you check the latest commit diff, I kept in lines `140-147` of `metadata.rs`. Let me know if you think it was the right move.

Dekkonot (approved):

kennethloeffler:
Okay, closing thoughts:

I strongly believe that Rojo's next major release should emit Scripts using `RunContext` by default, and provide only legacy support for `LocalScript`. `RunContext` is the future, and it would be wise for us to stay up to date with current best practices. As sasial mentioned, this will be a breaking change. We still need to figure out exactly what to do here - I think it may be possible to do a migration, but migrations and project versioning generally must be discussed further in #706, and the details of a `RunContext` migration must be examined in a new issue.

If we agree that `RunContext` should be the default, then the `scriptType` field may not be necessary in the future, and we might want to change the name right now to let users know that it is temporary. @Dekkonot and @sasial-dev, what do you think?

Dekkonot:
I believe in the next major release, it would be fair to change what the default behavior is, but supporting both RunContext and Legacy isn't a bad thing. Older projects may want to guarantee the legacy behavior for compatibility, as an example. Keeping the field but changing its default is the path I see forward.

That said, I'm open to changing the name or behavior. If we go at it with the expectation that the current behavior is legacy and we'll move away from it eventually, a better option may be a boolean field named something like `useLegacyScripts` that for now would default to `true`.

nezuo:
@kennethloeffler Are you saying that in the next major release, `script.client.lua` should emit a `Script` with `RunContext.Client`? If so, what does legacy support for `LocalScript` look like? What exactly would the migrations be fixing?

kennethloeffler:
> @kennethloeffler Are you saying that in the next major release, `script.client.lua` should emit a `Script` with `RunContext.Client`?
Yes

> If so, what does legacy support for LocalScript look like? What exactly would the migrations be fixing?
We would use a different file extension for LocalScripts, and rename existing `*.client.lua` files for migration. I've heard that Roblox's `.rbxp` [?] format uses an extension with "legacy" in the name, we could name similarly

If you want to discuss further, please open a new issue!

sasial-dev:
Personally, I believe that the default can switch to RunContext if it's considered a breaking/major change. I also believe, the field should stay as there are valid usecases for both ScriptTypes. It also keeps the option open for other script type modes (IE: Plugin) down the line when it is released -> (if scriptMode is Plugin then `.server.lua` could go to plugin runcontext or something)

sasial-dev:
> If so, what does legacy support for LocalScript look like? What exactly would the migrations be fixing?
You'd use `$scriptType` with `"Class"` if we choose to keep it in. Removing `$scriptType` and forcing RunContext would break some odd usecases -> I can think of a Server Script in ReplicatedStorage (that may be copied over from SSS) that'll run straight away instead of on reparent.

kennethloeffler:
> Personally, I believe that the default can switch to RunContext if it's considered a breaking/major change. I also believe, the field should stay as there are valid usecases for both ScriptTypes. It also keeps the option open for other script type modes (IE: Plugin) down the line when it is released -> (if scriptMode is Plugin then `.server.lua` could go to plugin runcontext or something) 

I think it's a better idea to introduce an additional file extension for legacy script support (as well as the Plugin context) over a project-level setting in the future because:

* it's more flexible, since RunContext and legacy scripts may be mixed and matched in the same project
* it's immediately obvious what kind of script will be emitted, simply by looking at the extension
* it leaves a path open for us to provide a clean migration

I think I agree with @Dekkonot in that 'scriptType' should have a different name and be a Boolean. I understand that it doesn't feel great for this setting to be a stopgap... but don't worry, we'll still keep a lot of your work here!

sasial-dev:
If we used different paths, how would we differenciate between legacy server and runcontext server scripts?

sasial-dev:
If that's the path you want to take, what do you want done code-wise with this PR?

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@
 * Added rich Source diffs in patch visualizer ([#748])
 * Fix PatchTree performance issues ([#755])
 * Don't override the initial enabled state for source diffing ([#760])
+* A `$scriptType` field has been added to the project.json schema, allowing for scripts to be differentiated through classes or RunContext ([#765]).
```
kennethloeffler, line 28:
```suggestion
* Added the `emitLegacyScripts` field to the project format. 
When `emitLegacyScripts` is `false`, Rojo emits Scripts with the appropriate `RunContext` for `*.client.lua` and `*.server.lua` files in the project. 
When `emitLegacyScripts` is `true`, Rojo emits LocalScripts and legacy Scripts (same behavior as always). Right now, it defaults to `true`.
```
Could maybe use a table that shows the new mapping...  idk if necessary though

kennethloeffler (changes requested):
> If we used different paths, how would we differenciate between legacy server and runcontext server scripts?

We could use additional extensions, `*.legacy.server.lua` and `*.legacy.client.lua` or something - I'll open another issue for this soon, but I kinda want to wait to see exactly what Roblox did for this before we commit to anything

> If that's the path you want to take, what do you want done code-wise with this PR?

Let's change the name of `scriptType` to `emitLegacyScripts` and change its type to `Option<bool>`. 

If `"emitLegacyScripts": true`, the behavior should be equivalent to `"scriptType": "Class"`. 
If `"emitLegacyScripts": false`, the behavior should be equivalent to `"scriptType": "RunContext"`.
If the field is not present, the behavior should be equivalent to `"scriptType": "Class"`.

I don't think we really need the enum... this feature is looking like it will only be an on-off switch

sasial-dev:
@Dekkonot / @nezuo - Just confirming you guys are happy with the above?

nezuo:
I thought that `emitLegacyScripts` would default to `false` though. Do we not want to make a breaking change anymore or is the plan to do that in a separate PR? If the latter is the case, I'm on board with it.

Dekkonot:
I think long-term it's up in the air what we would do to distinguish between legacy and runcontext scripts. For now though, I think having `emitLegacyScripts` is fine.


> I thought that `emitLegacyScripts` would default to `false` though.
@nezuo That will have to be in a major version since it's breaking. Odds in favor of us bumping a minor version for the next release.

Dekkonot:
@Corecii Just to ensure we aren't misinformed or out of touch, does the plan above seem sound?

TL;DR is that long-term we would move towards preferring `RunContext` over the two distinct Script classes. This change would add a toggle for people to use *now* to use `RunContext` before we changed its default in the next major release.

Corecii:
That sounds reasonable to me. I agree that we should only change the default behavior with a major version increase since it's a breaking change.

src/snapshot/metadata.rs
```diff
@@ -134,26 +135,27 @@ impl InstanceContext {
...
+    pub fn set_emit_legacy_scripts(&mut self, emit_legacy_scripts: bool) {
+        self.emit_legacy_scripts = emit_legacy_scripts
     }
 }
 
 // serve_session always passes an option from the config file, but tests want it to be explict
 #[cfg(test)]
-impl From<ScriptContextType> for InstanceContext {
-    fn from(script_type: ScriptContextType) -> Self {
+impl From<bool> for InstanceContext {
```
Dekkonot, line 145: I'm going to have to veto having a `From<bool>` implementation for `InstanceContext`, I'm afraid. I would rather we have a constructor named `InstanceContext::with_legacy_emissions` or something to the effect instead.

This is because `From<T> for U` also implies `Into<U> for T` and I don't really want us to have the power to use a boolean where we need an `InstanceContext` by just using `into()`. It's asking for trouble.

src/snapshot/metadata.rs
```diff
@@ -134,26 +135,27 @@ impl InstanceContext {
...
-            script_type,
+            emit_legacy_scripts,
             ..Self::new()
         }
     }
 }
 
-impl From<Option<ScriptContextType>> for InstanceContext {
-    fn from(script_type: Option<ScriptContextType>) -> Self {
+impl From<Option<bool>> for InstanceContext {
```
Dekkonot, line 154: Same comment with `From<bool>` but for `From<Option<bool>>`. I'm vetoing it.

src/project.rs
```diff
@@ -73,6 +75,15 @@ pub struct Project {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub serve_address: Option<IpAddr>,
 
+    /// The mode to use when mapping scripts into Roblox.
+    /// Can be either `Class` or `RunContext` and determines whether script
```
Dekkonot, line 79: Missed this in my earlier review but this comment needs to be updated to reflect that it's a bool and not an enum now.

sasial-dev: *Just* committed before I saw that lol - let me fix

src/snapshot_middleware/project.rs
```diff
@@ -30,6 +30,11 @@ pub fn snapshot_project(
     });
 
     context.add_path_ignore_rules(rules);
+    context.set_emit_legacy_scripts(
+        project
+            .emit_legacy_scripts
+            .unwrap_or_else(|| emit_legacy_scripts_default().unwrap()),
```
Dekkonot, line 36:
```suggestion
        project
            .emit_legacy_scripts
            .or_else(emit_legacy_scripts_default).unwrap(),
```
This is more idiomatic and avoids the closure + double unwrapping

CHANGELOG.md
```diff
@@ -28,6 +28,12 @@
 * Added support for `Terrain.MaterialColors` ([#770])
 * Allow `Terrain` to be specified without a classname ([#771])
 * Add Confirmation Behavior setting ([#774])
+* Added the `emitLegacyScripts` field to the project format ([#765]). The behavior is outlined below:
+
+| `emitLegacyScripts` Value  | Action Taken by Rojo                                                                                             |
+|----------------------------|------------------------------------------------------------------------------------------------------------------|
+| false                      | Rojo emits Scripts with the appropriate `RunContext` for `*.client.lua` and `*.server.lua` files in the project. |
+| true   (default)           | Rojo emits LocalScripts and Scripts with legacy Runcontext (same behavior as previously).                        |
```
Dekkonot, line 36:
```suggestion
| `emitLegacyScripts` Value  | Action Taken by Rojo                                                                                               |
|----------------------------|--------------------------------------------------------------------------------------------------------------------|
| false                      | Rojo emits Scripts with the appropriate `RunContext` for `*.client.lua` and `*.server.lua` files in the project.   |
| true   (default)           | Rojo emits LocalScripts and Scripts with legacy `RunContext` (same behavior as previously).                        |
```

Dekkonot:
Okay there is a hiccup with merging. I don't have the authority to bypass that CI check so it can't be merged.

sasial-dev:
Thank you so much guys!

fewkz:
Is there support for gradual adoption of RunContext? Mixing Classes and RunContext together?  Considering RunContext and Class have very different behavior, supporting both only makes sense

sasial-dev:
You can use nested `.project.json` files to gradually adopt RunContext. Or, just keep the existing behavior of class scripts as it is not a breaking change.

PR #766, opened 2023-08-07
Mitigate problems with uploading to data.roblox.com

pizzaboxer:
See https://devforum.roblox.com/t/dataroblox-upload-endpoint-appears-to-no-longer-function/2510735

Recently, the error page for /Data/Upload.ashx has broken, throwing off Rojo's process for uploading to it, as it relies on it to obtain a CSRF token. In response, this changes it to always obtain a CSRF token from an alternate source.

Bonus improvement - proper content-type value.

OQD4jIsiXU

### DIFF
--- a/src/cli/upload.rs
+++ b/src/cli/upload.rs
@@ -138,7 +138,7 @@ fn do_upload(buffer: Vec<u8>, asset_id: u64, cookie: &str) -> anyhow::Result<()>
     // This endpoint is used only for fetching a CSRF token, as the upload endpoint 
     // requires one. Because this endpoint itself requires a CSRF token for all
     // requests, it will provide one in its response headers, with a HTTP 403 response.
-    let csrf_response = client.post("https://economy.roblox.com/v1/purchases/products/1")
+    let csrf_response = client.post("https://auth.roblox.com")
         .header(COOKIE, &cookie_string)
         .header(CONTENT_LENGTH, 0)
         .send()?;
--- a/src/cli/upload.rs
+++ b/src/cli/upload.rs
@@ -125,6 +125,12 @@ fn do_upload(buffer: Vec<u8>, asset_id: u64, cookie: &str) -> anyhow::Result<()>
         cookie
     );
 
+    let mut content_type = "application/octet-stream";
+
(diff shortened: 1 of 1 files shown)


pizzaboxer:
Just realized that this PR might not even matter in a few days since Roblox is going to have to fix this on their end eventually, but I'll keep this here anyway.

Dekkonot:
Roblox is looking into changing this, but in hindsight it feels like it was probably a mistake to ping the same endpoint twice like we do right now. Given we'd have to make a new release to fix this anyway though, I'm gonna wait a bit to merge this to see if a better solution comes up. Thank you very much though.

kennethloeffler:
This has been resolved by Roblox: https://devforum.roblox.com/t/dataroblox-upload-endpoint-appears-to-no-longer-function/2510735/14

Dekkonot:
As Ken has said, this has been resolved so for now I'm going to close this. We'll swap over to using a new endpoint if it becomes available but otherwise our strategy is fine for the time being.

OverHash:
If this PR makes fetching the CSRF token more stable using newer APIs than the current behavior, why not switch to it? Are we waiting for full open cloud API support?

Dekkonot:
I would like to wait for full open cloud support before changing our upload endpoint, but the CSRF endpoint is open to being changed. I hadn't actually looked at the latest revision to this PR that used the auth endpoint, so I didn't realize it was using a more stable endpoint now. I thought it was still using the economy endpoint from the initial commit. 😅 

Yeah in that case, we should probably swap to using this endpoint to get the CSRF Nevermind, I guess, on closing this!

pizzaboxer:
> but in hindsight it feels like it was probably a mistake to ping the same endpoint twice like we do right now

I realized something while working on a bunch of other projects. It may look like that it seems unreliable to have to query the error page of the endpoint we're trying to request just for the CSRF token, but that's literally how they're designed to be used. That's why uploading in Studio also broke. The token wouldn't just be there in the response headers otherwise.

I opened this before I realized that Roblox officially does the same thing. So, this kind of isn't necessary, as this workaround being needed would mean that Roblox has bigger problems on their hands.

PR #772, opened 2023-08-14
Protect against syncing and building the wrong project types

### LINKED ISSUE #736: Prevent building model as a place and vice versa
Currently, we trust the user to choose their own extension when building. However, this allows mistakes because people can build models as place files or vice versa.

Given that this should be as easy as checking whether the root of a file is a `DataModel`, it should be trivial to check whether the correct type of file is being built for the specified extension. We should probably do that.

Dekkonot:
Closes #736.

Prevents projects with a data model as a root from being built as a model file and vice versa. Also prevents syncing model files altogether.

This should mitigate a few of the more common problems with building Rojo projects. I'm not sure whether I should revert #691 or not. The check it adds is made moot with this PR, but it could be a good failsafe. Open to suggestion.

### DIFF
--- a/src/cli/serve.rs
+++ b/src/cli/serve.rs
@@ -16,6 +16,9 @@ use super::{resolve_path, GlobalOptions};
 const DEFAULT_BIND_ADDRESS: Ipv4Addr = Ipv4Addr::new(127, 0, 0, 1);
 const DEFAULT_PORT: u16 = 34872;
 
+const SYNCING_MODEL_ERR: &str = "Cannot sync a model project file. \
+    Projects must have a `DataModel` at their root to be synced.";
+
 /// Expose a Rojo project to the Rojo Studio plugin.
 #[derive(Debug, Parser)]
 pub struct ServeCommand {
@@ -40,6 +43,17 @@ impl ServeCommand {
         let vfs = Vfs::new_default();
 
         let session = Arc::new(ServeSession::new(vfs, project_path)?);
+
+        // `ServeSession.tree()` is locking, so we have to let it fall out
+        // of scope when we're done.
+        {
+            let tree = session.tree();
+            // This cannot fail because a root cant' ever not exist.
+            let root = tree.get_instance(tree.get_root_id()).unwrap();
+            if root.class_name() != "DataModel" {
+                anyhow::bail!(SYNCING_MODEL_ERR)
+            }
+        }
 
         let ip = self
             .address
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 * Fix PatchTree performance issues ([#755])
 * Don't override the initial enabled state for source diffing ([#760])
 * Projects can no longer be built with the wrong file extension ([#772])
+* Projects can no longer be served if they don't have a DataModel root ([#722])
 
 [#761]: https://github.com/rojo-rbx/rojo/pull/761
 [#745]: https://github.com/rojo-rbx/rojo/pull/745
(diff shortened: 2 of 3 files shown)


Dekkonot:
You know, that's on me I didn't think to test whether our own tests would break if we changed this, I will fix it soon:tm:. It is currently 80 degrees and my laptop is hot though, so it'll have to wait.

src/cli/serve.rs
```diff
@@ -16,6 +16,9 @@ use super::{resolve_path, GlobalOptions};
 const DEFAULT_BIND_ADDRESS: Ipv4Addr = Ipv4Addr::new(127, 0, 0, 1);
 const DEFAULT_PORT: u16 = 34872;
 
+const SYNCING_MODEL_ERR: &str = "Cannot sync a model project file. \
+    Projects must have a `DataModel` at their root to be synced.";
+
```
kennethloeffler, line 21: Can we say "serve" instead of "sync"? It's good to keep terminology consistent as much as possible

src/cli/serve.rs
```diff
@@ -41,6 +44,17 @@ impl ServeCommand {
 
         let session = Arc::new(ServeSession::new(vfs, project_path)?);
 
+        // `ServeSession.tree()` is locking, so we have to let it fall out
+        // of scope when we're done.
+        {
+            let tree = session.tree();
+            // This cannot fail because a root cant' ever not exist.
```
kennethloeffler, line 51:
```suggestion
            // This cannot fail because the root must always exist.
```

kennethloeffler (commented):
The fact that tests were relying on this behavior is pretty funny! 😂 

While technically a breaking change, I doubt anyone ever actually intended to do this, or got a working result out of it... so it's probably fine?

boatbomber:
Regarding #691, I think leaving that protection in the plugin is still valuable both as a failsafe and as protection against serves from otherwise compatible <=7.3.0 servers.

Dekkonot:
We can revisit tests later, since I'm not sure on the names of some of these. I want to keep the diff as relevant as possible though, since this turned into a bigger patch than anticipated.

Dekkonot:
..Apologies for whatever the hell notification that just sent you, Ken. Github's UI isn't intuitive for this.

src/cli/serve.rs
```diff
@@ -16,6 +16,9 @@ use super::{resolve_path, GlobalOptions};
 const DEFAULT_BIND_ADDRESS: Ipv4Addr = Ipv4Addr::new(127, 0, 0, 1);
 const DEFAULT_PORT: u16 = 34872;
 
+const SERVING_MODEL_ERR: &str = "Cannot serve a model project file. \
+    Projects must have a `DataModel` at their root to be served.";
```
boatbomber, line 20: Love the descriptive message 👍🏻

tests/tests/build.rs
```diff
@@ -6,67 +6,69 @@ use tempfile::tempdir;
...
+    rbxm_in_folder: "rbxmx",
+    rbxmx_in_folder: "rbxmx",
+    rbxmx_ref: "rbxmx",
+    script_meta_disabled: "rbxmx",
+    server_in_folder: "rbxmx",
+    server_init: "rbxmx",
+    txt: "rbxmx",
+    txt_in_folder: "rbxmx",
+    unresolved_values: "rbxlx",
+    weldconstraint: "rbxmx",
```
boatbomber, line 60: Do we use binary formats in tests at all, or do the snaps for that suck too much to use here? Either way, I think this is enough here as long as we do test that binary works elsewhere.

boatbomber (approved):

tests/tests/build.rs
Dekkonot: We don't use them because they're not human readable. Realistically, this is all we need to ensure the competency of the `build` command since there's much better test coverage in rbx-dom to ensure parity between the binary and xml format.

PR #774, opened 2023-08-15
Confirmation behaviors

### LINKED ISSUE #773: Allow changing sync popup mode between never, always & place-id based
When quickly adding patches or hotloading changes, the 'Sync changes to project X' becomes strenuous and hinders the workflow when quickly implementing changes. Thus, being able to hide the popup either permanently, or based on place-id would address these workflows.

Perhaps some consideration should also be made for separating small and 'big' changes, in case an incoming change affects too much of the codebase in a suspicious manner.

boatbomber:
Closes #773. Implements confirmation behavior setting.

image

- **Initial** (New Default)
    For a given place, each project file being served will only confirm once in a given Studio session (This resolves the hot workflows of constant connect/disconnect, while still protecting against syncing a different project)
- **Always**
    Always confirms before syncing (this is the current noisy behavior)
- **Large Changes**
    Only confirms when there are more than X changed instances, defined by the user
    image
    (Threshold setting is only shown while confirmation is set to Large Changes)
- **Unlisted PlaceId**
    Only confirms when the project doesn't include the placeId in servePlaceIds

Due to the nature of the threshold setting, I also took the time to make our Setting component more versatile and robust. It can now handle any input component (checkbox, dropdown, custom button/textbox, plus the reset button) without worrying about desc text overlap.


plugin/src/App/init.lua
```diff
@@ -462,6 +465,31 @@ function App:startSession()
...
+		local confirmationBehavior = Settings:get("confirmationBehavior")
+		if confirmationBehavior == "Initial" then
+			-- Only confirm if we haven't synced this project yet this session
+			if self.knownProjects[serverInfo.projectName] then
+				Log.trace("Accepting patch without confirmation because project has already been connected and behavior is set to Initial")
+				return "Accept"
+			end
+		elseif confirmationBehavior == "Large Changes" then
+			-- Only confirm if the patch impacts many instances
+			local instanceThreshold = 5
```
sasial-dev, line 477: This feels like it should be a constant moved up scope!

sasial-dev: Maybe even as a setting?

boatbomber: I originally had it inline and decided to make it a variable so it would have a clear name for future maintainers. I'm open to making it a user configurable setting though!

sasial-dev: I feel like a setting is the way to go, yeah!

Dekkonot (changes requested):
Can we change what color the text for the input is? It's very difficult to read when it's hovered over right now.

image

boatbomber:
Whoops 😅

Dekkonot (approved):
Looks good to me!

PR #775, opened 2023-08-16
Update rbx_dom_lua to rojo-rbx/rbx-dom@7b03132

kennethloeffler:
Fixes a problem where the MaterialColors Lua encoder would return nothing

(also has some changes to the EncodedValue test module, I guess they weren't carried over last time we updated rbx_dom_lua?)


Dekkonot (approved):
Forgot to hit approve, sorry!

PR #783, opened 2023-09-17
Use PollWatcher in memofs StdBackend on macOS

kennethloeffler:
This should hopefully finally put to rest the extreme slowness on macOS. Taken from upliftgames/rojo#2.

Looks like it fails end to end serve tests because write events are never received...☹Interestingly, this is similar to something that @sasial-dev observed a little while ago in the OSS Discord server, I wonder if it's related? It might tank this whole PR, unfortunately


kennethloeffler:
Ok, looks like the issue is confined to serve tests only (it works as expected during real usage). I'll work on sussing out the cause

kennethloeffler:
Would it be ok to just put this behind a `#[cfg(target_os = "macos")]`, and throw some macOS testers at it once we cut the 7.4.0 prerelease? Having a lot of trouble root causing the test failure

kennethloeffler:
OK, so I got @sasial-dev to test this and serve tests pass using PollWatcher on macOS [???]

I'll gladly take this, but it's pretty spooky so we should investigate further

Dekkonot:
I am uncomfortable with this set of problems because it feels suspiciously like a race condition. 

You can read the core loop of `PollWatcher` [here](
https://github.com/notify-rs/notify/blob/49ace955830685b69c3b707057e81fa745230a7e/src/poll.rs#L50). It's very simple, and there's only a few places that platform specific behavior could be in play. These are, by my count: `std::fs`, [`walkdir`](https://crates.io/crates/walkdir), and [`filetime`](https://crates.io/crates/filetime). I'm inclined to believe these don't have critical flaws in them such that they are causing our problems; `filetime` does platform-specific syscalls but it's relied upon by Cargo, so I would hope that someone would have noticed any serious bugs by now.

My suspicion is that we're deadlocking somewhere because Github CI runners [only have 2 cores](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources). This limits the amount of threads we can run quite severely. On local machines, this limit is bypassed even if you limit the number of jobs `cargo test` can use because the serve tests spawn a child process that won't respect it. I don't have a dual core processor on hand to test with (who does?) but I suspect we'd run into the same issue regardless of OS.

Either way, I don't think this is a problem we need to worry about. I'm comfortable enough with my audit that I'm willing to approve this since I don't believe it's some deep seated systemic bug in Notify.

Dekkonot (approved):

PR #794, opened 2023-10-03
Add check to ensure plugin version matches cargo version

Dekkonot:
This modifies Rojo's build script to throw a fit if we're building a plugin with a semver incompatible version. In the process, it moves the version of the plugin to a file named `Version.txt` that's parsed at runtime. This should be minimally invasive but it's technically worse for performance than the hardcoded table and string we had before.

This feels better than a CI check or just manually verifying because it makes it physically impossible for us to forget since Rojo won't build with it being wrong.


filiptibell (approved):
LGTM - I'm guessing not checking patch version in `build.rs` is intentional?

Dekkonot:
> I'm guessing not checking patch version in `build.rs` is intentional?

Yeah my intention there is that a patch to the server probably doesn't need to happen to the plugin, so it doesn't necessarily make sense to ensure they're *always* inline. That might confuse some people if we ever end up with a patch that's out of sync but that's a price I'm willing to pay.

nezuo (approved):

plugin/src/Config.lua
```diff
@@ -2,11 +2,23 @@ local strict = require(script.Parent.strict)
...
+	end
+end
+
 return strict("Config", {
 	isDevBuild = isDevBuild,
 	codename = "Epiphany",
-	version = { 7, 3, 0 },
-	expectedServerVersionString = "7.2 or newer",
+	version = realVersion,
+	expectedServerVersionString = ("%d.%d or newer"):format(realVersion[1], realVersion[2]),
```
OverHash, line 21: It should be noted that the error message displayed to users when the `protocolVersion` is not compatible between Rojo server and plugin client will now display the latest known compatible server version, rather than the least compatible server version.

This shouldn't be an issue, but is worthwhile mentioning.

Dekkonot: My belief is that this is essentially the same thing, since if the major or minor version have been incremented, the server isn't compatible anymore on paper. It is worth noting though, thank you.

boatbomber (approved):

PR #797, opened 2023-10-04
Properly handle build metadata in semver parsing in plugin

Dekkonot:
I made an oopsie and as a result versions of the plugin with metadata won't run. This fixes that.

### DIFF
--- a/plugin/src/Config.lua
+++ b/plugin/src/Config.lua
@@ -3,7 +3,7 @@ local strict = require(script.Parent.strict)
 local isDevBuild = script.Parent.Parent:FindFirstChild("ROJO_DEV_BUILD") ~= nil
 
 local Version = script.Parent.Parent.Version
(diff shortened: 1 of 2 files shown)


plugin/src/Config.lua
```diff
@@ -3,8 +3,9 @@ local strict = require(script.Parent.strict)
 local isDevBuild = script.Parent.Parent:FindFirstChild("ROJO_DEV_BUILD") ~= nil
 
 local Version = script.Parent.Parent.Version
-local realVersion = Version.Value:split(".")
+local major, minor, patch, metadata = Version.Value:match("^(%d+)%.(%d+)%.(%d+)(.+)$")
```
Corecii, line 6: This will fail in the case of e.g. `7.4.0` because the ending does not match `(%d+)(.+)`

```suggestion
local major, minor, patch, metadata = Version.Value:match("^(%d+)%.(%d+)%.(%d+)(.*)$")
```

Corecii (changes requested):

Corecii (approved):

PR #801, opened 2023-10-17
Bump rbx_dom_lua rojo-rbx/rbx-dom@e7a5b91c (ScriptEditorService)

kennethloeffler:
This PR updates the Rojo plugin's rbx_dom_lua version. Note that it does not change the server whatsoever - this should be fine, since the only real changes are a change to scriptability of `Script.Source` and `ModuleScript.Source`, plus the custom setters/getters.


kennethloeffler:
Does this need a changelog entry?

Dekkonot (approved):

Dekkonot:
Shouldn't need a changelog entry, no!

PR #804, opened 2023-10-21
Fix PatchTree incorrect changeList entries on decode failure

kennethloeffler:
While I was testing #803, I discovered that `PatchTree.build` can create malformed `changeList` entries when a property decode fails. This causes errors when other parts of the system (such as `DomLabel`) attempt to interact with them. The core of the problem is that we're using `next` to obtain prop type-value pairs from the encoded property values in the incoming patch. `next` returns multiple values (in this case, the type followed by the value), which are then passed as-is to `addProp` or the table cons. What we want instead is to only pass the prop value.

This PR simply uses a `select` to only pass the prop value. I can also assign the result of `next` to variables in these cases if we think `select` is too ugly...

### DIFF
--- a/plugin/src/PatchTree.lua
+++ b/plugin/src/PatchTree.lua
@@ -233,7 +233,7 @@ function PatchTree.build(patch, instanceMap, changeListHeaders)
 				addProp(
 					prop,
 					if currentSuccess then currentValue else "[Error]",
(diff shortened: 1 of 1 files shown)


Dekkonot:
This seems like a fine patch to me, no need to store it in a variable.

Dekkonot (approved):

PR #805, opened 2023-10-23
Demote unapplied patch warnings to debug logs

kennethloeffler:
These warnings always appear for properties like `Capabilities`, `SourceAssetId`, etc. and tend to scare users who are syncing models. This information is now surfaced in the patch visualizer, so I think these warnings can be demoted to debug logs.


Dekkonot (approved):
I don't believe this needs a changelog entry despite being *technically* use facing. Thanks though, this was something I wanted to do but forgot. :(

PR #813, opened 2023-10-31
Add a new `syncRules` field project files to allow users to specify middleware to use for files

Dekkonot:
This PR adds the ability to specify a snapshot middleware to use for files that match a given glob, essentially allowing user-customization of Rojo's snapshotting system. For better UX, the name used for this feature is `Sync Rules` since it aligns with the concepts we use elsewhere (`Transformer Rules` would also work but I worry that it would imply we were modifying the contents of files), rather than anything mentioning middleware or snapshotting. 

An explanation of the syntax is provided in the changelog.

The bulk of this implementation is the process of generalizing the snapshot middleware to not rely upon file names so heavily. This behavior is still present in `init` files, since the semantics of changing sync rules for those is unclear (should it just be every file named `init` if it matches a middleware rule?). A `suffix` field is included to support multiple file extensions like Rojo uses, which means that everything we currently do (minus `.meta.json` files) is representable as a syncing rule.

My motivation for this feature is two-fold:

1. A more rough implementation of this is used at Uplift Games to sync our `wally.lock` and git ref information into Adopt Me for the sake of build integrity and identification. It's a blocker for us updating Rojo, and I would like to get it added here so that any subsequent changes I make at Uplift can be pushed directly to Rojo rather than to our fork.

2. The existence of syncing rules makes it more practical to change file extensions or support per-script RunContext adjustments since there's a way to maintain compatibility now. It may be worth introducing new variants of the Middleware enum for RunContext for scripts to support users specifying their own extensions to support it.

I'm not 100% on the names of the middleware, but I'm open to feedback on them.


kennethloeffler (commented):
Looking pretty ok - I need to go feed myself right now, but I'll be back shortly for a closer look at `snapshot_middleware/mod.rs`. A few comments:

src/snapshot_middleware/mod.rs
```diff
@@ -54,89 +58,245 @@ pub fn snapshot_from_vfs(
...
-            match name {
-                // init scripts are handled elsewhere and should not turn into
-                // their own children.
-                "init" | "init.client" | "init.server" => return Ok(None),
+/// Represents a possible 'transformer' used by Rojo to turn a file system
+/// item into a Roblox Instance. Missing from this list are directories and
+/// metadata. This is deliberate, as metadata is not a snapshot middleware
+/// and directories do not make sense to turn into files.
+#[derive(Debug, Clone, Copy, PartialEq, Deserialize, Serialize)]
+#[serde(rename_all = "lowercase")]
```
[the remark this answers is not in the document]
Dekkonot: I wasn't sure, because it feels weird to go `"use": "jsonModel"` but after thinking about it I agree. It looks better and it's more consistent.

src/snapshot_middleware/mod.rs
```diff
@@ -54,89 +58,245 @@ pub fn snapshot_from_vfs(
...
+    if vfs.metadata(&init_path).with_not_found()?.is_some() {
+        return Ok(Some(init_path));
+    }
+
+    Ok(None)
+}
+
+/// Gets a snapshot for a path given an InstanceContext and Vfs, taking
+/// user specified sync rules into account.
+fn snapshot_from_path(
```
kennethloeffler, line 149: ...can we remove this long if-else chain (as well as the one in `Middleware::from_path`), and handle the default syncing behavior as SyncRules instead (perhaps by adding them after we add the user-defined SyncRules in `middleware/project.rs`)? This could be a great opportunity to clean up this module!

Dekkonot: I considered this, but I was leery of it since it felt rather invasive to propose without any conversation first. If you're okay with it though, I'm more than willing to swap over because it makes things a lot more consistent and easy to change.

Any thoughts on how we should handle `init` files if we do that though?

kennethloeffler: Probably the same way is acceptable? But looking at the code more we can't just add the default rules in `snapshot_project` though... up to you if you want to change it, I just think it'd be nice to have

Dekkonot: The same way we do now would work, though it'd still involve an if-else chain. After giving it some though, I think I'm open to doing this in the future but right now it'd involve some weirdness, so I'm not sure I want to bloat this PR with it. It's already pretty big.

Corecii (changes requested):

src/snapshot_middleware/mod.rs
```diff
@@ -54,89 +58,245 @@ pub fn snapshot_from_vfs(
...
+        path: &Path,
+        name: &str,
+    ) -> anyhow::Result<Option<InstanceSnapshot>> {
+        match self {
+            Self::Csv => snapshot_csv(context, vfs, path, name),
+            Self::JsonModel => snapshot_json_model(context, vfs, path, name),
+            Self::Json => snapshot_json(context, vfs, path, name),
+            Self::ServerScript => snapshot_lua(context, vfs, path, name, ScriptType::Server),
+            Self::ClientScript => snapshot_lua(context, vfs, path, name, ScriptType::Client),
+            Self::ModuleScript => snapshot_lua(context, vfs, path, name, ScriptType::Module),
```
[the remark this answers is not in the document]
Dekkonot: My idea is that we can use `Middleware` for this going forward, yeah. It should make the underlying API more or less invisible to the rest of the project, which is ideal.

src/snapshot/metadata.rs
```diff
@@ -216,3 +247,54 @@ impl From<&Path> for InstigatingSource {
...
     }
 }
+
+/// Represents an user-specified rule for transforming files
+/// into Instances using a given middleware.
+#[derive(Debug, Deserialize, Serialize, Clone, PartialEq)]
+pub struct SyncRule {
+    /// The pattern specified by the user
+    #[serde(rename = "pattern")]
+    pub glob: Glob,
```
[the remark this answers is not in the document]
Dekkonot: I think for the general use case, `pattern` is probably better than `include` so I'm inclined to keep it named that. That said, exclusions make it easier to switch to using sync rules for our internal logic, so I'm probably going to if nothing else implement `exclude`.

src/snapshot_middleware/mod.rs
Dekkonot: Upon further thought, it turns out this is minimally invasive and actually just makes things cleaner. I've gone ahead and swapped us over to using sync rules for our default behavior, with the exception of `init` files. I imagine we could expose middleware for that too, but it would add a lot more 'default' syncing rules to iterate through so I'm inclined to not do it for the time being.

src/snapshot_middleware/mod.rs
```diff
@@ -18,30 +18,39 @@ mod toml;
...
     toml::snapshot_toml,
     txt::snapshot_txt,
-    util::PathExt,
 };
 
 pub use self::{project::snapshot_project_node, util::emit_legacy_scripts_default};
 
-/// The main entrypoint to the snapshot function. This function can be pointed
-/// at any path and will return something if Rojo knows how to deal with it.
+static DEFAULT_SYNC_RULES: OnceLock<Vec<SyncRule>> = OnceLock::new();
```
Corecii, line 48: If you think it's any better, you can stick this in the `default_sync_rules` function so that the two aren't split by the rest of the file. This might also be good because `DEFAULT_SYNC_RULES` is not meant to be used directly, it's only meant to be used indirectly by calling `default_sync_rules`.

src/snapshot_middleware/mod.rs
```diff
@@ -54,89 +63,232 @@ pub fn snapshot_from_vfs(
...
+}
 
-        let init_path = path.join("init.client.lua");
-        if vfs.metadata(&init_path).with_not_found()?.is_some() {
-            return snapshot_lua_init(context, vfs, &init_path);
-        }
+/// Gets an `init` path for the given directory.
+/// This uses an intrinsic priority list and for compatibility,
+/// it should not be changed.
+fn get_init_path<P: AsRef<Path>>(vfs: &Vfs, dir: P) -> anyhow::Result<Option<PathBuf>> {
```
Corecii, line 113: All of these have the same if block. We can make this an iteration over potential init names and run the same condition for each.

Corecii (approved):
Two potential minor improvements that aren't blockers -- fix them or leave them up to you!

src/snapshot_middleware/mod.rs
Dekkonot: The intent is we'll eventually rip this out anyway, so I'm going to decline to do it for now.

Corecii (approved):

src/snapshot/metadata.rs
```diff
@@ -144,9 +150,34 @@ impl InstanceContext {
...
+    /// Returns the middleware specified by the first sync rule that
+    /// matches the provided path. This does not handle default syncing rules.
+    pub fn get_user_sync_rule(&self, path: &Path) -> Option<&SyncRule> {
+        for rule in &self.sync_rules {
+            if rule.matches(path) {
+                return Some(rule);
+            }
+        }
+
+        None
```
chriscerie, line 179:
```suggestion
        self.sync_rules.iter().find(|&rule| rule.matches(path))
```

chriscerie (commented):
Overall looks good

Corecii (approved):

Corecii (approved):
still looks good to me!

PR #823, opened 2023-12-21
Place use of PollWatcher on mac behind a feature flag

Dekkonot:
Due to the rather large uptick in CPU usage the PollWatcher backend causes, we decided on Discord to add a flag of some kind to it.

This change adds a feature to memofs to use PollWatcher but doesn't enable it by default, allowing people to build Rojo with it if they need the performance uptick. This is in line with the discussions on Discord.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,9 +40,7 @@ name = "build"
 harness = false
 
 [dependencies]
-memofs = { version = "0.2.0", path = "crates/memofs", features = [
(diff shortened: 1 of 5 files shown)


Cargo.toml
jackTabsCode, line 1: The default behavior should be to not poll. If someone wants polling, they should build Rojo themselves with the feature enabled.

jackTabsCode (changes requested):

Issue #609, opened 2022-08-05
Slow performance in large projects


grilme99:
Sometime in the past couple of months, we have noticed significant performance regressions on the BedWars and Rift Royale codebases during syncing. We have been using Rojo `7.0.0` for over seven months, but this issue persists until the latest version. This makes me think it's related to some change in our codebase or assets.

This issue only occurs to people with an M1 Mac in our team. It can take up to 2 minutes to start the sync server for me, and syncing just sometimes stops working altogether (requiring a Rojo restart). It happens to people running through both Foreman and Rojo directly. Rojo has always been slow for our engineers on M1, but there has been a significant regression sometime recently.

Running `rojo build` only takes a few seconds.

Is there any way to profile what's happening behind the scenes to help diagnose the issue? This is having a significant impact on our workflows.

grilme99:
For some extra info, BedWars has ~30MB of binary model files and ~8MB of Lua files.

LPGhatguy:
Which build of Rojo are you running? Starting in 7.1.1, we started shipping native AArch64 (Apple Silicon) builds. I don't know if Foreman recognizes these builds, but it might be worth poking your system to see if it's downloading the right ones.

7.2.0 added a bunch of performance improvements gained from profiling on an M1 Mac. In theory, there should've been a _big_ performance boost, not a regression! 😢

You can run Rojo with `-v`, `-vv`, or `-vvv` to get more info about what Rojo is doing. That might provide some insight. Otherwise, you can download [Tracy](https://github.com/wolfpld/tracy) and build Rojo using

```rust
cargo build --release --features profile-with-tracy
```

to get a build that you can profile on your system. That should tell you fairly precisely what's taking up all that time. This is the same tooling we used to profile and optimize a few bad cases recently.

lextoumbourou:
I'm trying to debug what sounds like a similar issue today. I am running Rojo 7.2.1 using an m1. I don't think I've updated Rojo, but all of a sudden Rojo takes about 5 - 10 minutes to start. Previously it was starting in 10-20 seconds.

When I run with `-vvv` it seems to be hanging on this line:

```
[TRACE librojo::serve_session] Generating snapshot of instances from VFS
```

Boegie19:
@lextoumbourou I think I know what the issue is it is that it starts watching the files before the first creation and I have a fix ready for it it only needs to be tested since I am not 100% sure it fixes the issue

lextoumbourou:
[...]

PR #824, opened 2023-12-28
Fix broken serve tests on macOS

kennethloeffler:
Right now, serve tests will fail when Rojo is built with the FSEvent backend. The cause is essentially due to the fact that `/var` (where temporary directories for serve tests are located) on macOS is actually a symlink to `/private/var`. Paths coming from FSEvent always have symlinks expanded, but Rojo never expands symlinks. So, Rojo's paths during these tests look like `/var/*` while the FSEvent paths look like `/private/var/*`. When Rojo's change processor receives these events, it considers them outside the project and does not apply any changes, causing serve tests to time out.

To work around this, we can call `Path::canonicalize` before passing the project path to `rojo serve` during serve tests. Rojo does need to better support symlinks (which would also solve the problem), but I think that can be left for another day because it's larger in scope and I mostly just want working tests before addressing #609.

### DIFF
--- a/tests/rojo_test/serve_util.rs
+++ b/tests/rojo_test/serve_util.rs
@@ -66,7 +66,11 @@ impl TestServeSession {
 
         let source_path = Path::new(SERVE_TESTS_PATH).join(name);
         let dir = tempdir().expect("Couldn't create temporary directory");
-        let project_path = dir.path().join(name);
+        let project_path = dir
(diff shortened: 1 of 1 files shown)


Dekkonot (approved):
That is an unexpectedly annoying problem, but I agree that symlink support is best left for another day.

PR #826, opened 2023-12-30
Fix move_folders_of_stuff

kennethloeffler:
This is a fairly important test verifying whether the action of moving a folder into a watched folder is correctly detected and processed. It was disabled in https://github.com/rojo-rbx/rojo/commit/b43b45be8f2eb01161a389afeab56749fa82e459. The fact that it failed indicates a possible bug in change processing, so in this PR, I'll re-enable the test, investigate why it fails, and fix it.

### DIFF
--- a/src/change_processor.rs
+++ b/src/change_processor.rs
@@ -1,6 +1,5 @@
 use std::{
     fs,
-    path::PathBuf,
     sync::{Arc, Mutex},
 };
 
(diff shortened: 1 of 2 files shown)


kennethloeffler:
After some testing on a Linux machine, I determined that this failed because notify's inotify backend causes memofs to emit Write events as a result of folder moves (which also happens to be the case for the FSEvents backend), but right now, Rojo's change processor explicitly does not handle Write events for directories: 
https://github.com/rojo-rbx/rojo/blob/11fa08e6d6962a3c54ddd16ecea552f8407bc066/src/change_processor.rs#L126-L142

From what I can gather reading the OSS Discord serve around this time, it seems like this behavior was introduced in https://github.com/rojo-rbx/rojo/commit/17de912608a8d728e1879bf8ef6229e296ae1a1d as an attempted performance optimization. It breaks folder move detection under inotify and FSEvents, and I think this sort of optimization would be far better done with debouncing in memofs rather than silently ignoring certain events in the change processor.

Dekkonot (approved):

PR #827, opened 2023-12-31
Conserve CI minutes via cache, skip macOS+Windows MSRV builds

kennethloeffler:
Windows and macOS runners consume GitHub Actions minutes at [2x and 10x the rate of Linux runners, respectively](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#minute-multipliers). This is a bit concerning now that there are two Windows jobs and two macOS jobs, introduced in #825. 

This PR aims to reduce the cost by:
* Adding [rust-cache](https://github.com/Swatinem/rust-cache/) to reduce the amount of time spent. I'm aware there were some concerns raised about CI caches in general in #496 - are they still a blocker?
* Removing the unnecessary Windows and macOS MSRV build jobs. If an MSRV build fails on one platform due to usage of new language features, then it will fail on all of them.

@Kampfkarren may have to change this repository's required status checks before this PR can be merged

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,6 +19,16 @@ jobs:
         rust_version: [stable, 1.70.0]
         os: [ubuntu-latest, windows-latest, macos-latest]
 
+        # It's unnecessary to build using MSRV on macOS and Windows
+        # because MSRV builds will not fail on a platform specific
+        # basis. So, we skip these two configurations to conserve
+        # GitHub Actions minutes.
+        exclude:
+          - os: macos-latest
+            rust_version: 1.70.0
+          - os: windows-latest
+            rust_version: 1.70.0
+
     steps:
     - uses: actions/checkout@v3
       with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,6 +87,9 @@ jobs:
         override: true
         components: rustfmt, clippy
 
+    - name: Rust cache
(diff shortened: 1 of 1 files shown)


Dekkonot (approved):

Dekkonot:
I think you're likely right about Kampf needing to change the required status checks, but this looks good to me.

That said, in the last commit to the master branch the CI took 107 minutes when you account for the multipliers, which is 5.3% of our monthly budget. It's definitely a good idea to cache these but this PR taking a day or two to get merged shouldn't be a disaster.

Dekkonot:
Actually, upon further inspection I'm a big concerned about the cache size here. rust-cache caches dependencies, but on my system the dependency build files are 8GB. Given the max size of a cache is 10GB, I'm not sure if this will even be particularly effective.

It might be worth just eating the multiplier on windows and caching on MacOS? Unsure.

OverHash:
> Actually, upon further inspection I'm a big concerned about the cache size here. rust-cache caches dependencies, but on my system the dependency build files are 8GB. Given the max size of a cache is 10GB, I'm not sure if this will even be particularly effective.

The ubuntu test logs indicate that the cache is only ~510mb, so we should be good across all the checks.
https://github.com/rojo-rbx/rojo/actions/runs/7368716794/job/20053268423?pr=827
```
 Cache Size: ~510 MB (534911718 B)
Cache saved successfully
```

PR #816, opened 2023-11-11
Upgrade to Notify 6

jackTabsCode:

PR #830, opened 2024-01-02
Downgrade to Notify 4, use FSEvents, use minimal recursive watches

### LINKED ISSUE #609: Slow performance in large projects

### LINKED ISSUE #815: High CPU usage on MacOS
CPU usage hovering around 33%. 
image

Some notes:
- macOS Sonoma 14.1.1
- Studio is not running
- Persistent after a fresh restart
- Memory usage is low
- Rojo version 7.4.0-rc.3. Does not occur on 7.3.0
- MacBook M2 Pro

kennethloeffler:
This PR makes a few changes to the file watching component of memofs' `StdBackend` to address performance problems on macOS. Closes #815 and closes #609. The amount of code churn here is unfortunate, but we haven't had a maintainer who works on macOS until now.

### Downgrade to Notify 4
First, I downgraded back to Notify 4 by reverting #816. The reason for this is the fact that Notify 6's debouncer provides incorrect results when using the FSEvents backend. Specifically, FSEvents under some circumstances emits a Create event immediately followed by a Remove event for file deletions [???], which Notify 6's debouncer then collates into zero events, never reporting any changes. Notify 4's debouncer does not have this problem.

Because the raw events can be quite noisy, we do need some form of debouncing to avoid unnecessary patch re-computations. So, we're blocked on upgrading Notify unless memofs implements its own deboucing, or we maintain our own patch until it's fixed upstream.

### Use FSEvents backend instead of PollWatcher
Next, I reverted #783. The use of `PollWatcher` worked around the massive latency described in #609, but incurred prohibitive cost as reported in #815.

### Minimize number of watch calls by leveraging Notify's recursive watch mode
I determined that the performance problem in #609 is caused by excessive calls to `FsEventWatcher::watch`. This method is potentially very expensive, and `rojo serve` calls it for every file in the project during initialization. It seems to get slower and slower the more it is called, which starts to be problematic once the number of files in the project exceeds about 1000. I profiled `rojo serve` using a debug build on a project containing 2000 Lua files, and calls to `FsEventWatcher::watch` dominated the profile, accounting for over 90% of the execution time:

image

It's possible to avoid this by using Notify's recursive watch mode, and only calling `watch` on files that aren't already covered by an existing recursive watch. FSEvents is designed to watch an entire hierarchy - in other words, it will report events for any descendant of a watched directory. By using the recursive watch mode, we only have to call `watch` once per top-level directory, and once per "naked" file, i.e. a file that is directly referred to using `$path` that is not already a descendant of a watched directory. With this change, calls to `FsEventWatcher::watch` virtually disappear in the profile:

image

### Takeaways and alternatives?
The main takeaways here are that the underlying file watching APIs are racy, unreliable, and possibly very costly. It's also important to note that changes to memofs are hazardous, and must be tested thoroughly on each of Rojo's main supported platforms.

I looked into using the kqueue backend on macOS (which is far better behaved compared FSEvents), but it's still rather slow, and presents an additional issue: the kqueue backend opens every file it watches and stores the file descriptors, which risks exceeding the operating system's maximum open files limit. While this limit can be increased, I'd rather not have to worry about it all.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Rojo Changelog
 
 ## Unreleased Changes
+* Significantly improved performance of `rojo serve`, `rojo build --watch`, and `rojo sourcemap --watch` on macOS. [#830]
+
+[#830]: https://github.com/rojo-rbx/rojo/pull/830
 
 ## [7.4.0-rc3] - October 25, 2023
 * Changed `sourcemap --watch` to only generate the sourcemap when it's necessary ([#800])
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -138,7 +138,6 @@
 * Add buttons for navigation on the Connected page ([#722])
 
 ### Fixes
-* Significantly improved performance of `rojo serve` and `rojo build` on macOS. [#783]
 * Significantly improved performance of `rojo sourcemap` ([#668])
 * Fixed the diff visualizer of connected sessions. ([#674])
 * Fixed disconnected session activity. ([#675])
@@ -172,7 +171,6 @@
 [#770]: https://github.com/rojo-rbx/rojo/pull/770
 [#771]: https://github.com/rojo-rbx/rojo/pull/771
 [#774]: https://github.com/rojo-rbx/rojo/pull/774
-[#783]: https://github.com/rojo-rbx/rojo/pull/783
 [rbx-dom#299]: https://github.com/rojo-rbx/rbx-dom/pull/299
 [rbx-dom#296]: https://github.com/rojo-rbx/rbx-dom/pull/296
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1852,7 +1852,6 @@ dependencies = [
  "log",
  "maplit",
  "memofs",
- "notify",
  "num_cpus",
  "opener",
  "paste",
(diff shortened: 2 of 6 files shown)


jackTabsCode:
> The reason for this is the fact that Notify 6's debouncer provides incorrect results when using the FSEvents backend.
Perhaps we should get in contact with the Notify maintainers?

kennethloeffler:
I don't think we materially lose much by sticking with Notify 4 for now, but yes, the problem ought to be raised with Notify's maintainers

Dekkonot (approved):
Thank you for taking the time to handle this! The codechurn is bad, but I'm glad we're actually solving this problem.

I imagine this also improves performance on Windows and Linux, but that's just a guess.

I do agree with Jack that having to downgrade Notify feels like a loss in general but I'm not terribly concerned about it. While we should report this to the Notify maintainers, I don't want to wait on them to fix it before we can make a new release and as far as I'm aware there's nothing wrong with Notify 4 beyond it being old.

We can revisit an upgrade to Notify later though.

PR #831, opened 2024-01-02
Changed file extensions of lua to luau

Barocena:
This PR Changes *.lua files that init command generates to *.luau 
and updated tests/snapshots


Dekkonot:
Changelog but otherwise I'm fine with this

Dekkonot (approved):

PR #832, opened 2024-01-05
Check for compatible updates in plugin

boatbomber:
Adds a setting to checking for updates.
image

When a newer compatible version is detected, it gives the user a notification. It will not notify about major version updates, since those are potentially breaking and the user may have reason to remain on the older version.

image

When the setting is enabled in a local install (ie: not the marketplace published plugin), it also adds a setting for being notified about prereleases (defaults to off).

image

image

This should hopefully reduce the number of reports we get from people simply using outdated versions.


Dekkonot:
I'm of two minds here.

The first loves this because it'd serve as a reminder for people to actually update their dang Rojo releases.

The second thinks this'll be annoying and that prompting people for pre-releases might be a problem since we don't plan to publish the plugin to the website for those versions. I appreciate how you can turn it off but I'm not sure how much I like a feature that my first thought was "I hope you can turn it off" for.

boatbomber:
Well, @Dekkonot, it's disabled by default. It's less "I hope I can turn this off" for most people and more "I would like to enable this" for some people.

boatbomber:
If you install Rojo via marketplace, the Include Prerelease setting is hidden to avoid confusion.

Dekkonot (approved):
I've reconsidered this and I think we should include it, haha.

PR #834, opened 2024-01-12
Visualize table changes

### LINKED ISSUE #758: Support better visualization of `Instance.Attributes` in patch visualizer
Currently, `Instance.Attributes` is visualized as a dictionary of attribute values.

This is not good and we should display it better, since a sufficiently large dictionary is unreadable.

boatbomber:
Closes #758.
Implements a pop out diff view for table properties like Attributes and Tags.

Dictionary Diff (eg Attributes)
image

Array Diff (eg Tags)
image


Dekkonot (approved):
There are some potential efficiency concerns with the actual diffing process for tables, but conveniently that only happens when a user clicks the button so I'm not actually concerned! Thank you!

PR #840, opened 2024-01-30
Play Solo & Test Server auto connect

boatbomber:
image

When enabled, the `baseurl` of the session is written to `workspace:SetAttribute("__Rojo_ConnectionUrl")` so that the test server can connect to that session automatically.

This works for Play Solo and Local Test Server. It is marked experimental for now (and disabled by default) since connecting during a playtest session is... not polished. Rojo may overwrite things and cause headaches. Further work can be done later, this PR just gets the feature out since #468 has been sitting for 3 years already.


jackTabsCode:
Tested it on my end, seems to work flawlessly.

Dekkonot (approved):

PR #843, opened 2024-01-31
Support setting referent properties via attributes

### LINKED ISSUE #427: How to write Ref or Refs?
In a `.model.json` file, how do I write a Ref? I checked the documentation and the documentation doesn't provide guidance.

Consider a file `Rat.model.json`
```json
{
  "Name": "Rat",
  "ClassName": "Model",
  "Properties": {
    "PrimaryPart": "Torso"
  },
  "Children": [
    {
      "Name": "Torso",
      "ClassName": "Part",
      "Properties": {
        "Size": {
          "Type": "Vector3",
          "Value": [1, 1, 3]
        }
      }
    },
    {
      "Name": "Humanoid",
      "ClassName": "Humanoid"
    }
  ]
}
```

This fails with `Wrong type of value for property Model.PrimaryPart. Expected Ref, got a string`
How do I Ref "Torso"?

I am using Rojo v7.0.0-alpha.4

Dekkonot:
Ref properties have been neglected by Rojo for a long time, in no small part because of the potential complexity of their implementation. We've gotten by since they've been handled in `rbxm` and `rbxmx` files just fine but we should do better. This PR does the groundwork for us to do exactly that, by adding a system for manually specifying Ref properties with IDs and attributes.

I will be the first to say this: the UX of this approach is bad. However, I don't think there's actually a better way to support this universally. If an `rbxm` file needs to refer to a `model.json` Instance, that 'pointer' has to be stored inside the file somehow, which gives us relatively few options. Attributes are the only ones that are arbitrary key-value pairs, so it makes the most sense to use them. As a bonus, they persist through saving and loading into Studio so they'll be suitable for syncback and eventually two way sync.

Long-term, I want to support using IDs directly as properties in JSON files because it's a much better UX. I didn't want to do that in this PR though, because it will involve altering property resolution in some capacity and that's a potentially invasive change that I want to make stand-alone.

## Usage

An ID for an Instance can be specified by putting an `id`/`$id` field in any of the JSON file formats, or by setting an attribute named `Rojo_Id` to a string.

This ID can be referred to by other Instances to set a Ref property via an attribute named `Rojo_Target_PROP_NAME`, where `PROP_NAME` is the name of a property. The attribute should be a string that references an ID defined elsewhere in the file. You can see an example of that in the tests files I included. Again, I don't love this UX but if anyone has better suggestions I'm happy to hear them.

## Design Notes

`RojoRef` has been added and is a newtype surrounding `Arc<String>`. We may want to cache these eventually (at which point we could probably swap to using `SharedString` under the hood, though it's a `Vec<u8>`) but right now it feels like overkill to do so. However, they are cloned a fair bit so using `Arc` feels like the right call.

They're currently loaded during patch application because it's straightforward and attributes need to be traversed during that process anyway. I realize this might not be the best option though, so I do have an idea of how to do it 'properly' during the snapshotting process. It'd be a lot more invasive than this change though, so I defaulted to this one just to be safe.

---

Closes #427.


Dekkonot:
Hm. The aftman timeout is concerning.

Dekkonot:
Ok whatever it was, wasn't me. 👍🏼

src/snapshot/tree.rs
```diff
@@ -137,6 +141,14 @@ impl RojoTree {
...
                     }
                 }
+                if existing_metadata.specified_id != metadata.specified_id {
+                    if let Some(old) = &existing_metadata.specified_id {
+                        self.specified_id_to_refs.remove(old);
+                    }
+                    if let Some(new) = &metadata.specified_id {
+                        self.specified_id_to_refs.insert(new.clone(), id);
+                    }
+                }
```
kennethloeffler, line 151: This needs a duplicated ID check like `RojoTree::insert_metadata`. Failing the check should probably result in an error message instead of a warning. Also, with the current implementation, a user specifying duplicate IDs can leave `specified_ids_to_refs` in an incorrect state:
```rust
                if existing_metadata.specified_id != metadata.specified_id {
                    // Adding, then removing a duped ID during a serve session 
                    // clobbers the preexisting mapping - this snippet should
                    // not run if we detect a duped ID
                    if let Some(old) = &existing_metadata.specified_id {
                        self.specified_id_to_refs.remove(old);
                    }
                    if let Some(new) = &metadata.specified_id {
                        self.specified_id_to_refs.insert(new.clone(), id);
                    }
                }
```

src/snapshot/tree.rs
```diff
@@ -161,11 +173,32 @@ impl RojoTree {
...
+    }
+
     fn insert_metadata(&mut self, id: Ref, metadata: InstanceMetadata) {
         for path in &metadata.relevant_paths {
             self.path_to_ids.insert(path.clone(), id);
         }
 
+        if let Some(specified_id) = &metadata.specified_id {
+            if self.get_specified_id(specified_id).is_some() {
+                log::warn!("Duplicate user-specified referent {specified_id}")
```
kennethloeffler, line 196:
```suggestion
                log::error!("Duplicate user-specified referent {specified_id}")
```

src/snapshot/tree.rs
```diff
@@ -174,6 +207,10 @@ impl RojoTree {
     fn remove_metadata(&mut self, id: Ref) {
         let metadata = self.metadata_map.remove(&id).unwrap();
 
+        if let Some(specified) = metadata.specified_id {
+            self.specified_id_to_refs.remove(&specified);
+        }
+
```
kennethloeffler, line 213: This probably needs a duplicate ID check as well?

kennethloeffler (changes requested):
Awesome, this has been a long-needed feature. It's looking mostly sound - a few comments:

kennethloeffler:
I'm a little concerned that there are two different ways to specify IDs... it could be tricky to continue to ensure that everything stays consistent. This is mostly caused by the requirement to use attributes to represent refs (we absolutely need this for cross-model refs, right?), so I'm not sure there's much we can do to mitigate this :(

src/snapshot/tree.rs
kennethloeffler: On second thought.... no, that doesn't make sense 🤪

src/rojo_ref.rs
```diff
@@ -0,0 +1,40 @@
...
+use serde::{Deserialize, Serialize};
+
+pub const REF_ID_ATTRIBUTE_NAME: &str = "Rojo_Id";
+pub const REF_POINTER_ATTRIBUTE_PREFIX: &str = "Rojo_Target_";
+
+// TODO add an internment strategy for RojoRefs
+// Something like what rbx-dom does for SharedStrings probably works
+
+#[derive(Debug, Default, PartialEq, Hash, Clone, Serialize, Deserialize, Eq)]
+pub struct RojoRef(Arc<Vec<u8>>);
```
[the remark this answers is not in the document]
Dekkonot: The primary reason was that user-defined attributes are BinaryStrings. The idea being that `String::from_utf8` is fallible and patch application isn't. In hindsight though, I don't know why I did it that way and it makes more sense to raise a warning or error. I think an error is probably better, so I'll revert the change and just try to convert user attributes into strings.

Do you think it's worth raising an actual error, which involves making `apply_patch_set` return a `Result` and handling that (I haven't looked at how much work that'd be, it might be zero) or should I just log it as an error?

src/snapshot/tree.rs
Dekkonot: Good catch, I'll add a check.

src/snapshot/tree.rs
kennethloeffler: On third thought (sorry for the wavering, this was a last-minute addition)...

If I'm thinking about this right, this *could* cause an inappropriate removal from `specified_id_to_refs` when the ID is a duplicate. We should check that the entry in the map actually corresponds to the instance we're removing metadata from.

Dekkonot: You're correct, this could cause some problems. We need to uphold the invariant that there can't be duplicates and this doesn't at the moment. I'll fix that.

Dekkonot:
> I'm a little concerned that there are two different ways to specify IDs... it could be tricky to continue to ensure that everything stays consistent. This is mostly caused by the requirement to use attributes to represent refs (we absolutely need this for cross-model refs, right?), so I'm not sure there's much we can do to mitigate this :(

We need something like it. Since we want to be able to support pointing into and out of `rbxm`s, there are very few options. The most obvious one is UniqueId, which we could use for Instance IDs, but that doesn't give us an easy way to specify properties still. Since we have to use attributes either way, we may as well use it for both.

I think it's probably fine to support both ways though. We have two ways to specify attributes in most if not all places, but that's not a real concern since people don't generally use both. It doesn't cause problems unless a user makes it a problem. If you're concerned we can make it an error if you do both though!

src/snapshot/tree.rs
```diff
@@ -161,11 +179,35 @@ impl RojoTree {
         self.metadata_map.get(&id)
     }
 
+    pub fn get_specified_id(&self, specified: &RojoRef) -> Option<Ref> {
+        match self.specified_id_to_refs.get(specified)[..] {
+            [referent] => Some(referent),
+            _ => None,
+        }
+    }
```
kennethloeffler, line 187:
```suggestion
    /// Get the backing Ref of the given RojoRef. If the RojoRef maps to exactly 
    /// one Ref, this method returns Some. Otherwise, it returns None.
    pub fn get_specified_id(&self, specified: &RojoRef) -> Option<Ref> {
        match self.specified_id_to_refs.get(specified)[..] {
            [referent] => Some(referent),
            _ => None,
        }
    }
```

src/snapshot/patch_apply.rs
```diff
@@ -208,9 +232,72 @@ fn apply_update_child(context: &mut PatchApplyContext, tree: &mut RojoTree, patc
...
+    for (attr_name, attr_value) in attributes.iter() {
+        if attr_name == REF_ID_ATTRIBUTE_NAME {
+            if let Variant::String(specified_id) = attr_value {
+                attr_id = Some(RojoRef::new(specified_id.clone()));
+            } else if let Variant::BinaryString(specified_id) = attr_value {
+                if let Ok(str) = std::str::from_utf8(specified_id.as_ref()) {
+                    attr_id = Some(RojoRef::new(str.to_string()))
+                } else {
+                    log::error!("Specified IDs must be valid UTF-8 strings.")
+                }
```
kennethloeffler, line 265: Do we actually need these branches for BinaryString anymore? Since non UTF-8 strings are now unrepresentable with RojoRef, I think we could just have one branch for `String`

kennethloeffler:
Let's also make sure to close the relevant issue for this feature: https://github.com/rojo-rbx/rojo/issues/427

src/snapshot/patch_apply.rs
Dekkonot: rbx-types reads String attributes as BinaryStrings, so yes, we do.

kennethloeffler:
I think we're pretty much done here - just need to fix angry rustfmt, and I'd like to try to break this one more time. I also noticed a few Ref sync bugs during the course of my testing, so I'll spin off to document/fix them shortly

kennethloeffler:
Okay, so I can't find any more issues directly related to this PR, but there remain some problems one can run into when using the feature. 

The first one I found is related to the order of addition operations within the Rojo plugin's reconciler:
https://github.com/rojo-rbx/rojo/blob/5c4260f3ac6f967fe380aec2867e2eff84660709/plugin/src/Reconciler/applyPatch.lua#L40-L79

There's a problem at line 73. In this loop, we reify each incoming added instance one after the other, thereby creating each one with `Instance.new` and applying any properties (which includes Refs). It's possible to break this by having two sibling instances, and creating a Ref property on one that points to the other:
```json
{
  "name": "big-bad-refs",
  "tree": {
    "$className": "DataModel",
    "ReplicatedStorage": {
      "MyCoolPart": {
        "$className": "Part",
        "$id": "MyCoolPart"
      },
      "ObjectValue": {
        "$className": "ObjectValue",
        "$attributes": {
          "Rojo_Target_Value": "MyCoolPart"
        }
      }
    }
  }
}
```

When I sync this project into Roblox Studio, it fails to apply `ObjectValue.Value`. If I disconnect and reconnect, then the property applies correctly, suggesting that the cause is related to the order of operations (i.e. `MyCoolPart` doesn't yet exist at the time at the time of the ObjectValue's reification). For correctness, we should be creating *all* added instances before attempting to apply Ref properties. We'll probably need to decouple instance creation and property application in this case.

We can either merge this PR, document the issue separately, and fix it sometime later, or else we can fix it right here in this PR. I'd go with the former (we don't want to add too much bloat), but I'll leave the final decision up to you.

Dekkonot:
I'm inclined to agree with you about not touching the plugin code with this PR. I'm inclined to say we need to change that code anyway to support async in reify (for MeshParts/Unions) so it'll be fine to wait on it.

Where do you think we should document that? Intuition says an issue is the right place.

kennethloeffler:
Yes, an issue will be fine

Dekkonot:
Looking to get this merged soon if possible since it's a blocker on getting syncback closer to merging upstream @kennethloeffler

kennethloeffler (approved):
LGTM, thanks for the heads up!

PR #846, opened 2024-02-01
Lint plugin src

boatbomber:
This has been on my pet peeves list for a while.

Enforces selene via CI. This PR comes with a whole bunch of tiny fixes for things that make my editor yell at me whenever I work on Rojo.

Also, after fixing lints on `createSignal` I noticed that it's totally unused. It's 4 years old with the last commit being "for future work" which evidently never came. So, I removed it.


Dekkonot (approved):

Dekkonot:
Next step: static types, right? :P

boatbomber:
We'll drag Rojo into the modern era kicking and screaming if we have to

PR #847, opened 2024-02-01
Catch failed http requests that didn't error so we can handle them correctly

boatbomber:
The `http.get` promise treated any request that made it through the `pcall` without erroring as a success to `resolve`.
`ApiContext` would then check the status code and `reject`, but by then it's too late to handle gracefully and we just show the error code and stop the sync.

This improves that. We create a relevant `Error` object for requests that are not success codes, so that `ApiContext` can handle them correctly.

The immediate impact is that GitHub Codespaces is now compatible with Rojo, since requests to the [forwarded port](https://docs.github.com/en/codespaces/developing-in-a-codespace/forwarding-ports-in-your-codespace) receive a [504 Gateway Timeout](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/504) after 100 seconds. Given our long polling, we simply resend requests that return a timeout.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * Added popout diff visualizer for table properties like Attributes and Tags ([#834])
 * Updated Theme to use Studio colors ([#838])
 * Added experimental setting for Auto Connect in playtests ([#840])
+* Fixed http error handling so Rojo can be used in Github Codespaces ([#847])
 * Projects may now specify rules for syncing files as if they had a different file extension. ([#813])
  	This is specified via a new field on project files, `syncRules`:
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 * Added popout diff visualizer for table properties like Attributes and Tags ([#834])
 * Updated Theme to use Studio colors ([#838])
 * Added experimental setting for Auto Connect in playtests ([#840])
+* Fixed http error handling so Rojo can be used in Github Codespaces ([#847])
```
Dekkonot, line 7: I'm not sure I think this is the best entry (it feels like working with GitHub Codespaces is a side effect) but we can revisit it before the next release if I still care.

Dekkonot (approved):

CHANGELOG.md
boatbomber: While sorta a side effect, it was the reason and impact for this PR.

PR #850, opened 2024-02-02
Add benchmarking, perf gains, and better settings UI

boatbomber:
- Mark dev debug settings like we mark experimental settings
- Create dev setting to enable timing logs
  - When timing is disabled, timer functions are replaced with a no-op to minimize overhead 
- Add timing logs to major internal actions
- Make patchTree notably faster in a few key areas
- Make confirming page's patch visualizer reduce wasteful work

Also fixed a pet peeve of mine- the back button in the settings menu no longer scrolls off screen so you can now go back without having to scroll back up

image

### DIFF
--- a/plugin/src/PatchTree.lua
+++ b/plugin/src/PatchTree.lua
@@ -134,6 +134,7 @@ function Tree:addNode(parent, props)
 		for k, v in props do
 			node[k] = v
 		end
+		Timer.stop()
 		return node
 	end
 
@@ -144,6 +145,7 @@ function Tree:addNode(parent, props)
 	local parentNode = self:getNode(parent)
 	if not parentNode then
 		Log.warn("Failed to create node since parent doesnt exist: {}, {}", parent, props)
+		Timer.stop()
 		return
 	end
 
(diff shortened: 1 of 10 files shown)


plugin/src/PatchTree.lua
```diff
@@ -142,22 +145,26 @@ function Tree:addNode(parent, props)
...
 
 -- Given a list of ancestor ids in descending order, builds the nodes for them
 -- using the patch and instanceMap info
 function Tree:buildAncestryNodes(previousId: string?, ancestryIds: { string }, patch, instanceMap)
+	Timer.start("Tree:buildAncestryNodes")
 	-- Build nodes for ancestry by going up the tree
 	previousId = previousId or "ROOT"
 
-	for _, ancestorId in ancestryIds do
+	for i = #ancestryIds, 1, -1 do
```
Dekkonot, line 166: What was the motivation behind this change? Feels really arbitrary, haha.

plugin/src/Timer.lua
```diff
@@ -0,0 +1,57 @@
...
+local clock = os.clock
+
+local Timer = {
+	_entries = {},
+}
+
+function Timer._start(label)
+	local start = clock()
+	if not label then
+		error("[Rojo-Timer] Timer.start: label is required")
```
Dekkonot, line 12:
```suggestion
		error("[Rojo-Timer] Timer.start: label is required", 2)
```

plugin/src/Timer.lua
```diff
@@ -0,0 +1,57 @@
...
+
+	table.insert(Timer._entries, { label, start })
+end
+
+function Timer._stop()
+	local stop = clock()
+
+	local entry = table.remove(Timer._entries)
+	if not entry then
+		error("[Rojo-Timer] Timer.stop: no label to stop")
```
Dekkonot, line 24:
```suggestion
		error("[Rojo-Timer] Timer.stop: no label to stop", 2)
```

Dekkonot (commented):
Looks good, minor nitpicks with the Timer module and a question about a change.

plugin/src/PatchTree.lua
boatbomber: Creating the array backwards and iterating forwards means the build step constantly shifts items down as it has to insert at the start, which this avoids by building forward and just iterating backwards.

Dekkonot (approved):

PR #854, opened 2024-02-06
Normalize line endings to LF in Lua middleware

### LINKED ISSUE #776: Rojo should remove Carriage Returns when syncing on Windows
Today when I tried to sync into my project's team create place, I encountered an interesting issue—Rojo wants to modify every single source file in the game.

image

I was freaking out a bit 😅, but as it turns out, when you sync a Rojo project with Windows, it uses CRLF characters for new lines. I was syncing over a team create place on my Macbook over my coworker's changes earlier.

I think Rojo should always convert CRLF to LF when syncing, or at very least omit these changes in the sync preview. This is because on Roblox's built-in script editor, hitting "Enter" on any device always creates a Line Feed with no Carriage Return character. This is also reflected in the default ModuleScript:
```lua
local module = {}

return module

```

As it turns out, even when *pasting* code with carriage returns into roblox's script editor, the carriage returns are removed. So it seems that the only possible way for a carriage return to end up in a roblox script is by directly modifying the Source property of the script (i.e. syncing via Rojo).

Because Roblox standardizes LF in their own script editor, Rojo should also convert CRLF to LF. On windows, git can be configured with the `core.autocrlf` value, which always uses CRLF locally, and LF on the git repository itself. VS Code has CRLF enabled by default for new files on windows as well, so even if a git repo uses Line Feeds, it will be converted to CRLF on windows.

I actually looked through some free models, and you can easily identify which ones were synced using a Rojo codebase by searching for a "\13\10" character sequence in the script Source properties. When copying/pasting the code in rojo-synced code to another module, it gets sanitized back to LF.

kennethloeffler:
Closes #776 by implementing a new method `Vfs::read_to_string_lf_normalized` and changing the Lua middleware to use it.

I'm not really sure why the middlewares were calling `to_owned` on `contents_str` - could it have been a limitation of earlier rbx-dom versions?

### DIFF
--- a/src/snapshot_middleware/txt.rs
+++ b/src/snapshot_middleware/txt.rs
@@ -13,7 +13,7 @@ pub fn snapshot_txt(
     path: &Path,
     name: &str,
 ) -> anyhow::Result<Option<InstanceSnapshot>> {
-    let contents = vfs.read_to_string_lf_normalized(path)?;
+    let contents = vfs.read_to_string(path)?;
     let contents_str = contents.as_str();
 
     let properties = hashmap! {
--- a/src/snapshot_middleware/txt.rs
+++ b/src/snapshot_middleware/txt.rs
@@ -1,6 +1,5 @@
 use std::{path::Path, str};
 
-use anyhow::Context;
 use maplit::hashmap;
 use memofs::{IoResultExt, Vfs};
 
@@ -14,10 +13,8 @@ pub fn snapshot_txt(
     path: &Path,
     name: &str,
 ) -> anyhow::Result<Option<InstanceSnapshot>> {
-    let contents = vfs.read(path)?;
-    let contents_str = str::from_utf8(&contents)
-        .with_context(|| format!("File was not valid UTF-8: {}", path.display()))?
-        .to_owned();
+    let contents = vfs.read_to_string_lf_normalized(path)?;
+    let contents_str = contents.as_str();
 
     let properties = hashmap! {
         "Value".to_owned() => contents_str.into(),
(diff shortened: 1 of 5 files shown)


kennethloeffler:
... is this breaking? :( 

User code could theoretically be matching against only CRLF line endings currently (I'm mostly thinking about text files/StringValues - this is moot for scripts since Roblox now aggressively normalizes scripts to LF). I guess such code is busted regardless...

src/snapshot_middleware/txt.rs
```diff
@@ -14,10 +13,8 @@ pub fn snapshot_txt(
     path: &Path,
     name: &str,
 ) -> anyhow::Result<Option<InstanceSnapshot>> {
-    let contents = vfs.read(path)?;
-    let contents_str = str::from_utf8(&contents)
-        .with_context(|| format!("File was not valid UTF-8: {}", path.display()))?
-        .to_owned();
+    let contents = vfs.read_to_string_lf_normalized(path)?;
+    let contents_str = contents.as_str();
```
Dekkonot, line 17: I'm inclined to say that normalizing line endings for StringValues is a bad call since Roblox themselves don't, so we'd be inventing behavior. The reason we're bothering with scripts is that Roblox does it and we want to match their behavior.

kennethloeffler: Right - it's just that letting CRLF line endings through for text files could bite teams that don't all work on the same OS. The only thing that could possibly go "wrong" is annoying diffs, and they could (and definitely should) configure their VCS to normalize line endings. 

I'm open to leaving it out of this PR, but we should stick a pin into it in case anyone runs into this issue again.

Dekkonot (approved):

Issue #859, opened 2024-02-14
Migrate to using Rustls instead of OpenSSL


filiptibell:
Currently, Rojo depends on `reqwest`, which in turn depends on `libssl` (OpenSSL) for the `Rojo upload` command: https://github.com/rojo-rbx/rojo/blob/8ff064fe28a5acd4db3c411f8912ee53bbea7090/src/cli/upload.rs#L7-L10

A frequent issue for new users getting started with Rojo has been version mismatches with this dependency (https://github.com/rojo-rbx/rojo/issues/585), and Rojo has made experimental releases in the past ([v7.2.1-static-openssl](https://github.com/rojo-rbx/rojo/releases/tag/v7.2.1-static-openssl)) that improved on but have not fully solved this issue.

Other tools such as [Wally](https://github.com/UpliftGames/wally) also depend on `libssl`, but depend on a ***different version***, leading to users having issues installing either one or the other, or *globally installing multiple openssl versions* just to be able to use ecosystem tooling, which is heavily discouraged for security reasons.

I propose that Rojo migrate to using [Rustls](https://github.com/rustls/rustls) to resolve these building & distribution issues. This is a Rust-native dependency that would resolve all of the abovementioned issues, and has a near-flawless security record, while OpenSSL does not.

PR #861, opened 2024-02-14
Migrate to using Rustls

filiptibell:
Implements #859


Dekkonot (approved):
We'll have to keep an eye out for any changes this causes, but I'm willing to merge this. It feels like a no brainer change.

Issue #858, opened 2024-02-14
Optional top-level default.project.json "name" field


osyrisrblx:
If a user does not fill in the "name" field of default.project.json file, it should default to the name of the parent folder.

---

To provide more context: this would be useful for roblox-ts npm packages.

Imagine as a user you:
1. Clone roblox-ts package template, this creates a package.json + default.project.json + etc.
2. You set the name of your package in package.json
3. You forgot to rename "name" in default.project.json!
4. Now when we load it into Studio, it has some generic name on it.

roblox-ts _can_ resolve this path correctly by looking in default.project.json..

but it could cause conflicts between multiple packages with the same default.project.json "name" field

PR #864, opened 2024-02-15
Optional Project Name Prototype

osyrisrblx:
#858

### DIFF
--- a/src/project.rs
+++ b/src/project.rs
@@ -39,6 +39,7 @@ enum Error {
 #[serde(deny_unknown_fields, rename_all = "camelCase")]
 pub struct Project {
     /// The name of the top-level instance described by the project.
+    #[serde(default)]
     pub name: String,
 
     /// The tree of instances described by this project. Projects always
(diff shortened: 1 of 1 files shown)


Dekkonot:
I had a bunch of thoughts typed up on this, but then I left them over the weekend and they are reduced to atoms. So instead I'll try to summarize.

I don't think this is the best way to handle this. I think we should make `name` optional on `Project`, then during the actual snapshotting process we can generate a name for it that will be overridden by the `name` field if it's filled.

This has a notable consequence due to `ServeSession::project_name` though. For that, we'll probably just want to verify that the top-level project (the one used during `ServeSession::new`) has a `name` and raise an error if it doesn't.

This is a bit more involved than the solution you have right now, so let me know if you want one of us to take over, but that's probably going to be the cleanest way to do this.

osyrisrblx:
That makes sense to me :+1:

> This is a bit more involved than the solution you have right now, so let me know if you want one of us to take over, but that's probably going to be the cleanest way to do this.

Please do! I don't have a ton of time to invest into this right now and lack of Rust experience doesn't help. 😆

Happy to help test ofc.

Dekkonot:
Alright! I've implemented it and opened #870. Going to close this in favor of that.

PR #865, opened 2024-02-16
Insert Model.NeedsPivotMigration in insert_instance when missing

### LINKED ISSUE #628: Rojo does not build model pivots correctly when no PrimaryPart is set
When attempting to build a project file that includes a model saved as a .rbxm with:

1) A pivot point offset from its center
2) No PrimaryPart set

Rojo will not set the pivot point of the model correctly in the resulting .rbxl

Below is a zip file containing a .rbxm which contains 2 identical models that demonstrate the issue. One has a PrimaryPart set and one does not. Inserting this file into studio directly, you'll see 2 identical models. Including it in a rojo build, you'll see 2 identical models with 2 different pivots.

[FloatingRock.zip](https://github.com/rojo-rbx/rojo/files/9353761/FloatingRock.zip)

kennethloeffler:
This PR closes #628 by adding a change to the 7.4.x branch that inserts `Model.NeedsPivotMigration` when it's missing from any instance that inherits from `Model`.

This is a pretty poor solution, but this bug needs to be fixed for 7.4.1 and the scope of the work in rojo-rbx/rbx-dom#385 (which when closed, will properly fix the underlying problem) is a little too large to comfortably fit into 7.4.1. 

Please make sure I got all the classes, and of course test the changes (I did confirm that they work myself, but it's good to have more sets of eyes)

### DIFF
--- a/rojo-test/build-test-snapshots/end_to_end__tests__build__rbxmx_in_folder.snap
+++ b/rojo-test/build-test-snapshots/end_to_end__tests__build__rbxmx_in_folder.snap
@@ -1,7 +1,6 @@
 ---
 source: tests/tests/build.rs
 expression: contents
-
 ---
 <roblox version="4">
   <Item class="Folder" referent="0">
@@ -25,6 +24,7 @@ expression: contents
           <R21>0</R21>
           <R22>1</R22>
         </CoordinateFrame>
+        <bool name="NeedsPivotMigration">false</bool>
         <Ref name="PrimaryPart">null</Ref>
         <BinaryString name="Tags"></BinaryString>
       </Properties>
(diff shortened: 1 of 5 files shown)


Dekkonot (approved):
Going to hesitantly approve this because we need to fix this. I don't like it though.

kennethloeffler:
> Actually upon further inspection I'm not actually able to tell if this has worked. The test model in #628 looks to be identical in both versions, but it's possible I am missing something.
> 
> What test were you using to check this?
> 
> EDIT: After using the test file you gave in #866, it seems to work. I wonder if the test file provided by the original issue was flawed in some way?

It does work for the first issue - the pivot in the first issue's repro is positioned at the bottom:
Screenshot 2024-02-19 at 2 25 05 PM

Dekkonot:
...So it is. Sometimes I wonder about myself. Alrighty, nevermind then!

Dekkonot (approved):

PR #870, opened 2024-02-20
Make the `name` field in projects optional

### LINKED ISSUE #858: Optional top-level default.project.json "name" field

Dekkonot:
Closes #858.

If a project is named `default.project.json`, it acts as an `init` file and gains the name of the folder it's inside of. If it is named something other than `default.project.json`, it gains the name of the file with `.project.json` trimmed off. So e.g. `foo.project.json` becomes `foo`.

### DIFF
--- a/src/serve_session.rs
+++ b/src/serve_session.rs
@@ -110,7 +110,7 @@ impl ServeSession {
 
         log::debug!("Loading project file from {}", project_path.display());
 
-        let root_project = match vfs.read(&project_path).with_not_found()? {
+        let mut root_project = match vfs.read(&project_path).with_not_found()? {
             Some(contents) => Project::load_from_slice(&contents, &project_path)?,
             None => {
                 return Err(ServeSessionError::NoProjectFound {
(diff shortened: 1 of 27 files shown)


Dekkonot:
Backport for 7.4.x is here: https://github.com/rojo-rbx/rojo/pull/871

src/serve_session.rs
```diff
@@ -118,6 +118,9 @@ impl ServeSession {
                 });
             }
         };
+        if root_project.name.is_none() {
+            return Err(ServeSessionError::NoProjectName);
+        }
```
kennethloeffler, line 123: Does this PR actually close #858? My understanding of the issue is that this should be allowed, and the project name in this case should match the name of the parent folder. I could be misunderstanding...

Dekkonot: That... is a concern actually. I was thinking that it was intended to be used in packages, so they'd never be the top-level project file but upon rereading the issue it's clear that I misunderstood.

Simple enough to fix though.

src/serve_session.rs
```diff
@@ -190,7 +219,10 @@ impl ServeSession {
     }
 
     pub fn project_name(&self) -> &str {
-        &self.root_project.name
+        self.root_project
+            .name
+            .as_ref()
+            .expect("all top-level projects must have their name set")
```
kennethloeffler, line 225: Is this panic reachable?

kennethloeffler (commented):
Looking good, just had a small question about `ServeSession::project_name` and then I will approbe

src/serve_session.rs
Dekkonot: It shouldn't be. I gave it a separate warning just in case though.

kennethloeffler: Cool, I'm approving this then and we'll have @osyrisrblx check out #871 before merging since that's the one that's shipping right now

kennethloeffler (approved):

PR #871, opened 2024-02-20
Backport #870 (optional project names) to 7.4.x

Dekkonot:
Unlike most of the other backports, this code couldn't be directly translated so it had to be re-implemented. Luckily, it is very simple. This implementation is a bit messy and heavy handed with potential panics, but I think it's probably fine since file names that aren't UTF-8 aren't really supported anyway. The original implementation is a lot cleaner though.

The test snapshots are (almost) all identical between the 7.5 implementation and this one. The sole exception is with the path in the `snapshot_middleware::project` test, since I didn't feel like adding a `name` parameter to `snapshot_project` in this implementation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Rojo Changelog
 
 ## Unreleased Changes
+* Made the `name` field optional on project files when they are not the top-level project ([#870])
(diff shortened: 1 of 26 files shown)


kennethloeffler:
Make sure to bring over the change from https://github.com/rojo-rbx/rojo/pull/870/commits/3591fd736694a216bf8cb193ab468e83499bae1a as well!

PR #882, opened 2024-02-29
Build tree ancestry correctly

boatbomber:
I really screwed this one up. Don't know how I managed that.

Tree was being built out of order, but only for some change types. This meant that some changes built a good tree, some built an upside down one. I suppose that's how it slipped past us.


Dekkonot:
Can you provide an example of a tree that built upside down so that we can verify it's been fixed?

boatbomber:
Project Structure:
image

From `master`:
image

From `fix/patch-tree-build-ancestry`:
image

Dekkonot (approved):
I was hoping that example would be like, a zip file but that works!

PR #884, opened 2024-02-29
Use GuiState instead of manual calculation for tooltips

boatbomber:
This ends up being more reliable and less wasteful. GuiState seems to update more reliably than MouseEnter/Leave.
The alternative way to make this more reliable would be to run our manual mouse bounds check every frame, which I'd like to avoid.


Dekkonot (approved):
I haven't tested this, so the joke is on me if this breaks tooltips, but I assume you would have noticed them not working at all so I am going to approve it since the code looks sound.

PR #886, opened 2024-03-02
Settings redesign

boatbomber:
Before:

image image

After:

image image

### DIFF
--- a/plugin/src/App/StatusPages/Settings/Setting.lua
+++ b/plugin/src/App/StatusPages/Settings/Setting.lua
@@ -13,6 +13,7 @@ local Theme = require(Plugin.App.Theme)
 local Checkbox = require(Plugin.App.Components.Checkbox)
 local Dropdown = require(Plugin.App.Components.Dropdown)
 local IconButton = require(Plugin.App.Components.IconButton)
+local SlicedImage = require(Plugin.App.Components.SlicedImage)
 
 local e = Roact.createElement
 
@@ -51,7 +52,7 @@ end
 
 function Setting:render()
 	return Theme.with(function(theme)
-		theme = theme.Settings
+		local settingsTheme = theme.Settings
 
 		return e("Frame", {
 			Size = self.contentSize:map(function(value)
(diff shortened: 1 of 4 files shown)


dphfox:
Those Experimental badges look hard to read in light theme. Perhaps make the background a highly translucent version of the accent colour, and have the text be the opaque version of the accent colour, like this:

boatbomber:
Oh yeah, that looks way nicer.
image
image

boatbomber:
image

How about this? Gives back a lot of vertical space, reduces annoying scrolling through settings.

Artoshia:
> image
> 
> How about this? Gives back a lot of vertical space, reduces annoying scrolling through settings.

Looks good, easy to read and well defined buttons are still large enough to click easy, but info is nice and dense where it counts.

plugin/src/App/StatusPages/Settings/Setting.lua
```diff
@@ -120,37 +121,55 @@ function Setting:render()
...
+							text = if self.props.unstable then "UNSTABLE" else "DEBUG",
+							icon = if self.props.unstable
+								then Assets.Images.Icons.Warning
+								else Assets.Images.Icons.Debug,
+							color = if self.props.unstable
+								then settingsTheme.Setting.UnstableColor
+								elseif self.props.debug then settingsTheme.Setting.DebugColor
+								else settingsTheme.BrandColor,
+						})
+						else nil,
```
Dekkonot, line 147: This section makes me kind of sad but since there's only two types of tags it's probably fine. If we ever end up adding another tag, I'm gonna propose we generalize it somehow though.

Dekkonot (commented):
The new icons need to be placed into the [`assets`](https://github.com/rojo-rbx/rojo/tree/master/assets) folder in this repo. We've neglected that for a while now but I want to make sure that's something we start doing again.

Otherwise, looks good to me.

plugin/src/App/StatusPages/Settings/Setting.lua
boatbomber: Nah you're right, this is lazy. I fixed it, thanks for calling it out.

boatbomber:
I took the opportunity to organize the assets folder a bit more and added all the icon images.

Dekkonot (approved):

PR #887, opened 2024-03-04
Update Cargo dependencies

Dekkonot:
So, earlier today I got a notification that we had a new security vulnerability open on Rojo. For the curious, it is https://github.com/advisories/GHSA-r8w9-5wcg-vfj7 and I don't think it actually impacts us.

What it made me realize however was that our dependencies were fairly out of date and the version listed in Cargo.toml for most of them was not the version we were actually using and it's been a long time since we just blanket updated things. So, I ran `cargo update` and then changed the version in our `Cargo.toml`s to reflect the version that's now in the `Cargo.lock` file.

This is not something I would recommend we do very often, but Tokio was listed as being `1.12.0` despite us actually using `1.32.0` before this change. We should try to at least be somewhat accurate with our dependencies.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,18 +19,18 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.1"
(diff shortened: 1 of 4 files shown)


kennethloeffler (approved):
LGTM

PR #889, opened 2024-03-12
Trim plugin version string

EgoMoose:
This PR is a very small change that fixes the string pattern that reads the rojo version from `Version.txt`. Currently this reads an extra new-line character which makes reading the version text in the plugin difficult.

It seems the rust side of things already trims the string when comparing, but the lua side does not.

Current:
pO6gtOXAZq

Fix:
RobloxStudioBeta_GHmiJKAoa3

Apologies if I skipped over some process of submitting a bug and / or am basing on the wrong branch etc.


kennethloeffler:
Can you open a PR with this commit against master, as well? We'll want to fix it there too

EgoMoose:
Here's the PR against master
https://github.com/rojo-rbx/rojo/pull/890

PR #896, opened 2024-03-21
Add rojo open

### LINKED ISSUE #305: Flag to open built place file in Roblox Studio

nezuo:
Closes #305.

- [ ] Support existing local places (partially-managed).
- [ ] Potentially support cloud places (except for automatically connecting)
- [ ] Potentially installing the rojo plugin locally

### DIFF
--- a/aftman.toml
+++ b/aftman.toml
@@ -1,5 +1,5 @@
 [tools]
-rojo = "rojo-rbx/rojo@7.3.0"
+rojo = "rojo-rbx/rojo@7.4.1"
 selene = "Kampfkarren/selene@0.26.1"
-stylua = "JohnnyMorganz/stylua@0.18.2"
+stylua = "JohnnyMorganz/stylua@0.20.0"
 run-in-roblox = "rojo-rbx/run-in-roblox@0.3.0"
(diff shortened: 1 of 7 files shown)


nezuo:
My current thoughts on the API:

`rojo open <PROJECT_FILE>`

Flags:
`--output (-o) <PATH>`: Builds the place from scratch and outputs at `<PATH>`.
`--place (-p) <PATH_OR_CLOUD_PLACE_ID>`: Opens an existing place, doesn't build anything from scratch and doesn't skip the initial sync.

`--output` and `--place` can't both be specified. 

I also want a flag to skip initial sync even with `--place`. I have projects that are fully managed but part of the build step uses Lune. This means I have existing place files that I can't build from scratch with Rojo but I still want to skip the initial sync. I'm worried that this flag might introduce confusion for users and I don't know what to call it.

Thoughts?

nezuo:
One flaw with injecting a `StringValue` is that if the user opens an existing place and doesn't save it, the next time they open it the `StringValue` will still be there.

Yumacide:
> Rojo saves previous hosts/ports you've connected to. When we generate a port, we don't want that to be saved.

It might be better not to randomize the port then. The behavior shouldn't be different from `rojo serve`.

iminlikewithyou:
is the idea of this command to be exactly like `serve`, except you also specify one of..?:
- an output to build to
- a place file to use
- a Roblox place ID

PR #901, opened 2024-04-09
Switch to checkout v4 and dtolnay/rust-toolchain in release workflow

kennethloeffler:


kennethloeffler:
wrong repository, sorry 🤪

PR #902, opened 2024-04-11
Release workflow maintenance

kennethloeffler:
This PR performs a number of relatively straightforward maintenance tasks and improvements for the "Release" workflow:

* Upgrades to actions/checkout@v4 and switches to dtolnay/rust-toolchain instead in favor of actions-rs/toolchain

    Reasoning is the same as #900 

* Switches to softprops/create-gh-release

    actions/create-release is unmaintained and has been archived for more than three years. 

    I considered using the GitHub CLI's (which is preinstalled on GitHub-hosted runners)`gh release create` to do this so we wouldn't have to pull in another third-party action, but it requires a checkout first - but doing this only adds a few more seconds before the release is created comapred to create-gh-release, so it may be acceptable?

* Switches to using preinstalled GitHub CLI to upload release archives

    There's absolutely no point in using a third party action for this, so we won't!

* Simplifies the and consolidates the way we generate names for release archives

Here's an example of the results of this workflow on my fork:
https://github.com/kennethloeffler/rojo/actions/runs/8642870695
https://github.com/kennethloeffler/rojo/releases/tag/v7.4.0-releasevvv-workflow-tewstingsbvv

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,7 +27,7 @@ jobs:
     name: Build Roblox Studio Plugin
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           submodules: true
 
@@ -89,7 +89,7 @@ jobs:
     env:
       BIN: rojo
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           submodules: true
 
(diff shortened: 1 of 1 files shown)


Dekkonot:
Does using GitHub's CLI for creating a release substantially increase the complexity? Glancing at the docs makes me think no, and it's probably worth using it over an action even if it adds a small delay.

Otherwise, I like the changes and thank you for taking care of this. Touching workflows is always work I want to do but never get around to.

.github/workflows/release.yml
```diff
@@ -8,26 +8,20 @@ jobs:
...
-        uses: actions/create-release@v1
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        with:
-          tag_name: ${{ github.ref }}
-          release_name: ${{ github.ref }}
-          draft: true
-          prerelease: false
+        run: |
+          gh release create ${{ github.ref_name }} --draft --verify-tag --title ${{ github.ref_name }}
```
kennethloeffler, line 17: The command is quite simple, so we may as well do it

Here are the results of this revision:
https://github.com/kennethloeffler/rojo/actions/runs/8657764221 (I had to re-run the macOS ARM job because I got rate limited, but the rest of the artifacts are visible in the previous run)
https://github.com/kennethloeffler/rojo/releases/tag/v7.4.0-release-workflow-testing-2

Dekkonot (approved):

PR #903, opened 2024-04-17
Fix removing trailing newlines by using str::replace in memofs

### LINKED ISSUE #899: Rojo unexpectedly trims trailing newline characters when syncing to Studio.
My code in Visual Studio Code:
image
Note that there are two lines in the document, with the second one being blank.

My code when synced to Roblox Studio:

image
Note that there is only one line in the document. The blank line is missing.

This makes doing a comparison between `script.source`s a problem in some cases such as mine. Some may argue there shouldn't be a trailing newline here, but more predictable behavior would be more ideal to me. It is up to the developer to follow good programming practices.

Workaround:
Add a second trailing empty line in Visual Studio Code so there will be at least one in Roblox Studio.

Let me know if I put this in the wrong place.

kennethloeffler:
This PR closes #899 by changing the implementation of `Vfs::read_to_string_lf_normalized` to use `str::replace` to replace `\r\n` with `\n`, rather than joining the result of `str::lines`. 

This is also a small performance win because less work needs to be done when the file only contains LF line endings, and we never need to collect into an intermediate `Vec<&str>`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Rojo Changelog
 
 ## Unreleased Changes
+* Fixed removing trailing newlines ([#903])
 * Added Never option to Confirmation ([#893])
 * Added popout diff visualizer for table properties like Attributes and Tags ([#834])
 * Updated Theme to use Studio colors ([#838])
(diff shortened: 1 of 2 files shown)


Dekkonot (approved):
Part of me wonders if we should add more tests just for competency's sake, but I imagine if `String::replace` fails, our other tests will fail quickly.

PR #908, opened 2024-05-04
Use msgpack for API

### LINKED ISSUE #363: No support for infinity

### LINKED ISSUE #881: warning "value cannot be converted to a number" on Billboard Gui's "MaxDistance" property
I have attached a [Repro File (.zip)](https://github.com/rojo-rbx/rojo/files/14442202/Repro.zip). It contains a billboard gui located in ReplicatedStorage

Repro Version: Rojo 7.4.1.
Repro Steps: 
1. run `rojo build`
2. run `rojo serve`
3. sync with Studio Plugin

```
value cannot be converted to a number  -  Edit
Stack Begin  -  Studio
Script 'user_RojoManagedPlugin.rbxm.Rojo.Packages.RbxDom.PropertyDescriptor', Line 19 - function set  -  Studio
Script 'user_RojoManagedPlugin.rbxm.Rojo.Packages.RbxDom.PropertyDescriptor', Line 77 - function write  -  Studio
Script 'user_RojoManagedPlugin.rbxm.Rojo.Plugin.Reconciler.setProperty', Line 31 - function setProperty  -  Studio
Script 'user_RojoManagedPlugin.rbxm.Rojo.Plugin.Reconciler.applyPatch', Line 204 - function applyPatch  -  Studio
Script 'user_RojoManagedPlugin.rbxm.Rojo.Plugin.Reconciler', Line 67 - function applyPatch  -  Studio
Script 'user_RojoManagedPlugin.rbxm.Rojo.Plugin.ServeSession', Line 285  -  Studio
Script 'user_RojoManagedPlugin.rbxm.Rojo.Packages.Promise', Line 172 - function runExecutor  -  Studio
Script 'user_RojoManagedPlugin.rbxm.Rojo.Packages.Promise', Line 181  -  Studio
Stack End  -  Studio
```

nezuo:
Fixes #363.
Fixes #881.

# Remaining Tasks
- [ ] Test more extensively.
- [ ] Use `msgpack` for `subscribe` api request.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -26,6 +26,11 @@ fn snapshot_from_fs_path(path: &Path) -> io::Result<VfsSnapshot> {
                 continue;
             }
 
+            // Ignore images in msgpack-luau because they aren't UTF-8 encoded.
+            if file_name.ends_with(".png") {
+                continue;
+            }
+
             let child_snapshot = snapshot_from_fs_path(&entry.path())?;
             children.push((file_name, child_snapshot));
         }
--- a/.gitmodules
+++ b/.gitmodules
@@ -16,3 +16,6 @@
 [submodule "plugin/Packages/Highlighter"]
 	path = plugin/Packages/Highlighter
 	url = https://github.com/boatbomber/highlighter.git
+[submodule "plugin/Packages/msgpack-luau"]
+	path = plugin/Packages/msgpack-luau
+	url = https://github.com/cipharius/msgpack-luau
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1719,9 +1719,9 @@ dependencies = [
 
 [[package]]
 name = "rmp"
-version = "0.8.12"
+version = "0.8.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f9860a6cc38ed1da53456442089b4dfa35e7cedaa326df63017af88385e6b20"
+checksum = "228ed7c16fa39782c3b3468e974aec2795e9089153cd08ee2e9aefb3613334c4"
 dependencies = [
  "byteorder",
  "num-traits",
@@ -1730,9 +1730,9 @@ dependencies = [
 
 [[package]]
 name = "rmp-serde"
-version = "1.1.2"
+version = "1.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bffea85eea980d8a74453e5d02a8d93028f3c34725de143085a844ebe953258a"
+checksum = "52e599a477cf9840e92f2cde9a7189e67b42c57532749bf90aea6ec10facd4db"
 dependencies = [
  "byteorder",
  "rmp",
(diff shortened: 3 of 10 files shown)


nezuo:
`patch_compute.rs` will treat `NaN` properties as always changed. This means that you will see `nan` property changes in the rojo plugin UI (when nothing actually changed).

I'm not sure what's the best way to go about fixing this. Ideally we also account for it types that have multiple numbers like `CFrame` or `NumberSequence`.

nezuo:
I just pushed a commit to account for `NaN` values. It's a lot of complexity just to handle `NaN` (and hopefully we don't forget to update it with any new variants because `Variant` is non_exhaustive). I'm not really sold on the approach, but I'd like to hear anyone's thoughts on it.

src/snapshot/patch_compute.rs
```diff
@@ -224,6 +227,128 @@ fn compute_children_patches(
     }
 }
 
+/// Trait where NaN values must not be treated as different.
+trait Different {
```
Dekkonot, line 231: This whole trait is... not ideal. I would strongly prefer we move the equality check for `Variant` to its own file, and make it a simple function instead of a trait.

I have a function already implemented in my syncback fork [here](https://github.com/UpliftGames/rojo/blob/syncback-internal-release/src/variant_eq.rs) if you want to steal it.

It does require us to pull in `float_cmp` as a crate, but I think that's fine.

nezuo:
I'm now using the `variant_eq` function from your fork. I did change the the `Attributes` comparison to simplify it if you want to check that out.

Dekkonot (changes requested):
This still has the added `Different` trait despite it not being used for anything anymore. It should be removed.

nezuo:
That's my bad, I thought I removed it.

build.rs
```diff
@@ -30,6 +30,11 @@ fn snapshot_from_fs_path(path: &Path) -> io::Result<VfsSnapshot> {
                 continue;
             }
 
+            // Ignore images in msgpack-luau because they aren't UTF-8 encoded.
+            if file_name.ends_with(".png") {
+                continue;
+            }
+
```
Dekkonot, line 37: This could probably stand to be made more robust rather than special-casing png files, but I'm also willing to push that off into a different PR since this is primarily plugin-facing.

Dekkonot (commented):
I see no obvious problems with this change, but I'm scared of deploying it since it's not battle tested. `msgpack-luau` is reasonably well tested and used by Argon so I trust it, but I still worry.

@kennethloeffler What do you think?

kennethloeffler:
Because we have urgent changes in the pipeline (namely, support for the new `Content` datatype), and already have several changes of large surface area in both Rojo and its backing libraries, this change is too risky to include in Rojo 7.5. We'll definitely want to get this in after we're done there though. 

The big thing we'll have to tackle here is more comprehensive testing. I'd looove to have automated tests for this - it's probably something we can feasibly accomplish with Lune!

nezuo:
Closing in favor of #1176.

PR #909, opened 2024-05-07
Add legacy and runContext script sync rule middlewares

nezuo:


nezuo:
I just saw that sync rules don't apply to `init.` scripts. It would be impossible to make a sync rule that turns scripts into their legacy counterparts with `emitLegacyScripts` off that works with `init.` scripts.

Dekkonot:
My thought was that we'd use the existing middleware for Legacy behavior and add middleware for the RunContext variants, since it'd allow changing a single script easier. I'm not married to that idea, but I think it makes more sense than the other way around.

What do you think?

Dekkonot:
> I just saw that sync rules don't apply to init. scripts. It would be impossible to make a sync rule that turns scripts into their legacy counterparts with emitLegacyScripts off that works with init. scripts.

This is something we need to address eventually. While implementing syncback, I swapped directories (you can see that [here](https://github.com/UpliftGames/rojo/blob/syncback-internal-release/src/snapshot_middleware/mod.rs#L183)) to be Middleware and it made everything easier. We can probably piggyback off of that when syncback gets merged to support customizing them somehow.

nezuo:
> My thought was that we'd use the existing middleware for Legacy behavior and add middleware for the RunContext variants, since it'd allow changing a single script easier. I'm not married to that idea, but I think it makes more sense than the other way around.
> 
> What do you think?

If we do this, there won't be a way to say "give me a script based on the `emitLegacyScripts` option." That said, I can't think of a case where that's useful.

I do think this would make defining rules more succinct. For example, with the current solution my PR has:
```json
// Gradually adopt Legacy scripts.
"emitLegacyScripts": false,
"syncRules": [
	{
		"pattern": "*.legacy.server.lua",
		"use": "legacyServerScript",
		"suffix": ".legacy.server.lua"
	},
	{
		"pattern": "*.legacy.client.lua",
		"use": "legacyClientScript",
		"suffix": ".legacy.client.lua"
	},
],
```
```json
// Gradually adopt RunContext scripts.
"emitLegacyScripts": false,
"syncRules": [
	{
		"pattern": "*.server.lua",
		"use": "legacyServerScript",
		"suffix": ".server.lua"
	},
	{
		"pattern": "*.client.lua",
		"use": "legacyClientScript",
		"suffix": ".client.lua"
	},
	{
		"pattern": "*.rc-server.lua",
		"use": "serverScript",
		"suffix": ".rc-server.lua"
	},
	{
		"pattern": "*.rc-client.lua",
		"use": "clientScript",
		"suffix": ".rc-client.lua"
	}
],
```

However, with your proposal these would become:
```json
// Gradually adopt Legacy scripts.
"emitLegacyScripts": false,
"syncRules": [
	{
		"pattern": "*.legacy.server.lua",
		"use": "serverScript", // This is identical but it's now serverScript instead of legacyServerScript.
		"suffix": ".legacy.server.lua"
	},
	{
		"pattern": "*.legacy.client.lua",
		"use": "clientScript",
		"suffix": ".legacy.client.lua"
	},
],
```
```json
// Gradually adopt RunContext scripts.
"emitLegacyScripts": true,
"syncRules": [
	{
		"pattern": "*.rc-server.lua",
		"use": "runContextServerScript",
		"suffix": ".rc-server.lua"
	},
	{
		"pattern": "*.rc-client.lua",
		"use": "runContextClientScript",
		"suffix": ".rc-client.lua"
	},
],
```

Gradually adopting RunContext scripts now only uses 2 rules instead of 4.

I'm on board with your proposal, do you have an opinion on what the RunContext `use` values should be called? `runContextServerScript`?

Dekkonot:
That's a mouthful, but I can't think of anything better at this moment. It might be fine to abbreviate RunContext to `rc` just for brevity, but I'm also mindful of the fact that `rcServerScript` is basically the same as `serverScript` if you aren't paying attention.

Dekkonot (commented):
As implemented right now, this makes `emitLegacyScripts` not function. I imagine this isn't deliberate?

With regards to the changelog, I don't know! The table I made was meant as a quick guide to syncing rules, but I think it's probably fine to have this as its own changelog entry (probably directly below the main sync rules one).

nezuo:
I've introduced `legacyServerScript` and `legacyClientScript` as a fix for `emitLegacyScripts`. Now `serverScript` and `clientScript` will respect the project's `emitLegacyScripts` option.

Dekkonot (approved):

Dekkonot (approved):

PR #910, opened 2024-05-07
Update workflow and tooling versions

Dekkonot:
- Update the `setup-aftman` action to `v0.4.2` instead of using `v0.1.0` or `v0.3.0`
- Start using Aftman v0.2.8
- Update tools versions in the Aftman.toml file
- Fix clippy lints in a few files
- Update version of `actions/checkout` to `v4` in our changelog check

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,10 +26,8 @@ jobs:
           submodules: true
 
       - name: Setup Aftman
-        uses: ok-nick/setup-aftman@v0.1.0
+        uses: ok-nick/setup-aftman@v0.4.2
         with:
-          token: ${{ secrets.GITHUB_TOKEN }}
-          trust-check: false
           version: 'v0.2.6'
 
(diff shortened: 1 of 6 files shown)


.github/workflows/release.yml
```diff
@@ -89,11 +87,9 @@ jobs:
           targets: ${{ matrix.target }}
 
       - name: Setup Aftman
-        uses: ok-nick/setup-aftman@v0.1.0
+        uses: ok-nick/setup-aftman@v0.4.2
         with:
-          token: ${{ secrets.GITHUB_TOKEN }}
-          trust-check: false
-          version: 'v0.2.6'
+          version: 'v0.2.8'
```
kennethloeffler, line 92: Should this be `v0.3.0`?

Dekkonot: Good catch!

kennethloeffler (approved):

PR #913, opened 2024-05-13
Add YAML middleware that behaves like TOML and JSON

Dekkonot:
After adding TOML in addition to JSON, it seems reasonable to also support YAML doing the same thing. However, I hate yaml, so I'm not sure whether we should have it be a default middleware.

I will leave it up to the committee (anyone who reads this) to weigh in on whether we should just by-default sync `.yml` files or if we should just expose a middleware for them.


boatbomber:
I think that YAML is common enough that it should be supported out of the box, personal tastes aside. I don't think anyone _likes_ YAML, but everyone uses it anyway.

src/snapshot_middleware/yaml.rs
```diff
@@ -0,0 +1,116 @@
...
+    let value: serde_yaml::Value = serde_yaml::from_slice(&contents)
+        .with_context(|| format!("File contains malformed YAML: {}", path.display()))?;
+
+    let as_lua = yaml_to_lua(value).to_string();
+
+    let properties = hashmap! {
+        "Source".to_owned() => as_lua.into(),
+    };
+
+    let meta_path = path.with_file_name(format!("{}.meta.json", name));
```
Kampfkarren, line 31: Nit

```suggestion
    let meta_path = path.with_file_name(format!("{name}.meta.json"));
```

src/snapshot_middleware/yaml.rs
```diff
@@ -0,0 +1,116 @@
...
+    name: &str,
+) -> anyhow::Result<Option<InstanceSnapshot>> {
+    let contents = vfs.read(path)?;
+
+    let value: serde_yaml::Value = serde_yaml::from_slice(&contents)
+        .with_context(|| format!("File contains malformed YAML: {}", path.display()))?;
+
+    let as_lua = yaml_to_lua(value).to_string();
+
+    let properties = hashmap! {
```
Kampfkarren, line 27: `hashmap!` doesn't really need to exist anymore in Rust 2021.

```rs
HashMap::from_iter([
    ("Source".to_owned(), as_lua.into()),
]),
```

...works, and has the benefit of not needing a macro

Kampfkarren (commented):
I agree that yaml might make sense to just have on by default since it's obvious what the intent is if you put a .yml file in the codebase. No real passionate opinion there but I sway on default

Also are the docs stored in this repo? Should we update them?

UpliftMacaw:
Docs would need to be updated here: https://github.com/rojo-rbx/rojo.space/blob/master/docs/sync-details.md

PR #915, opened 2024-05-14
Use history recording and don't do anything permanent

### LINKED ISSUE #914: Plugin cannot undo rbxm sync if the root instance has changed type
Reproduction steps:
* Save an instance to an rbxm in a syncable folder
* Overwrite the rbxm with a different instance type
* Ctrl-z

The synced instance is now gone, further undos will not bring any version of it back. The usage of undo implies that overwriting the rbxm was a mistake, making this a particularly destructive bug, it essentially makes it impossible to recover from the mistake unless you have a copy of the overwritten file on disc or a copy of the correct instances in studio.

boatbomber:
Adopts the latest ChangeHistoryService APIs.  Our Undo/Redo notification warnings (to prevent accidentally undoing a Rojo patch) still work fine with this new API.

Avoids the permanent Destroy action, uses Remove instead. This closes #914.

### DIFF
--- a/plugin/src/Reconciler/applyPatch.lua
+++ b/plugin/src/Reconciler/applyPatch.lua
@@ -173,11 +173,11 @@ local function applyPatch(instanceMap, patch)
 
 			-- Because the user might want to Undo this change, we cannot use Destroy
 			-- since that locks that parent and prevents ChangeHistoryService from
-			-- ever bringing it back. Instead, we use Remove.
+			-- ever bringing it back. Instead, we parent to nil.
 
 			-- TODO: Can this fail? Some kinds of instance may not appreciate
-			-- being Removed, like services.
-			instance:Remove()
+			-- being reparented, like services.
+			instance.Parent = nil
 
 			-- This completes your rebuilding a plane mid-flight safety
 			-- instruction. Please sit back, relax, and enjoy your flight.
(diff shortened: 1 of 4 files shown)


plugin/src/Reconciler/applyPatch.lua
```diff
@@ -164,10 +172,14 @@ local function applyPatch(instanceMap, patch)
...
+
+			-- Because the user might want to Undo this change, we cannot use Destroy
+			-- since that locks that parent and prevents ChangeHistoryService from
+			-- ever bringing it back. Instead, we use Remove.
+
 			-- TODO: Can this fail? Some kinds of instance may not appreciate
-			-- being destroyed, like services.
-			instance:Destroy()
+			-- being Removed, like services.
+			instance:Remove()
```
Dekkonot, line 182: I'd prefer we just parent it to `nil` over using Remove, if it's all the same.

Dekkonot (commented):
I have concerns about memory usage if we make this change, since connections will remain active on Instances now. I know there's not much we can do about it, but I'm still unhappy.

Otherwise, would prefer we not use Remove but looks good to me.

boatbomber:
> not much we can do about it, but I'm still unhappy

Sums up a lot of Rojo work, doesn't it 😭

kennethloeffler:
Does the problem in the linked issue only occur for changed class names? We destroy instances for normal removals, too

boatbomber:
Good catch, Ken. Updated all uses of `Destroy` to be `Parent = nil`, and updated all the relevant tests to pin this behavior.

PR #917, opened 2024-05-15
Correct issue with default.project.json files with no name being named `default` after change

### LINKED ISSUE #916: A `default.project.json` with an omitted `name` field sets the name of its top-level instance to `default` after subsequent changes
This issue is reproducible on latest mater, and is related to [#858 ](https://github.com/rojo-rbx/rojo/pull/870).

Repro steps:
1. Create a `default.project.json` and omit the `name` field. I used the following project: [no_name_top_level_project.zip](https://github.com/rojo-rbx/rojo/files/15312483/no_name_top_level_project.zip).
2. Run `rojo serve` on the project, and visit `http://localhost:34872/show-instances`. Observe that the resulting instance has the correct name: 
image
3. Make any change to the project file (I added a newline). Refresh `http://localhost:34872/show-instances` and observe that the instance's name has changed to `default`:
image

Dekkonot:
Closes #916.

Turns out we had a test that should've caught this, it's just that nobody caught it when this feature was implemented. We'll likely want to backport this into the 7.4.x branch as well.

Let me know if you have a better way to calculate project names in the middleware, I'm not in love with the way it's implemented here.

### DIFF
--- a/src/project.rs
+++ b/src/project.rs
@@ -177,22 +177,15 @@ impl Project {
         } else if let Some(fallback) = fallback {
             self.name = Some(fallback.to_string());
         } else {
-            for rule in self.sync_rules.iter().chain(default_sync_rules()) {
-                // I don't believe there's any way to get here without the
-                // matching middleware being a Project middleware.
-                if rule.matches(&self.file_location) {
-                    self.name = Some(
-                        rule.file_name_for_path(&self.file_location)
-                            .map_err(|_| Error::ProjectNameInvalid {
-                                path: self.file_location.clone(),
-                            })?
-                            .to_string(),
-                    )
-                }
-            }
-            return Err(Error::ProjectNameInvalid {
-                path: std::mem::take(&mut self.file_location),
-            });
+            // As of the time of writing (July 10, 2024) there is no way for
+            // this code path to be reachable. It can in theory be reached from
+            // both `load_fuzzy` and `load_exact` but in practice it's never
+            // invoked.
+            // If you're adding this codepath, make sure a test for it exists
+            // and that it handles sync rules appropriately.
+            todo!(
+                "set_file_name doesn't support loading project files that aren't default.project.json without a fallback provided"
+            );
         }
 
         Ok(())
--- a/src/project.rs
+++ b/src/project.rs
@@ -1,5 +1,6 @@
 use std::{
     collections::{BTreeMap, HashMap, HashSet},
+    ffi::OsStr,
     fs, io,
     net::IpAddr,
     path::{Path, PathBuf},
@@ -20,6 +21,14 @@ pub struct ProjectError(#[from] Error);
 
 #[derive(Debug, Error)]
 enum Error {
+    #[error("The folder for the provided project cannot be used as a project name: {}\n\
+            Consider setting the `name` field on this project.", .path.display())]
+    FolderNameInvalid { path: PathBuf },
+
+    #[error("The file name of the provided project cannot be used as a project name: {}.\n\
+            Consider setting the `name` field on this project.", .path.display())]
+    ProjectNameInvalid { path: PathBuf },
+
     #[error(transparent)]
     Io {
         #[from]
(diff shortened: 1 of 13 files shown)


src/snapshot_middleware/project.rs
```diff
@@ -23,7 +23,17 @@ pub fn snapshot_project(
...
+        Some(name) => name,
+        None => match path.file_name().and_then(OsStr::to_str) {
+            Some("default.project.json") => project
+                .folder_location()
+                .file_name()
+                .and_then(OsStr::to_str)
+                .unwrap_or(name),
+            _ => name,
+        },
+    };
```
kennethloeffler, line 36: ... and then, I replaced this with
```suggestion
    let project_name = match project.name.as_deref() {
        Some(name) => name,
        None => panic!("Project is missing a name"),
    };
```
Panicking is appropriate here, because if we ever load a project to snapshot and it's missing a name field, then it's a bug.

After these changes, optional name handling is entirely defined in the project module, and all tests pass. Are there any drawbacks you can see?

kennethloeffler (changes requested):
I think optional name handling should probably be consolidated into the project module. Having optional name logic done in two different places is harder to follow and gives too much room for mistakes, and I don't see any benefits in doing so. 

I transferred the following block into a function in the project module (which is then called in `Project::load_from_slice` and `Project::load_exact` [maybe we should remove one of these too? I'm not sure why we need both]), and also moved the associated error variants into `ProjectError`:
https://github.com/rojo-rbx/rojo/blob/3d4e387d350bd12366026623d1f0340a034bb1c4/src/serve_session.rs#L121-L147

kennethloeffler:
So there is a problem with the above approach: sync rules mean that project files won't necessarily have `.project.json` as a suffix, meaning that such files must contain a `name` field, otherwise the project will fail to load with `ProjectNameInvalid`. I think we should still allow a missing name field in this case. I'd still like the optional name logic to be defined in one place, so maybe we should give `Project::load_from_slice` a `name` parameter to fall back to in case we can't strip the `.project.json` suffix?

Dekkonot:
I went through and redid the name resolution and it's now generalized. As long as we use the right methods on `Project`, it'll always work the same now.

In the process, I did modify project loading to always use a Vfs, which impacts the fmt-project subcommand. Nothing invasive though.

src/snapshot_middleware/mod.rs
```diff
@@ -63,6 +63,9 @@ pub fn snapshot_from_vfs(
     if meta.is_dir() {
         if let Some(init_path) = get_init_path(vfs, path)? {
             // TODO: support user-defined init paths
+            // If and when we do, make sure to go support it in
+            // `Project::load_from_slice`, as right now it special-cases
```
kennethloeffler, line 67:
```suggestion
            // `Project::set_file_name`, as right now it special-cases
```

src/project.rs
```diff
@@ -135,43 +148,104 @@ impl Project {
...
+                            .map_err(|_| Error::ProjectNameInvalid {
+                                path: self.file_location.clone(),
+                            })?
+                            .to_string(),
+                    )
+                }
+            }
+            return Err(Error::ProjectNameInvalid {
+                path: std::mem::take(&mut self.file_location),
+            });
```
kennethloeffler, line 195: This code path is not being tested - I'm not sure if it's actually reachable? I think the only way we could possibly reach this right now is through the `load_exact` call in ServeSession, which is the only place where we pass `None` for the fallback name - but since serve session cannot load a project that doesn't end in `.project.json`, I don't think sync rules can apply?

kennethloeffler (commented):
Looking good, just a couple comments:

src/project.rs
Dekkonot: I don't think it's reachable right now, but it's something I want to handle now instead of just adding a `todo!` or something similar. That said, I can write a test for it if you'd like.

kennethloeffler: Yeah, try to find a way to exercise it in a test - but if it's impossible to test without changes to ServeSession's root project resolution, I prefer a `todo!` (also as it stands, there's no early return once we find a matching middleware, so this code path will always error)

I also question whether it will ever make sense to allow sync rules to override ServeSession's behavior here - maybe the fallback name should just not be optional?

Dekkonot:
...I 100% brainfarted and added `middleware` to that last commit message even though it is not related to the middleware. Don't worry about it.

kennethloeffler (approved):

PR #936, opened 2024-07-10
Add a `$paths` field for folders

Nolan-O:
Not sure if this is an acceptable feature but I wanted to see if it was possible 🙂

Allows a roblox folder's contents to be the sum of multiple flesystem folders' contents. Enhances the ability for rojo to change a project's form based on project files. E.G. multi-place games, libraries with tiered features for releases, a `tests.project.json` file

This feature may conflict with syncback since it introduces an ambiguity where new contents of a roblox folder should be stored. If possible, I think it would be acceptable to use the first path in the list of paths as a "dominant" one, treating it like the base folder on the filesystem.

Notes:
* $paths is mutually exclusive with $path
* $paths can only be on Folder nodes (maybe unnecessary?)
* $paths can only point to folders
* $paths ignores the metadata files in those folders (because none of them are the sole instance being synced)

### DIFF
--- a/rojo-test/serve-test-snapshots/end_to_end__tests__serve__multiple_paths_all.snap
+++ b/rojo-test/serve-test-snapshots/end_to_end__tests__serve__multiple_paths_all.snap
@@ -1,0 +1,41 @@
+---
+source: tests/tests/serve.rs
+expression: "read_response.intern_and_redact(&mut redactions, root_id)"
+
+---
+instances:
+  id-2:
+    Children:
+      - id-3
+      - id-4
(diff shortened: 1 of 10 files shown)


Dekkonot:
Howdy! I'm going to wait to give this a review until syncback is merged since rather selfishly I don't want to handle the merge conflicts on a PR that's already massive. I will get back to you though, promise.

Nolan-O:
Agreed tbh, makes sense for the large structural change to go first. This PR should have to work with syncback rather than the other way around.

Issue #929, opened 2024-06-22
Updates for Ref properties specified with attributes are always `null` in `/api/subscribe` responses


kennethloeffler:
To reproduce:
1. Run `rojo plugin install`
2. Run `rojo serve` on the [`ref_properties` test project](https://github.com/rojo-rbx/rojo/tree/7e2bab921aa71f76d07d0424e4bb4064e8b7c995/rojo-test/serve-tests/ref_properties)
3. Open a fresh baseplate in Roblox Studio, enable trace logging in the Rojo plugin, and click Connect
4. Make any change to ref_properties' `default.project.json` (I added a newline)
5. Observe the output, and notice that the subscribe response contains `null` for all the `Ref` properties specified with attributes in ref_properties' `default.project.json`:
```json
{
  "sessionId": "d264957a-76ce-486b-8df6-7151de8da0c1",
  "messageCursor": 2,
  "messages": [
    {
      "removed": [],
      "added": {},
      "updated": [
        {
          "id": "516ae8375fd349b4a61ac47529c969cf",
          "changedName": null,
          "changedClassName": null,
          "changedProperties": {
            "Value": null
          },
          "changedMetadata": null
        },
        {
          "id": "1aee8e1cc88e674f9917cc95e9d56bf6",
          "changedName": null,
          "changedClassName": null,
          "changedProperties": {
            "Value": null
          },
          "changedMetadata": null
        },
        {
          "id": "5e3e89750a0826e8d5d83b1ee86cad1a",
          "changedName": null,
          "changedClassName": null,
          "changedProperties": {
            "PrimaryPart": null
          },
          "changedMetadata": null
        },
        {
          "id": "04b43e42134d58c42408ff2b9ba1c9a4",
          "changedName": null,
          "changedClassName": null,
          "changedProperties": {
            "Value": null
          },
          "changedMetadata": null
        }
      ]
    }
  ]
}
``` 
I noticed this behavior in some test failures after #843 was merged: https://github.com/rojo-rbx/rojo/actions/runs/9605402645/job/26492956529#step:7:762. This makes it impossible for the Rojo plugin to live update a Ref property, could cause problems once rbx_dom_lua gains better ability to represent optional values, and can currently cause test failures on macOS due to seemingly inherent noisiness in FSEvents. 

One reason this could happen is because deferred Ref property application appears to not influence the returned result of patch application (i.e. `PatchApplyContext.applied_patch_set`), but rather only influences the tree:
[...]

PR #937, opened 2024-07-10
Implement Syncback to support converting Roblox files to a Rojo project

Dekkonot:
Well, the time has come. Syncback is real. Over 300 commits and 9 months later, here we are.

I'd like to apologize. There are a *lot* of files in this pull request and most of them are for tests. I went back and forth about whether they should be included in the this PR or added in a second one, but ultimately I decided that someone would have to go through them either way so I may as well include them now. If you want them removed so that the code is easier to review, I can do so.

This is a general tech overview of what's changed, what's added, and how it works. It is't a detailed tech dive because I'm a firm believer in reading code, but this should serve as a primer for reading the code.

## Overview

Syncback is a feature designed to solve the longstanding problem of converting Roblox files into Rojo projects. It works by accepting a Rojo project file and a Roblox file, then iterating through the file to match Instances to the Rojo project. Then, every Instance that's different has a middleware assigned to it that then has a method for converting the Instance back into the file type the middleware represents.

The middleware used is the old one (if an Instance is already present in the Rojo project) or a reasonable default depending upon the class of the Instance.

The path an Instance is written to is dependent upon where it is in project tree. So if you had a project like this:

```json
{
    "tree": {
        "DataModel": {
            "ReplicatedStorage": {
                "$path": "src/Shared"
            }
        }
    }
}
```

All children of `ReplicatedStorage` would be placed under `src/Shared`. If one or more of those children became a directory (such as a `ModuleScript` with children becoming a directory with `init.luau` in it), then descendants of that Instance would be placed under that directory. This is recursive for every path that exists within a project file.

A direct consequence of this is that the tree of a project file controls syncback almost directly.

## Change Summary

The following changes were made to Rojo to help support syncback. It does not cover syncback-specific additions, those will be later.

- Added `create_dir` and `create_dir_all` functions to `memofs`
- `InstigatingSource::ProjectNode` was changed to be a struct variant instead of a tuple variant
- A fair few things were made public exports of `librojo`
- Projects, JsonModels and Metadata now use `BTreeMap` instead of `HashMap` for properties and attributes
- CSV `example` field now has an alias named `examples`
- CSV now uses `Cow<str>` instead of just `&str`
- Directories (both ones that become Folders and ones with `init` files) now use `Middleware` as their backing
- The `relevant_paths` of directories now only contains actually existent paths
- `Middleware` is now stored in the metadata for `Instance`
- `Middleware` now displays in the web ui

Of all of these, I have an explanation for all of them except for changing what type of variant `InstigatingSource::ProjectNode` is. In my defense, that change as made 7 months ago and it probably made sense at the time.

You may have concerns with the `BTreeMap` change. I do too. I don't know what the performance hit of using a `BTreeMap` is at scale because I don't have any tests that uses projects, json models, or metadata files enough to stress test it. The testing I did seems to indicate it has a minimal impact because of how infrequently they're used though; at most it adds a bit of startup time for the server. A change of this nature is necessary in some capacity because the order of these maps has to be deterministic for syncback to function well.

I'm happy to answer any other questions you have on the changes.

## Syncback Additions

There are three main parts to syncback:

- The `SyncbackSnapshot` and the related `SyncbackReturn` structs
- The `syncback` method for `Middleware` items
- The core loop that iterates through the Roblox file provided

Throughout the code for syncback you will see that I've used the terms 'old' and 'new'. In all cases, the term 'old' refers to the Rojo project and the term 'new' refers to the Roblox file. So e.g. the 'new dom' refers to the DOM from the Roblox file. I can change this if you'd like, but it's pervasive throughout the code and I think it's cleaner than 'project' and 'roblox' or whatever.

A `SyncbackSnapshot` contains data for performing syncback on a specific Instance. It's utilized both by the syncback method for each `Middleware` as well as the core syncback loop. You'll want to read the code for exact details, but in summary these snapshots contain data that's global for syncback (things like a reference to a VFS) and data for syncing back a specific Instance from the Roblox file. This includes a path to that it should be written to and the Instance it corresponds to from the Rojo file.

A `SyncbackReturn` contains the actual results of running syncback on a `SyncbackSnapshot`. It's returned from the syncback methods for each `Middleware`, as you may expect. It includes a `FsSnapshot` that contains the serialized form of the Instance (including any meta.json files), a set of `SyncbackSnapshots` to run syncback over (this is used by directories and projects), and a list of paths that should be *removed* by Syncback. These are consumed by the core syncback loop to know what changes to make to the file system as well as to continue the syncback process.

The bulk of the work is done in the individual syncback methods for each Middleware. These are stored in the (now-misnamed) `snapshot_middleware` submodules along with the snapshot functions for each middleware. They accept a `SyncbackSnapshot` and return a `SyncbackReturn`. Using the information contained within the passed `SyncbackSnapshot`, each `Middleware` reserializes the provided Instance and if necessary returns a list of other Instances that need to have syncback ran on them. The vast majority of `Middleware` variants have a very simple syncback implementation. Directories and project files are the exception, so they're documented a bit more below.

The core loop is essentially a queue of `SyncbackSnapshot`s that's extended by pulling the returned snapshots out of each `SyncbackReturn`. The loop [does some comparisons](#instance-comparisons) to skip identical Instances, checks that a given snapshot is *allowed* to be written (it may be blocked by one of the syncback rules), and then determines a `Middleware` to use for each snapshot before running the syncback method for the `Middleware`.

The core loop assumes that the root of the project file and the root of the Roblox file are the same thing, so the loop is able to get started by constructing a `SyncbackSnapshot` that contains the roots.

This sounds simple because it is. By delegating most of the logic to the individual middleware implementations, the core design of syncback is relatively straightforward to follow.

## Instance Matching

Syncback for a directory or project file has to perform matching between the Rojo project and Roblox file rather than relying upon `SyncbackSnapshot` like other middleware. This is because they need to generate the snapshots for their children and project nodes respectively. For directories, this also allows children that are present on the Rojo project but not in the Roblox file to be removed from the file system, which is an important quality of life feature for syncback.

Instance matching is done by name, since the file system requires unique names unlike Roblox. This runs into some problems with generic Instance names (such as 'Tree' or 'Part') so it isn't perfect. Rojo's project format also doesn't support multiple models having the same name but both being in the same place on the file system. Unless we want to adopt a method of this nature (multiroot models, multiple paths for a folder, or whatever), matching by name is going to have to work.

This is a surprisingly complex problem but both `Dir` and `Project` middlewares have a fair few comments describing the implementation as a result. Thus, I will simply encourage you to read the implementation and then ask me questions when you have them. It will be easier that way.

## Instance Comparisons

Syncback's core loop compare Instances from the `SyncbackSnapshot` and skips if them and all of their descendants are identical. This is done for UX more than anything, because it makes for a bad experience if files are rewritten when they don't have to be. Syncback can't just naively skip if two Instances are the same (identical name, classname, and properties map) because it's is designed such that parents initiate syncback for their children.

To quickly tell if two Instances are identical, there's a variety of ways you could handle it. The easiest one is to just compare two Instances' names, class name, and properties. This works well, but it doesn't work for comparing *subtrees*. To do that, you'd potentially have to compare infinite Instances from each tree and match them by name and class and... Well, it gets out of hand. So, rather than doing that syncback needed a way to quickly and efficiently compare entire subtrees, regardless of the number of descendants they had.

This problem was solved by hashing and comparing the hashes of two different Instances. At a glance, this doesn't seem to solve the descendant comparison problem, but by making it so that each parent incorporates the hashes of its children, it makes it so that hashes are only identical *if and only if* their descendants are also identical. All it takes to accomplish this is by hashing a DOM depth-first and then making it so that a parent incorporates the hashes of its children during the hashing process. This has the consequence of making it so you have to frontload calculating hashes, but it's relatively quick so it isn't a big deal. Syncback precomputes hashes before the core loop even begins so that there's only an initial overhead.

The current hashing algorithm used is Blake3. There's no real reasoning behind this except that Blake3 is used by `rbx-types` already for hashing `SharedString`, so it doesn't add a direct dependency. Really the only requirement is that the hash be reasonably quick (not cryptographic), and to that end something like XXH3 or GxHash may be more appropriate, but that's mostly hypothetical. The hashing process is by no means a bottleneck for syncback.

An interesting note is the prospect of hashing floating point numbers. As most Roblox data types use floats, it's inevitable that we run into imprecisions. To try to mitigate this, floats are rounded somewhat (literally via: `((x * 10.0).round() / 10.0)`) before hashing, but that only solves minor imprecisions. I do not have a solution for larger ones. In practice, this has proven to not be a big deal for Adopt Me and other tests, but it's still worth noting as a potential problem.


Dekkonot:
I will be honest: I don't know what's going wrong with the MacOS tests. I've ran them extensively on Windows and they work just fine, but the fact that they fail on MacOS is a mystery to me. I'd love some input from someone on Mac with regards to fixing this.

Dekkonot:
Test failure is caused by an unrelated bug. Fix for it is in #917.

src/syncback/mod.rs
```diff
@@ -0,0 +1,495 @@
...
+    #[serde(default, skip_serializing_if = "HashMap::is_empty")]
+    ignore_properties: HashMap<String, Vec<String>>,
+    /// Whether or not the `CurrentCamera` of `Workspace` is included in the
+    /// syncback or not. Defaults to `false`.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    sync_current_camera: Option<bool>,
+    /// Whether or not to sync properties that cannot be modified via scripts.
+    /// Defaults to `false`.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    sync_unscriptable: Option<bool>,
```
kennethloeffler, line 353: Defaulting to false might not be a great idea. This means users have to care about whether the properties they want to write back are scriptable or not, and they won't find out about it until they want to write back a property that happens to be unscriptable. This will seem like a bug to them! 

I understand that the purpose is to prevent the Rojo plugin from displaying warnings when syncing a project generated by syncback, but this doesn't seem like a pressing issue because most syncback workflows will always start with a `rojo build`, so the Rojo plugin won't attempt to set the unscriptable properties.

src/syncback/property_filter.rs
```diff
@@ -0,0 +1,112 @@
...
+    if let Some(class_data) = class_data {
+        let defaults = &class_data.default_properties;
+        for (name, value) in &inst.properties {
+            if predicate(name, value) {
+                continue;
+            }
+            if let Some(default) = defaults.get(name.as_str()) {
+                if !variant_eq(value, default) {
+                    allocation.push((name, value));
+                }
```
kennethloeffler, line 66: If I set a property to a non-default value and sync it back, then set the property back to its default value and sync it back, the property will be filtered out and the final result will be the first, non-default value. I'm not sure at this point how to fix this

We should generally be very careful in this function, and ensure that we don't erroneously filter out any properties!

kennethloeffler (commented):
This is exciting! The overall approach is good, so we'll mostly be looking for bugs and places where UX can be improved. 

> I went back and forth about whether they should be included in the this PR or added in a second one

Including the test snapshots is fine, I'll be using git diff for the bulk of my review and I can limit to Rust files if I want to see it that way 

As you've mentioned, there are a couple unrelated bugs currently causing CI to fail:
* https://github.com/rojo-rbx/rojo/issues/929
* https://github.com/rojo-rbx/rojo/issues/916

Though they are not related to this feature, I think we should get these fixed up first thing so CI will be green and to get any possible merge conflicts out of the way.

We'll also need to remember to release any supporting changes to rbx-dom and update Rojo's dependencies accordingly:
* https://github.com/rojo-rbx/rbx-dom/pull/411
* Make `UniqueId` serialize (not required, but would remove a bit of cruft in this PR) 
* Perhaps others? Let me know if I missed anything!

Since this is a large PR, I'll chip away at the changes in small pieces over the next week. Some comments will concern the feature itself as presented, while others will be about further integrating the feature and reusing some of the new code within other parts of Rojo. For the latter sort of comment, I'll first mention them here, then once this PR is merged, spin them off into issues to be addressed in later PRs.

src/syncback/mod.rs
Dekkonot: You nailed why I had this default to false. The workflow I had in mind is new users who are doing one time conversions of their games to Rojo, but I think it might actually be bad behavior for them too. I'll change it to default to `true`; good call.

src/syncback/snapshot.rs
```diff
@@ -0,0 +1,226 @@
...
+    pub fn get_new_inst_path(&self, referent: Ref) -> String {
+        let mut path = Vec::new();
+        let mut path_capacity = 0;
+
+        let mut inst = self.get_new_instance(referent);
+        while let Some(instance) = inst {
+            path.push(&instance.name);
+            path_capacity += instance.name.len() + 1;
+            inst = self.get_new_instance(instance.parent());
+        }
```
kennethloeffler, line 139: ...because the root instance will have the same name of the project. In other words. the above example will only work if the project is named `DataModel`. This behavior is unintuitive.

src/syncback/snapshot.rs
```diff
@@ -0,0 +1,226 @@
...
+        let mut str = String::with_capacity(path_capacity);
+        while let Some(segment) = path.pop() {
+            str.push_str(segment);
+            str.push('/')
+        }
+        str.pop();
+
+        str
+    }
```
kennethloeffler, line 148: As an aside, we can simplify this by using some standard library methods:
 ```suggestion
     pub fn get_new_inst_path(&self, referent: Ref) -> String {
        let mut path = Vec::new();

        let mut inst = self.get_new_instance(referent);
        while let Some(instance) = inst {
            path.push(instance.name.as_str());
            inst = self.get_new_instance(instance.parent());
        }

        path.reverse();
        path.join("/")
    }
```

File: src/syncback/mod.rs [hunk 5]
```diff
@@ -0,0 +1,495 @@
...
+/// it will use `JsonModel`.
+///
+/// This will **not** override existing `Rbxm` middleware. It will only impact
+/// new files.
+const DEBUG_MODEL_FORMAT_VAR: &str = "ROJO_SYNCBACK_DEBUG";
+
+/// A glob that can be used to tell if a path contains a `.git` folder.
+static GIT_IGNORE_GLOB: OnceLock<Glob> = OnceLock::new();
+
+pub fn syncback_loop(
```
kennethloeffler, line 50: One idea is to prepend the root instance name to ignoreTrees entries when we start the syncback loop.

The ignoreTrees example becomes 
```json
{ "ignoreTrees": [ "ServerStorage/ImportantSecrets" ] }
```
This would make the behavior of ignoreTrees entries more similar to ignorePaths, in that they implicitly start at the root.

kennethloeffler (commented):
Some comments about ignoreTrees (I also couldn't find any tests for ignoreTrees, we should add one):

src/syncback/mod.rs
Dekkonot: `ignoreTrees` is honestly something I implemented ages ago and haven't touched all that much recently. I agree this is more intuitive behavior and it's something I wanted to fix.

For context, originally the root of the Roblox file wasn't renamed to match the project's name, so it was generally always `DataModel`. This was... not ideal, but it at least made them work consistently. Now we have no such luck. This is a good catch though, since it slipped my mind!

kennethloeffler: Since this is different from how it works on the Uplift Games fork, do you think it'd be a good idea to document it in the changelog (and also do the same for any other deviations we may make)?

Dekkonot: It's probably worth noting if/when we make an RC for syncback on `master`, but for the final release I don't think it's necessary. The changelog will need to be updated either way, so I think this is fine.

src/cli/syncback.rs
```diff
@@ -0,0 +1,321 @@
...
+            }
+            log::trace!("Children of new root:");
+            for child in dom_new.root().children() {
+                let inst = dom_new.get_by_ref(*child).unwrap();
+                log::trace!("{} (class: {})", inst.name, inst.class);
+            }
+        }
+
+        let start = Instant::now();
+        log::info!("Beginning syncback...");
```
kennethloeffler, line 96: Consider removing or demoting to log::debug

src/cli/syncback.rs
```diff
@@ -0,0 +1,321 @@
...
+        let snapshot = syncback_loop(
+            session_old.vfs(),
+            &mut dom_old,
+            dom_new,
+            session_old.root_project(),
+        )?;
+        log::info!(
+            "Syncback finished in {:.02}s!",
+            start.elapsed().as_secs_f32()
+        );
```
kennethloeffler, line 106: Consider removing or demoting to log::debug

src/cli/syncback.rs
```diff
@@ -0,0 +1,321 @@
...
+                io::stdout().flush()?;
+                let mut line = String::with_capacity(1);
+                io::stdin().read_line(&mut line)?;
+                line = line.trim().to_lowercase();
+                if line != "y" {
+                    println!("Aborting due to user input!");
+                    return Ok(());
+                }
+            }
+            log::info!("Writing to the file system...");
```
kennethloeffler, line 130: Consider removing or demoting to log::debug

src/syncback/fs_snapshot.rs
```diff
@@ -0,0 +1,171 @@
...
+            lock.remove_file(base_path.join(path))?;
+        }
+        drop(lock);
+
+        if self.added_dirs.len() + self.added_files.len() > 0 {
+            log::info!(
+                "Wrote {} directories and {} files to the file system!",
+                self.added_dirs.len(),
+                self.added_files.len()
+            );
```
kennethloeffler, line 107: Consider demoting to log::debug

src/syncback/fs_snapshot.rs
```diff
@@ -0,0 +1,171 @@
...
+        }
+        if self.removed_dirs.len() + self.removed_files.len() > 0 {
+            log::info!(
+                "Removed {} directories and {} files from the file system. Yikes!",
+                self.removed_dirs.len(),
+                self.removed_files.len()
+            );
```
kennethloeffler, line 114: Consider demoting to log::debug and removing "Yikes!"

kennethloeffler (commented):
Running the syncback command is pretty noisy - I'm not sure that we need some of these logs period, and many of them could be use log::debug instead of log::info:

src/syncback/mod.rs
```diff
@@ -0,0 +1,495 @@
...
+            ));
+
+            let new_snapshot = snapshot.with_new_path(path, snapshot.new, snapshot.old);
+            log::warn!(
+                "Could not syncback {inst_path} as a Directory, it will \
+                instead be synced back as a {new_middleware:?}."
+            );
+            syncback_res = new_middleware.syncback(&new_snapshot);
+        }
+        let syncback = syncback_res.with_context(|| format!("Failed to syncback {inst_path}"))?;
```
kennethloeffler, line 225: This code doesn't tell users why an instance was not able to be synced back to a directory. We might also consider using pattern matching here rather than a mutable variable, to make the code a little clearer:
```suggestion
        let syncback = match middleware.syncback(&snapshot) {
            Ok(syncback) => syncback,
            Err(err) if middleware == Middleware::Dir => {
                let new_middleware = match env::var(DEBUG_MODEL_FORMAT_VAR) {
                    Ok(value) if value == "1" => Middleware::Rbxmx,
                    Ok(value) if value == "2" => Middleware::JsonModel,
                    _ => Middleware::Rbxm,
                };

                let file_name = snapshot
                    .path
                    .file_name()
                    .and_then(|s| s.to_str())
                    .context("Directory middleware should have a name in its path")?;
                let mut path = snapshot.path.clone();
                path.set_file_name(format!(
                    "{file_name}.{}",
                    extension_for_middleware(new_middleware)
                ));

                let new_snapshot = snapshot.with_new_path(path, snapshot.new, snapshot.old);
                log::warn!(
                    "Could not syncback {inst_path} as a Directory because {err}
It will instead be synced back as a {new_middleware:?}."
                );
                new_middleware
                    .syncback(&new_snapshot)
                    .with_context(|| "Failed to syncback {inst_path}")?
            }
            Err(err) => anyhow::bail!("Failed to syncback {inst_path} because {err}"),
        };
```

(I'm also noticing that attempting to sync back e.g. a folder containing children with duplicate names causes this log to be printed twice - any idea why?)

kennethloeffler (commented):
If syncback fails to sync back to a directory it should tell the user why:

src/syncback/mod.rs
Dekkonot: I'm not sure why the duplication is happening, but I'll take a looksie.

src/cli/syncback.rs
```diff
@@ -0,0 +1,320 @@
...
+    serve_session::ServeSession,
+    syncback::{syncback_loop, FsSnapshot},
+};
+
+use super::{resolve_path, GlobalOptions};
+
+const UNKNOWN_INPUT_KIND_ERR: &str = "Could not detect what kind of file was inputted. \
+                                       Expected input file to end in .rbxl, .rbxlx, .rbxm, or .rbxmx.";
+
+/// Performs syncback for a project file
```
nezuo, line 26: This is a bit of a circular definition. It would be good to explain what a "syncback" even if it's a short description.

nezuo (commented):
### Project file formatting
Syncback essentially forces you to use rojo's json formatting. I would be fine with this if the formatting was better. This can be done in a separate PR (I'll try to make an issue with my specific issues with the formatting) but I do think the default formatting needs to be improved.

Edit: Made an issue: #962. I noticed that `syncback` doesn't actually format the project file the same was `fmt-project` does. I definitely think the should match!

### Directories are shown as changed when a child changes
image
I modified the `src/server/init.server.luau` file. It says that it will also write to `src/server` which is confusing and somewhat redundant since the `src/server` path is in the first change already.

### `default.project.json` is marked as change every syncback even though it doesn't change at all
Repro file:
[rojo-syncback-testing-bug.zip](https://github.com/user-attachments/files/16646078/rojo-syncback-testing-bug.zip)

### Syncback adds lots of properties after first save
image
image
I'm not really sure what could or should be done about this but I thought I should bring it up at least.
I noticed after I built the default rojo project file and synced it back, Lighting was added with some default properties. Could we avoid syncing back properties that are default?

src/variant_eq.rs
```diff
@@ -0,0 +1,208 @@
...
+                && a.style == b.style
+                && a.family == b.family
+                && a.cached_face_id == b.cached_face_id
+        }
+        (Variant::Int32(a), Variant::Int32(b)) => a == b,
+        (Variant::Int64(a), Variant::Int64(b)) => a == b,
+        (Variant::MaterialColors(a), Variant::MaterialColors(b)) => a.encode() == b.encode(),
+        (Variant::NumberRange(a), Variant::NumberRange(b)) => {
+            approx_eq!(f32, a.max, b.max) && approx_eq!(f32, a.min, a.max)
+        }
```
kennethloeffler, line 102:
```suggestion
        (Variant::NumberRange(a), Variant::NumberRange(b)) => {
            approx_eq!(f32, a.max, b.max) && approx_eq!(f32, a.min, b.min)
        }
```

kennethloeffler (commented):
RE avoiding syncing back default properties: the implementation here already attempts to do so, but it won't work for properties that don't have defaults in the reflection database, which can happen when the reflection database is out of date. This is what happened for most of the Workspace properties in the screenshot, except for StreamingEnabled and WorldPivotData, which are not at their default values. We should update the reflection database on master, and maybe the two mentioned properties should have their defaults patched?

I'm also not seeing any unexpected Lighting properties after saving the rbxl of rojo init's "place" template in Roblox Studio, then syncing back - did you mean Workspace and/or StarterPlayer?

It looks like all the GameSettings properties appear because of a small mistake in variant_eq:

nezuo:
> I'm also not seeing any unexpected Lighting properties after saving the rbxl of rojo init's "place" template in Roblox Studio, then syncing back - did you mean Workspace and/or StarterPlayer?

That's my mistake on the Lighting properties, I thought they weren't already apart of the default project file.

src/variant_eq.rs
Dekkonot: Good catch! Fixed this and cleaned up the variant_eq function a bit more since it'd been 9 months since I changed it.

src/cli/syncback.rs
Dekkonot: Good callout. I've written some more helpful docs for the command.

src/syncback/property_filter.rs
kennethloeffler: So to fix this, meta, JSON model, and project syncback middlewares need to *remove* properties if they exist in the old property map and if they have default values. 

I'm not sure what is the best way to go about this, but one idea I have for a first step is to make the property filter fn aware of both the old and new instance, and compare old and new values when filtering default properties (i.e. only filter a default property if both old and new values are at the default, or when there is no old instance).

Alternatively, the middlewares themselves could be responsible for filtering properties, instead of the property filter fn. Is it important for hashing that default properties are filtered out?

kennethloeffler:
It looks like the change in 47c702b7990c4f541f74da49c1e262f98d5c699c slightly borked the warning message for when a property is unrepresentable in json:
```
[WARN  librojo::snapshot_middleware::project] Rojo cannot serialize the property Workspace.ModelMeshData in project files.
        If this is not acceptable, resave the Instance at '' manually as an RBXM or RBXMX.
```

Previously, this warning would say:
```
[WARN  librojo::snapshot_middleware::project] Rojo cannot serialize the property Workspace.ModelMeshData in project files.
        If this is not acceptable, resave the Instance at 'ken-test-project' manually as an RBXM or RBXMX.
```

Dekkonot:
I will write a better test harness for syncback that ignores new properties so that we don't have to change them every time a new database update happens. For now though, I'm just updating them manually after database updates.

Dekkonot:
Issue with the warning should be fixed. I've also prevented `RBX_` attributes from serializing in JSON files because they're internal facing and thus we don't like them.

src/syncback/property_filter.rs
Dekkonot: If memory serves, the reason the default properties are filtered out for hashing is to guard against hashes being different if a model file is updated (i.e. manually resaved from Studio) but not changed otherwise. The idea is that we only care about properties that aren't their default values when determining if an Instance is different.

That said, I think it's probably fair to have JSON middlewares handle this themselves, but I'm not sure what the best way to approach it is. I'm worried about things getting more spaghetti than they are, so I'm hesitant to add more functions lol

src/snapshot_middleware/json_model.rs
```diff
@@ -58,10 +64,87 @@ pub fn snapshot_json_model(
...
+
+    let mut properties = BTreeMap::new();
+    let mut attributes = BTreeMap::new();
+    for (name, value) in prop_buffer.drain(..) {
+        match value {
+            Variant::Attributes(attrs) => {
+                for (attr_name, attr_value) in attrs.iter() {
+                    // We (probably) don't want to preserve internal attributes,
+                    // only user defined ones.
+                    if attr_name.starts_with("RBX_") {
```
kennethloeffler, line 105:
```suggestion
                    if attr_name.starts_with("RBX") {
```

src/snapshot_middleware/meta_file.rs
```diff
@@ -42,6 +49,80 @@ impl AdjacentMetadata {
...
+            .unwrap_or_default();
+
+        let class = &snapshot.new_inst().class;
+        for (name, value) in snapshot.get_path_filtered_properties(snapshot.new).unwrap() {
+            match value {
+                Variant::Attributes(attrs) => {
+                    for (attr_name, attr_value) in attrs.iter() {
+                        // We (probably) don't want to preserve internal
+                        // attributes, only user defined ones.
+                        if attr_name.starts_with("RBX_") {
```
kennethloeffler, line 90:
```suggestion
                        if attr_name.starts_with("RBX") {
```

src/snapshot_middleware/meta_file.rs
```diff
@@ -136,6 +232,84 @@ impl DirectoryMetadata {
...
+            .unwrap_or_default();
+
+        let class = &snapshot.new_inst().class;
+        for (name, value) in snapshot.get_path_filtered_properties(snapshot.new).unwrap() {
+            match value {
+                Variant::Attributes(attrs) => {
+                    for (name, value) in attrs.iter() {
+                        // We (probably) don't want to preserve internal
+                        // attributes, only user defined ones.
+                        if name.starts_with("RBX_") {
```
kennethloeffler, line 276:
```suggestion
                        if name.starts_with("RBX") {
```

src/snapshot_middleware/project.rs
```diff
@@ -304,6 +315,324 @@ pub fn snapshot_project_node(
...
+) {
+    let properties = &mut node.properties;
+    let mut attributes = BTreeMap::new();
+    for (name, value) in filtered_properties {
+        match value {
+            Variant::Attributes(attrs) => {
+                for (attr_name, attr_value) in attrs.iter() {
+                    // We (probably) don't want to preserve internal attributes,
+                    // only user defined ones.
+                    if attr_name.starts_with("RBX_") {
```
kennethloeffler, line 542:
```suggestion
                    if attr_name.starts_with("RBX") {
```

kennethloeffler (commented):
Tiny problem: the protected attribute name prefix is actually `RBX` (no trailing underscore), which you can verify by running something like 
```luau
game:SetAttribute("RBXVerySpecialAttribute", true)
```
in Roblox Studio's command bar.

Since we're not sure what sort of process (if any) Roblox uses when rolling out protected attributes, we have no reason to believe that the underscore will be present, so we should check for an `RBX` prefix.

src/syncback/property_filter.rs
kennethloeffler: I've put together a tentative fix for this at https://github.com/UpliftGames/rojo/pull/6

src/cli/syncback.rs
```diff
@@ -0,0 +1,292 @@
...
+            for path in removed {
+                writeln!(
+                    &mut buffer,
+                    "{}",
+                    display_absolute(path.strip_prefix(base_path).unwrap_or(path))
+                )?;
+            }
+        }
+        buffer.set_color(&no_color)?;
+    }
```
kennethloeffler, line 289: A user running syncback in interactive mode will already know that zero files would be touched from earlier messages, and programmatic usage is simpler if stdout stays empty when there are zero writes.

We can skip checking whether the entire snapshot is empty and just check additions and removals. Along with the output format changes, that'd be:
```suggestion
    let added = snapshot.added_paths();
    if !added.is_empty() {
        buffer.set_color(&add_color)?;
        for path in added {
            writeln!(
                &mut buffer,
                "Writing {}",
                display_absolute(path.strip_prefix(base_path).unwrap_or(path))
            )?;
        }
    }
    let removed = snapshot.removed_paths();
    if !removed.is_empty() {
        buffer.set_color(&remove_color)?;
        for path in removed {
            writeln!(
                &mut buffer,
                "Removing {}",
                display_absolute(path.strip_prefix(base_path).unwrap_or(path))
            )?;
        }
    }
    buffer.set_color(&no_color)?;
```

PR #944, opened 2024-07-19
Actually include attribute-defined properties in patch computation

### LINKED ISSUE #929: Updates for Ref properties specified with attributes are always `null` in `/api/subscribe` responses

Dekkonot:
Closes #929.

It turns out the underlying issue was that we weren't accounting for ref properties at all in the patch computation. This resolves that issue by just collecting them during `compute_property_patch`.

Of note, I've decided to just reuse the `ref_properties` test files for this since it's ultimately academic either way. It adds noise to the snapshot files but I think it's fine because the only thing that really matters here is the `messageCursor` on the return from the read endpoint. Let me know if you'd prefer a cleaner snapshot.

### DIFF
--- a/tests/tests/serve.rs
+++ b/tests/tests/serve.rs
@@ -477,3 +477,36 @@ fn ref_properties_remove() {
         );
     });
 }
+
+#[test]
+fn ref_properties_patch_update() {
+    // Reusing ref_properties is fun and easy.
+    run_serve_test("ref_properties", |session, mut redactions| {
+        let info = session.get_api_rojo().unwrap();
+        let root_id = info.root_instance_id;
+
+        assert_yaml_snapshot!(
+            "ref_properties_patch_update_info",
+            redactions.redacted_yaml(info)
+        );
+
+        let read_response = session.get_api_read(root_id).unwrap();
+        assert_yaml_snapshot!(
+            "ref_properties_patch_update_all",
+            read_response.intern_and_redact(&mut redactions, root_id)
+        );
+
+        let project_path = session.path().join("default.project.json");
+        let mut project_content = fs::read(&project_path).unwrap();
+        // Adding a newline to force the change processor to pick up a change.
+        project_content.push(b'\n');
+
+        fs::write(project_path, project_content).unwrap();
+
+        let read_response = session.get_api_read(root_id).unwrap();
+        assert_yaml_snapshot!(
+            "ref_properties_patch_update_all-2",
+            read_response.intern_and_redact(&mut redactions, root_id)
+        );
+    });
+}
--- a/tests/tests/serve.rs
+++ b/tests/tests/serve.rs
@@ -478,6 +478,12 @@ fn ref_properties_remove() {
     });
 }
 
+/// When Ref properties were first implemented, a mistake was made that resulted
+/// in Ref properties defined via attributes not being included in patch
+/// computation, which resulted in subsequent patches setting those properties
+/// to `nil`.
+///
+/// See: https://github.com/rojo-rbx/rojo/issues/929
 #[test]
 fn ref_properties_patch_update() {
     // Reusing ref_properties is fun and easy.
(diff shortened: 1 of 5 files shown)


Dekkonot:
And of course, like clockwork, tests start failing on MacOS.

I promise they all pass on Windows. :-)

kennethloeffler:
Code looks good, and tests pass for me on macOS, at least most of the time...

These test failures have been occurring because macOS' FSEvents API occasionally reports events that occurred *before* the serve session starts - i.e. in here https://github.com/rojo-rbx/rojo/blob/3ca975d81dfd6f06e024dec5a87bedf5f796dfc2/tests/rojo_test/serve_util.rs#L77-L98 
we're sometimes getting events from copying the test files to temporary directories. The only public information I can find about this behavior is this Stack Overflow question: https://stackoverflow.com/questions/47679298/howto-avoid-receiving-old-events-in-fseventstream-callback-fsevents-framework-o. It seems to happen more often when the system is under higher load, but I haven't been able to make any hard correlations.

I haven't been able to find any way we can avoid this. At this point, I'm thinking we should just sleep in between copying files and starting the server, hope it's enough time for FSEvents to not report these events, and be done with it. I've done this here: #945

Dekkonot:
Good news: that seems to have done it. Hopefully we can find a solution that isn't that at some point, but for now I think MacOS tests taking slightly longer isn't that bad.

tests/tests/serve.rs
```diff
@@ -477,3 +477,36 @@ fn ref_properties_remove() {
         );
     });
 }
+
+#[test]
+fn ref_properties_patch_update() {
```
kennethloeffler, line 482: Maybe add some comments to this test to describe what we expect to happen here, and why this test was added? Maybe link to the issue

Dekkonot:
Comment: added.

kennethloeffler (approved):

PR #945, opened 2024-07-20
Sleep between file copies and serve for macOS serve tests

kennethloeffler:
This PR adds a 100ms sleep between copying test files and starting the Rojo server in macOS serve tests. I added a comment explaining the addition - I'm not sure why it happens, there's very scant information about this issue on the internet, and this isn't quite a proper way to fix it. Ideally, we could somehow get FSEvents to flush its event stream before starting to process events, but that will probably require changes to notify.


Dekkonot (approved):
I'm not in love with this tactics anymore than you but until we find a workaround at least it'll get CI working

PR #948, opened 2024-07-23
Update rbx_dom for 7.4.x branch

Dekkonot:
Standard rbx_dom update.

I honestly just straight up forgot to fix the setter/getter for `WorldPivotData`; it should be an alias in the database, ideally. It'll be fixed in rbx-dom's repo, as I've opened an issue for it here: https://github.com/rojo-rbx/rbx-dom/issues/430

Since it'll require a change to the database, I'm gonna leave it be for now though. No harm no foul.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1586,9 +1586,9 @@ dependencies = [
 
 [[package]]
 name = "rbx_binary"
-version = "0.7.4"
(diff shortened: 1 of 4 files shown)


Dekkonot:
going to merge this without approval because anarchy rules

PR #952, opened 2024-07-29
Use `.luau` instead of `.lua`

CompeyDev:
This PR renames all the existing Lua files to Luau files, since GitHub now supports Luau syntax highlighting, and Lua typically isn't written on Roblox. 

Let me know if I missed any files. The conversion was primarily a:

```sh
find . -type f -name "*.lua" -exec perl-rename 's/\.lua$/.luau/' '{}' \;
```


Barocena:
`.gitattributes` file with `*.lua linguist-language=Luau` would be easier

CompeyDev:
> .gitattributes file with *.lua linguist-language=Luau would be easier

This isn't the proper solution, but rather a hack. It misleads people into believing this is a lua project.

Dekkonot:
Thanks for opening this PR, but I'm not going to accept it. The reason is that it's not a priority and it has the potential to cause conflicts with other PRs. To say nothing of the impact it would have on the git history for the plugin.

The plugin is in need of TLC, so likely what will happen is we'll transition over time instead of all at once. Either way, we won't be doing this.

@Barocena Feel free to open a PR setting the linguists language to Luau. I'd be open to that.

PR #954, opened 2024-07-30
Sync lock toggle

NcxyZero:
In the company where I work, we sometimes have trouble with the sync lock. We have to ask someone every 5 seconds to unsync just to test something. Everything we write is stored on GitHub, so even if something is overridden, we can just resync and get our files back. It can be really problematic when someone is on break, and we have to wait or kick them from Team Create just to do something. Sticking to an older version of Rojo is not a viable long-term solution, so I implemented a Sync Lock toggle. This allows you to toggle it on and off (it is on by default).

### DIFF
--- a/plugin/src/App/Notifications.lua
+++ b/plugin/src/App/Notifications.lua
@@ -113,12 +113,9 @@ function Notification:render()
 		end
 
 		local paddingY, logoSize = 20, 32
-		local actionsY = if self.props.actions then 35 else 0
+		local actionsY = if self.props.actions then 37 else 0
 		local textXSpace = math.max(250, buttonsX) + 35
-		local textBounds = Vector2.new(
-			textXSpace,
-			getTextBoundsAsync(self.props.text, theme.Font.Main, theme.TextSize.Body, textXSpace).Y
-		)
+		local textBounds = getTextBoundsAsync(self.props.text, theme.Font.Main, theme.TextSize.Body, textXSpace)
 		local contentX = math.max(textBounds.X, buttonsX)
 
 		local size = self.binding:map(function(value)
@@ -162,10 +159,10 @@ function Notification:render()
 						TextColor3 = theme.Notification.InfoColor,
 						TextTransparency = transparency,
 						TextXAlignment = Enum.TextXAlignment.Left,
-						TextYAlignment = Enum.TextYAlignment.Top,
+						TextYAlignment = Enum.TextYAlignment.Center,
 						TextWrapped = true,
 
-						Size = UDim2.new(1, -35, 1, -actionsY),
+						Size = UDim2.new(0, textBounds.X, 1, -actionsY),
 						Position = UDim2.fromOffset(35, 0),
 
 						LayoutOrder = 1,


Dekkonot:
Howdy.

Please revert the changes to the version included in the PR and put the changelog entry under the Unreleased Changes header instead.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Rojo Changelog
 
+## [7.4.3] - 31 July, 2024
```
boatbomber, line 3: You should put this in the `## Unreleased Changes` section, you are not making a new release.

Cargo.lock
```diff
@@ -1752,7 +1752,7 @@ dependencies = [
 
 [[package]]
 name = "rojo"
-version = "7.4.0"
+version = "7.4.3"
```
boatbomber, line 1755: Undo version change.

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "rojo"
-version = "7.4.0"
+version = "7.4.3"
```
boatbomber, line 3: Undo version change.

plugin/Version.txt
```diff
@@ -1 +1 @@
-7.4.0
\ No newline at end of file
+7.4.3
```
boatbomber, line 1: Undo version change.

plugin/src/App/StatusPages/Settings/init.lua
```diff
@@ -104,6 +104,14 @@ function SettingsPage:render()
 					layoutOrder = layoutIncrement(),
 				}),
 
+				SyncLock = e(Setting, {
+					id = "syncLock",
+					name = "Sync Lock",
+					description = "Toggle sync lock",
```
boatbomber, line 110: This description is not clear at all, and doesn't explain any more than the title already says.
```suggestion
					description = "Enables Team Create sync locking to prevent conflicts and overwrites between multiple users.",
```

boatbomber, line 110: Add `tag = "caution"` and add that tag type to the `TAG_TYPES` [here.](https://github.com/rojo-rbx/rojo/blob/master/plugin/src/App/StatusPages/Settings/Setting.lua#L21-L32)

plugin/src/App/init.lua
```diff
@@ -299,7 +299,7 @@ function App:getHostAndPort()
 	return if #host > 0 then host else Config.defaultHost, if #port > 0 then port else Config.defaultPort
 end
 
-function App:isSyncLockAvailable()
+function App:isSyncLockAvailable()
```
boatbomber, line 302: Remove whitespace. Make sure to always format your changes with StyLua.

plugin/src/App/init.lua
```diff
@@ -408,21 +408,24 @@ function App:useRunningConnectionInfo()
 end
 
 function App:startSession()
-	local claimedLock, priorOwner = self:claimSyncLock()
-	if not claimedLock then
-		local msg = string.format("Could not sync because user '%s' is already syncing", tostring(priorOwner))
+	local syncLockEnabled = Settings:get("syncLock")
```
boatbomber, line 411: This is the wrong way to go about this.

Consider the case of two users, Zack and Micah. Zack has locking enabled, Micah has it disabled. Zack is currently syncing and claimed the lock.

With your implementation, Micah can steal the lock from Zack but Zack cannot steal it back. Zack's settings made it clear that he did not want anyone else to sync over him. Micah doesn't mind if people sync over him, yet Zack still cannot because his setting is checking the lock.

The correct way to implement this setting would be that users with locking disabled simply do not claim the lock themselves, thus allowing other users to claim it. If everyone on your team disables it, then the lock is always unclaimed and anyone can sync whenever.
The setting is saying "I don't care about locking", but that should not override users who do care about locking. The lock must always be respected.

boatbomber (changes requested):

plugin/src/App/init.lua
boatbomber: Also, what do we do in the case where someone without locking is syncing, and then someone with locking starts syncing? The new user doesn't want other people syncing. Should you be kicked?

Your feature needs to consider all the cases, not just your use case of the entire team setting it the same way.

PR #964, opened 2024-08-22
Release Rojo v7.4.4

Dekkonot:
I cannot believe the team behind Lighting has forced this onto us twice in a row, but here we are. This releases Rojo 7.4.4.

I don't want to keep doing this. If we reach 7.4.5, I will cry.


kennethloeffler:
We've also marked UniqueId properties as serializable - should we also mention this on the changelog?

kennethloeffler:
We need to bump the plugin version too!

Dekkonot:
I'm ambivalent towards mentioning the UniqueId change because to most people it'll be the equivalent to a new property being added (it'll suddenly appear in their files and then basically never change). It's technically a change but I don't think it's strictly speaking user-facing.

kennethloeffler (approved):

PR #973, opened 2024-09-27
Use proper output paratemeter name in README.md template for plugins

mxxnsk:
For plugin projects, the -p option is no longer recognized, updating it to --plugin


mxxnsk:
Nvm, I got confused with aftman configs pointing to different versions of rojo

PR #977, opened 2024-10-18
Correct mistake in build command docs

Dekkonot:


kennethloeffler (approved):
This has apparently said "serve" for several years, and you're the first one to say anything about it 😄

PR #982, opened 2024-10-29
Remove maplit dependency and stop using a macro for hashmaps

Dekkonot:
I'm tired of this macro. It will cause us trouble down the line when we try to swap to a faster hashmap, and it's just not idiomatic.

So, I have removed it. In practice this mostly impacts tests.


Dekkonot:
Okay, there's a chance that this will just make adopting a new Hasher annoying. We shall see.

PR #988, opened 2024-11-06
Use FontFace and consistent text sizing

boatbomber:
The `Gotham` font we use no longer exists. This PR upgrades to using `FontFace`, and also unifies our styling by putting fonts and text sizes into `Theme` instead of hardcoding them inline. This will make it easier to adjust our UI in the future (or perhaps add accessibility settings).

It's pretty much identical to before, but all the text sizes are consistent (no more 14 vs 15) and the notifs and tooltips are better at getting the correct text bounds (no cutting off edge cases anymore).

Screenshot 2024-11-05 180127


plugin/src/App/getTextBounds.lua
```diff
@@ -0,0 +1,26 @@
+local TextService = game:GetService("TextService")
+
+local Rojo = script:FindFirstAncestor("Rojo")
+local Packages = Rojo.Packages
+
+local Log = require(Packages.Log)
+
+local params = Instance.new("GetTextBoundsParams")
+
+local function getTextBounds(text: string, font: Font, textSize: number, width: number, richText: boolean?): Vector2
```
kennethloeffler, line 10: This function can yield, so it should probably have a name that reflects this - and it's also not OK to yield in Roact! I'm not sure exactly *when* it will yield, and our `TextService:GetTextBoundsAsync` usage has seemed fine so far, but it probably warrants further review. You're more familiar with this area of the codebase so you'll be more able to tell if it's a problem or not

plugin/src/App/getTextBounds.lua
```diff
@@ -0,0 +1,26 @@
...
+	params.Text = text
```
kennethloeffler, line 11: This property set should probably have a check for text with length >= 200000, right now it's possible for it to brick the entire Roblox UI

kennethloeffler (commented):
Nice, and this can probably be included in the changelog as "Fixed notification and tooltip text sometimes getting cut off"

plugin/src/App/getTextBounds.lua
Dekkonot: `GetTextBoundsAsync` only yields when the font isn't loaded, as per [Roblox's own docs](https://create.roblox.com/docs/reference/engine/classes/TextService#GetTextBoundsAsync). Assuming we trust them, we could ensure it never yields by preloading the fonts when the plugin initializes.

That said, I suspect that builtin fonts are just always loaded in Studio or load quick enough that it won't matter. This would be more of a concern if we were using a custom font.

Dekkonot (commented):
Ken already addressed the concerns I have; this needs a changelog entry, and our text bound check might yield. Otherwise, looks good.

Dekkonot (approved):

plugin/src/App/getTextBoundsAsync.lua
```diff
@@ -0,0 +1,37 @@
...
+local params = Instance.new("GetTextBoundsParams")
+
+local function getTextBoundsAsync(
+	text: string,
+	font: Font,
+	textSize: number,
+	width: number,
+	richText: boolean?
+): Vector2
+	if type(text) ~= "string" or #text > 200_000 then
```
kennethloeffler, line 17: `#text > 200_000` needs to be either `#text >= 200_000` or `#text > 199_999` here, since the limit is 199,999. Sorry, I could have been more clear in my initial comment here!

Also, it might not be a great idea to send a massive string through `Log.warn` - in cases where the text is too long, we might just want to say "text too long" without printing the actual string

kennethloeffler (changes requested):

kennethloeffler (approved):

PR #1006, opened 2025-01-13
NaN deserialization patch - causes Rojo sync failure

nightcycle:
Patches edge-case where the Rojo Sync will fail when the underlying rbx binary has a NaN for the NumberSequenceKeypoint value.


Dekkonot:
This needs to be fixed upstream in [rbx-dom](https://github.com/rojo-rbx/rbx-dom), as we pull rbx-dom-lua from there. See: https://github.com/rojo-rbx/rbx-dom/blob/master/rbx_dom_lua/src/EncodedValue.lua

PR #1020, opened 2025-03-24
Remove reference to Moonscript

UpliftMacaw:


Dekkonot:
Not willing to commit to _not_ doing this in the future yet! I appreciate the sentiment, but there's still conceivably a future where we support passing files through custom middleware, which could in theory include Moonwave. Going to close this for now.

Issue #961, opened 2024-08-15
Add a place ID blacklist to project config


michaeldougal:
Currently, Rojo has a nice configuration option called `servePlaceIds` that allows us to whitelist a set of places where a project can sync. I would also love the opposite (or a toggle) so that we can have a blacklist that prevents Rojo from syncing to specific place IDs.

This is because my team typically works with a production branch place and a set of development branch places, and I'd like to make sure that the production is restricted. At the same time, I want to leave them open to syncing in any of the other development places without having to add each one to the config manually.

Dekkonot:
Are you adding new development places often enough that this is a real concern? I'd be interested in learning more about your workflow that it's a burden to add each one to the config file.

I'm not opposed to this suggestion on merit or anything, I'm just curious why you're adding so many places.

michaeldougal:
Yeah so to give you a little more context, I am the Technical Director at Sawhorse Interactive. We do a lot of brand work, and we're pretty much pumping out a new experience every 2-3 months.

The teams for those projects typically consist of 2-4 engineers, plus our art team. Each engineer has their own development branch place or two, the art team has two dedicated art branches (an asset drop and level making), there's a staging branch, and then the production branch. Overall this results in at least seven development branches per project, often more for random things that come up.

We have a policy of not directly doing any work in the production branch, everything should just get pushed to staging, verified, and then staging overwrites production. However, occasionally you'll get a deviation from that because a dev gets lazy, and I'd like to enforce not working in production to mitigate that behavior and teach better habits.

So it's not an extreme problem to have to add each place ID to the project configs, we just make these new places often enough that it would be nice to be able to just blacklist the production branch once and be good to go.

Dekkonot:
Right, that makes sense. I can see how it'd get tedious if you're doing a bunch of projects. It's not something I think about doing that often since most teams are working on a fixed few games. I will definitely keep this in mind!

I make no promises on a timeline though. It's a busy part of the year to be a Roblox dev and Rojo has some larger code changes coming as-is, so it might not get done for a little bit.

PR #1021, opened 2025-03-24
Add place ID blacklist config

Parritz:
Addresses #961 

Along with `servePlaceIds`, this adds an `avoidPlaceIds` option to the project config that allows development teams to block certain place ids from being synced to, such as a production place.

### DIFF
--- a/plugin/src/ApiContext.lua
+++ b/plugin/src/ApiContext.lua
@@ -43,16 +43,22 @@ local function rejectWrongProtocolVersion(infoResponseBody)
 	return Promise.resolve(infoResponseBody)
 end
 
+local function findPlaceId(placeIds)
+	local foundId = false
+
+    for _, id in ipairs(placeIds) do
+        if id == game.PlaceId then
+            foundId = true
+            break
+        end
+    end
+
+    return foundId
+end
+
 local function rejectWrongPlaceId(infoResponseBody)
 	if infoResponseBody.expectedPlaceIds ~= nil then
-		local foundId = false
-
-		for _, id in ipairs(infoResponseBody.expectedPlaceIds) do
-			if id == game.PlaceId then
-				foundId = true
-				break
-			end
-		end
+		local foundId = findPlaceId(infoResponseBody.expectedPlaceIds)
 
 		if not foundId then
 			local idList = {}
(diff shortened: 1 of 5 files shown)


plugin/src/ApiContext.lua
```diff
@@ -43,16 +43,22 @@ local function rejectWrongProtocolVersion(infoResponseBody)
...
+
+    for _, id in ipairs(placeIds) do
+        if id == game.PlaceId then
+            foundId = true
+            break
+        end
+    end
+
+    return foundId
+end
```
Dekkonot, line 57: This entire function is essentially just `table.find(placeIds, game.PlaceId)`. I don't think we need it.

plugin/src/ApiContext.lua
```diff
@@ -71,6 +77,26 @@ local function rejectWrongPlaceId(infoResponseBody)
...
+			for _, id in ipairs(infoResponseBody.unexpectedPlaceIds) do
+				table.insert(idList, "- " .. tostring(id))
+			end
+
+			local message = (
+				"Found a Rojo server, but its project is set to not be used with a specific list of places."
+				.. "\nYour place ID is %s, but needs to not be one of these:"
+				.. "\n%s"
+				.. "\n\nTo change this list, edit 'avoidPlaceIds' in your .project.json file."
+			):format(tostring(game.PlaceId), table.concat(idList, "\n"))
```
Dekkonot, line 94: You can use string interpolation for this rather than having to format the string. I'd rather we did that than the classic `..` for newlines, but this is just a nitpick; if you don't want to swap to string interpolation, you can swap the first `%s` for `%u` and then just pass `game.PlaceId` rather than casting it to a string.

src/project.rs
```diff
@@ -78,6 +78,14 @@ pub struct Project {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub serve_place_ids: Option<HashSet<u64>>,
 
+    /// If specified, contains a set of place IDs that this project is
+    /// not compatible with when doing live sync.
+    ///
+    /// This setting is intended to help prevent syncing a Rojo project into the
+    /// wrong Roblox place.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub avoid_place_ids: Option<HashSet<u64>>,
```
Dekkonot, line 87: How do you feel about this being named `blocked_place_ids` instead of `avoid_place_ids`? I feel it communicates the idea more clearly.

plugin/src/ApiContext.lua
Parritz: As far as I know, string interpolation in Luau doesn't have multiline strings, so it would all have to be in one line in backticks and separating lines by \n's. If that's not wanted, it could still be used to remove the need for :format but still use .. for new lines. I'm not sure which would look better or not, so I'll change it to %u for now.

src/project.rs
Parritz: Yeah, this is better. I just wasn't sure what to name it at first.

plugin/src/ApiContext.lua
Parritz: Went ahead and removed this function and changed it to table.find.

Dekkonot (commented):
This needs test snapshots updated, but otherwise looks good.

Dekkonot (approved):

PR #1023, opened 2025-03-28
Update rbx-dom

Dekkonot:
This turns out to be relatively painless. Lots of replacing Cows and HashMaps but otherwise very straightforward.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Rojo Changelog
 
 ## Unreleased Changes
+* Added support for Roblox's `Content` type. This replaces the old `Content` type with `ContentId` to reflect Roblox's change.
+ 	If you were previously using the fully-qualified syntax for `Content` you will need to switch it to `ContentId`.
 * Added support for `Enum` attributes
 * Significantly improved performance of `.rbxm` parsing
 * Support for a `$schema` field in all special JSON files (`.project.json`, `.model.json`, and `.meta.json`) ([#974])
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Rojo Changelog
 
 ## Unreleased Changes
+* Added support for `Enum` attributes
+* Significantly improved performance of `.rbxm` parsing
 * Support for a `$schema` field in all special JSON files (`.project.json`, `.model.json`, and `.meta.json`) ([#974])
 * Projects may now manually link `Ref` properties together using `Attributes`. ([#843])
  	This has two parts: using `id` or `$id` in JSON files or a `Rojo_Target` attribute, an Instance
--- a/src/resolution.rs
+++ b/src/resolution.rs
@@ -2,8 +2,8 @@ use std::borrow::Borrow;
 
 use anyhow::{bail, format_err};
 use rbx_dom_weak::types::{
-    Attributes, CFrame, Color3, Content, Enum, Font, MaterialColors, Matrix3, Tags, Variant,
-    VariantType, Vector2, Vector3,
+    Attributes, CFrame, Color3, Content, ContentId, Enum, Font, MaterialColors, Matrix3, Tags,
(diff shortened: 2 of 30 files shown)


Dekkonot:
I added the `Skip Changelog` label out of reflex. Writing an entry right now.

Dekkonot:
Feels weird but I think the only user-facing changes we need to report are `Enum` attributes and the performance improvements? Any other suggestions are welcome, but I don't think we e.g. need to tell everyone about `SharedString` having serde or `WorldPivot` being correctly defaulted now.

This does remind me that we need to _not_ forget about `NeedsPivotMigration`. That isn't set properly on master, only in the 7.4.x branch and it causes a lot of problems if it isn't set.

src/snapshot/instance_snapshot.rs
```diff
@@ -57,23 +57,23 @@ impl InstanceSnapshot {
         }
     }
 
-    pub fn class_name(self, class_name: impl Into<String>) -> Self {
+    pub fn class_name<'a, S: Into<&'a str>>(self, class_name: S) -> Self {
         Self {
-            class_name: Cow::Owned(class_name.into()),
+            class_name: ustr(class_name.into()),
```
kennethloeffler, line 62: For better readability, we might prefer using `Into<Ustr>` as the trait bound:
```rust
    pub fn class_name<S: Into<Ustr>>(self, class_name: S) -> Self {
        Self {
            class_name: class_name.into(),
            ..self
        }
    }
```

src/snapshot/instance_snapshot.rs
```diff
@@ -57,23 +57,23 @@ impl InstanceSnapshot {
...
+    pub fn property<'a, K, V>(mut self, key: K, value: V) -> Self
     where
-        K: Into<String>,
+        K: Into<&'a str>,
         V: Into<Variant>,
     {
-        self.properties.insert(key.into(), value.into());
+        self.properties.insert(ustr(key.into()), value.into());
         self
     }
```
kennethloeffler, line 74: We might prefer using `Into<Ustr>` as the trait bound:
```rust
    pub fn property<K, V>(mut self, key: K, value: V) -> Self
    where
        K: Into<Ustr>,
        V: Into<Variant>,
    {
        self.properties.insert(key.into(), value.into());
        self
    }
```

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Rojo Changelog
 
 ## Unreleased Changes
```
kennethloeffler, line 3: We may also want to mention `Content` properties, and their possible impact on existing Rojo projects.

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
...
+* Added support for `Enum` attributes
+* Significantly improved performance of `.rbxm` parsing
```
kennethloeffler, line 5: It's likely that several common Rojo workloads beyond rbxm parsing see performance improvements with these changes. Maybe we can add a couple more benchmarks and report some rough figures for each scenario here?

Dekkonot: You're not incorrect, but I'm not sure what benchmarks are actually available. Not many projects that I have access to are big enough to benefit without also being heavily reliant upon `rbxm` parsing.

src/resolution.rs
```diff
@@ -116,6 +116,9 @@ impl AmbiguousValue {
                 (VariantType::Content, AmbiguousValue::String(value)) => {
                     Ok(Content::from(value).into())
                 }
```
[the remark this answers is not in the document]
Dekkonot: Yeah it'll probably have to be with `RojoRef`. It's just as well, because we need to look into it for normal Ref properties either way.

CHANGELOG.md
kennethloeffler: Yeah, my idea was to be more user-centric about the nature of the performance differences, but we don't really have the time to create new benchmarks for this

kennethloeffler (approved):

PR #1027, opened 2025-04-08
Re-add the hack to write NeedsPivotMigration as `false` for models

Dekkonot:
This exists on 7.4.x but never made it to master by my own instruction. We have not fixed it upstream yet, so we need to make sure it's in subsequent releases of Rojo.

Skipping the changelog on this because it's been in the last 4 releases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Rojo Changelog
 
 ## Unreleased Changes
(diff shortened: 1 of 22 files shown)


Dekkonot:
Tried on a different system. Not failing locally either. I have no idea what's going on.

Dekkonot:
...Bad news, it looks like this is a windows-only gremlin. I'll dig into it some more.

kennethloeffler (approved):

PR #1030, opened 2025-04-13
Use SerializationService as a fallback for when patch application fails

### LINKED ISSUE #1028: Use SerializationService to bypass unscriptable properties limits
Supersedes #205.

In release 668, Roblox added [two new APIs](https://robloxapi.github.io/ref/updates/2025.html#version-04a222f011414c81-17) for handling the serde of Instances:
Image

We can use these to bypass unscriptable properties by having the plugin request serialized files and deserializing them in-engine. This is equivalent to the `GetObjects` solution that #205 proposed, but it does not rely upon the file system.

Naively, we could replace the entire patching process with this, but I think we shouldn't. It'll mangle anything not managed by Rojo that references the Instances, and it makes us face potential memory leaks when dealing with ChangeHistoryService (you cannot Destroy Instances if you want them to be restored by a waypoint).

Instead, we should just use this when a patch fails to apply and continue to try to prioritize getters/setters when capable to avoid having to entirely reify an Instance.

It may be worth doing this every time for the first sync though! It's bound to be more efficient.

Dekkonot:
Closes #1028. I don't anticipate this going into 7.5 obviously, but soon?

## In CLI
- Adds a new endpoint: GET `/api/model/{ids}` that accepts a list of IDs to serialize as a binary model and then return
- The model is sent as JSON object that Roblox deserializes as a `buffer`. This is because `SerializationService:DeserializeAsync` accepts a buffer as an argument so it saves a step. I've also preemptively compressed the buffer that's sent but I'm not actually sure it's necessary.

## In Plugin
- Adds a `__applyPatch` method to `ServeSession`
- Moves `Reconciler` post- and pre- commit hooks to ServeSession
- Moves ChangeHistoryService for patch application to `ServeSession:__applyPatch`

These changes were made to avoid the reconciler having to have an ApiContext. This was based on the feedback given in the GetObjects version of this PR by @kennethloeffler.

## To Do

- [ ] Ref properties are not reset. Probably the Reconciler can handle this, since it knows what ref properties should be what.
- [ ] Studio's Selection isn't preserved if you have an Instance that's replaced selected
- [ ] the `/api/model/{ids}` serializer doesn't take into account Instances with special parent requirements

Attempting to replace Instances is an all or nothing scenario right now. Since all models that failed to be added or updated attempt to be updated at once, if one of them fails to serialize or deserialize, it causes the entire process to fail. This could be changed, but I think it's fine. Open to feedback on this.


plugin/src/ServeSession.lua
```diff
@@ -207,6 +247,162 @@ function ServeSession:__onActiveScriptChanged(activeScript)
 	self.__apiContext:open(scriptId)
 end
 
+function ServeSession:__replaceInstances(idList)
+	if #idList == 0 then
+		return true, PatchSet.newEmpty()
+	end
+	-- It would be annoying if selection went away, so we try to preserve it.
```
boatbomber, line 254: 🐐

boatbomber (approved):
Assuming SerializationService works as expected, the plugin changes look good to me.

plugin/src/ServeSession.lua
boatbomber: (Spoiler: we don't) https://github.com/boatbomber/rojo/blob/master/plugin/src/Reconciler/applyPatch.lua#L118

plugin/src/ServeSession.lua
```diff
@@ -207,6 +247,162 @@ function ServeSession:__onActiveScriptChanged(activeScript)
...
+	end
+
+	-- TODO: Should we do this in multiple requests so we can more granularly mark failures?
+	local modelSuccess, replacements = self.__apiContext
+		:serialize(idList)
+		:andThen(function(response)
+			Log.debug("Deserializing results from model endpoint")
+			local objects = SerializationService:DeserializeInstancesAsync(response.modelContents)
+			local instanceMap = {}
+			for _, item in objects[1]:GetChildren() do
```
boatbomber, line 268: Tiny concern here: `objects[1]` is never checked to be non nil and I don't trust that it won't somehow be empty

Dekkonot: I guess I've never really considered what would happen if you tried to deserialize an empty model using `SerializationService` but this is a valid concern. Should probably add a check to make sure that all of the requested objects actually got delivered in case something silently fails.

plugin/src/ServeSession.lua
```diff
@@ -206,6 +246,169 @@ function ServeSession:__onActiveScriptChanged(activeScript)
...
+		selectionMap[instance] = i
+	end
+
+	-- TODO: Should we do this in multiple requests so we can more granularly mark failures?
+	local modelSuccess, replacements = self.__apiContext
+		:serialize(idList)
+		:andThen(function(response)
+			Log.debug("Deserializing results from serialize endpoint")
+			local objects = SerializationService:DeserializeInstancesAsync(response.modelContents)
+			if objects[1] then
```
boatbomber, line 266: Surely this should be `if not objects[1] then`, right?

Dekkonot:
> I'm not confident that the serialize endpoint will never respond with an rbxm that the engine fails to deserialize, and I think it's important that this fallback fails gracefully.

I agree. We should do that. I'll make an issue for it.

PR #1034, opened 2025-04-25
Change release build linux version to ubuntu-latest

kennethloeffler:
May have consequences - it's possible that the binary will be ran on systems with an out-of-date glibc


Dekkonot (approved):
I don't know that anyone is running Rojo in an environment without the capabilities to update glibc but they can come out of the woodwork if they exist.

Issue #839, opened 2024-01-27
Reduce light theme intensity on served website


AridAjd:
The title is pretty self-explanatory, but the request is so that the background of the website isn't so intense that hurts the eyes.
![image](https://github.com/rojo-rbx/rojo/assets/118916772/f51afd9b-b468-43e0-9652-44d57099ea48)

Thank you.

PR #1051, opened 2025-05-18
Reduce Serve site background intensity

Helloburp:
Response to (#839)

Not the most exciting change in the world, but it seemed like a good candidate for my first ever PR :)

I nabbed the color from rojo.place's dark mode. If it isn't what you'd have in mind, it should be easy enough to change... I think it looks nice enough, though.

Please let me know if I did anything wrong here!

image

### DIFF
--- a/src/web/ui.rs
+++ b/src/web/ui.rs
@@ -271,7 +271,7 @@ impl UiService {
                     </style>
                 </head>
 
-                <body>
+                <body style="background-color:#1b1b1d;">
                     { body }
                 </body>
             </html>


Dekkonot:
I'm hesitant to do this because light theme is an accessibility thing and if we have to have just light or just dark, it has to be light. Ideally we'd get a light/dark theme toggle but that's non-trivial to implement without JS and I don't wanna send you down that road.

How do you feel about the color from https://dom.rojo.space/? It's not dark, but it's a lot better than pure white IMO.
image

That one is `#e7e7e7`.

Dekkonot:
After discussion in the ROSS server we're going to go ahead and move forward with `#e7e7e7`. Going to close this PR rather than updating it so that I'm not mangling your intentions.

See #1080.

PR #1066, opened 2025-06-15
Fix auto connect for play mode

### LINKED ISSUE #1049: Plugin: Auto-connect while playtesting no longer works
This used to work a few months ago. Now, it doesn't connect at all. I'll update the issue when I have some time to figure out the last version it worked on.

armandmorosanu:
Recently, [a change was made to :IsRunMode()](https://devforum.roblox.com/t/runserviceisrunmode-always-returns-true-even-when-using-play-hereplay-in-roblox-studio/534992/15), breaking the plugin's auto-connect in play mode.

This PR resolves #1049 by changing the plugin to check for :IsRunning() and :IsStudio() instead of :IsRunMode().

### DIFF
--- a/plugin/src/App/init.lua
+++ b/plugin/src/App/init.lua
@@ -403,7 +403,8 @@ function App:releaseSyncLock()
 end
 
 function App:isAutoConnectPlaytestServerAvailable()
-	return RunService:IsRunMode()
+	return RunService:IsRunning()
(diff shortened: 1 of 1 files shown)


boatbomber (approved):

Dekkonot:
This probably needs a changelog entry, but otherwise looks good to me.

PR #1081, opened 2025-07-24
Fixed nil -> nil props showing up as failing in patch visualizer (plugin)

### LINKED ISSUE #1037: Referent properties that are `nil` are always marked as failing to sync by the plugin
Reported on Discord. `nil` referent properties show as failed to sync in the plugin's patch visualizer:

Image

Doesn't seem to have behavior consequences, so there's no fire, but it's annoying so we should fix it.

I suspect this is a consequence of #1005 because it was the last thing to change referents in the plugin but we'll see.

Sebastian2852:
Fixes #1037

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,11 @@
 # Rojo Changelog
 
 ## Unreleased
+* Fixed nil -> nil properties showing up as failing to sync in plugin's patch visualizer ([#1081])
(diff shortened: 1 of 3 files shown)


Dekkonot (approved):

Sebastian2852:
@Dekkonot Do I need to resolve the conflict in change log?

Dekkonot:
I would prefer you do just because it involves pushing to your `master` branch. GitHub gives me a scary warning when I try.

Dekkonot (approved):

Issue #943, opened 2024-07-19
Table diffing has unreadable text color in dark theme


Dekkonot:
![image](https://github.com/user-attachments/assets/9d2c9cdc-3068-46ad-b4a9-acb998c55530)

I don't believe this is how this should look. Ideally the text would be black, I think.

![image](https://github.com/user-attachments/assets/bb99c87c-da02-4550-b666-58036beea341)

PR #1084, opened 2025-07-27
Fix Table diff colors

Sebastian2852:
This PR is a fix for #943

I have updated table diffs to have proper colors in dark theme, I have added screenshots to reflect the new look. The UI forces the text to be black when a change has happened (addition, removal, edit) and in all other situations uses the same theme color as it did before, I skimmed through the theme and didn't find any colors that looked good in both light and dark theme here so that's why its forced to black. From my testing this looks pretty good but I am open to feedback!

Pasted Graphic
Table diff
Attrib2

_(Ignore attribute name lol)_

### DIFF
--- a/plugin/src/App/Components/TableDiffVisualizer/Dictionary.lua
+++ b/plugin/src/App/Components/TableDiffVisualizer/Dictionary.lua
@@ -114,7 +114,7 @@ function Dictionary:render()
 						TextXAlignment = Enum.TextXAlignment.Left,
 						FontFace = theme.Font.Main,
 						TextSize = theme.TextSize.Body,
-						TextColor3 = Color3.new(0, 0, 0),
+						TextColor3 = if line.patchType ~= "Remain" then Color3.new(0, 0, 0) else theme.Settings.Setting.DescriptionColor,
 						TextTruncate = Enum.TextTruncate.AtEnd,
 					}),
 					OldValue = e("Frame", {
@@ -125,7 +125,7 @@ function Dictionary:render()
 						e(DisplayValue, {
 							value = oldValue,
 							transparency = self.props.transparency,
-							textColor = Color3.new(0, 0, 0),
+							textColor = if line.patchType ~= "Remain" then Color3.new(0, 0, 0) else theme.Settings.Setting.DescriptionColor,
 						}),
 					}),
 					NewValue = e("Frame", {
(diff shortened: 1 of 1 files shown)


boatbomber:
Thanks for tackling this!

I think we can take this as an opportunity to improve the theme handling here instead of adding more if expressions.

https://github.com/rojo-rbx/rojo/blob/master/plugin%2Fsrc%2FApp%2FTheme.lua#L167

Could you update the theme.Diff to have `Diff.Text[patch type]` and `Diff.Background[patch type]`? Including Remain as a key in the themes as well will allow us to remove all of these annoying if expressions and just have an actual theme defined.

Sebastian2852:
@boatbomber I made the changes you requested, however since I am unfamiliar with this code base I am unsure if there are any other places this new theme needs to be used. If you could let me know if there are any other places that need the new theme I'd be happy to make those changes too.

Sebastian2852:
Jsut noticed but I forgot to do the `Background[PatchType]` will do it soon but im about to get on a flight so it might be a bit.

Sebastian2852:
Note: I have retained the old theme keys but updated their colors to bright pink. This serves as a visual indicator to identify any areas that have not yet been updated to the new theme. If any instances remain unchanged, the bright pink text or background should make it immediately apparent that it is not intended to appear that way.

plugin/src/App/Theme.lua
```diff
@@ -165,12 +165,29 @@ function StudioProvider:updateTheme()
...
+				Remove = if isDark then Color3.fromRGB(242, 125, 125) else Color3.fromRGB(150, 29, 29),
+				Edit = if isDark then Color3.fromRGB(120, 154, 248) else Color3.fromRGB(0, 70, 160),
+				Remain = studioTheme:GetColor(Enum.StudioStyleGuideColor.BrightText),
+			},
+
+			Text = {
+				Add = if isDark then Color3.new(0, 0, 0) else Color3.new(1, 1, 1),
+				Remove = if isDark then Color3.new(0, 0, 0) else Color3.new(1, 1, 1),
+				Edit = if isDark then Color3.new(0, 0, 0) else Color3.new(1, 1, 1),
+				Remain = studioTheme:GetColor(Enum.StudioStyleGuideColor.MainText), -- FROM: theme.Settings.Setting.DescriptionColor
```
boatbomber, line 189: The comment can be removed, as this text is now being treated differently from the text in settings so we don't need to track them as related

boatbomber (approved):
After that comment gets removed, looks good to merge! Thanks for contributing

PR #1085, opened 2025-07-29
Normalize paths in sourcemap generation

Dekkonot:


kennethloeffler (approved):
I won't block this PR over it, but we should add some tests for sourcemap generation to ensure that the paths are written as expected on each of the supported platforms. I'm also realizing that this command is completely untested; we should create an issue to add more tests here!

Issue #1091, opened 2025-08-01
`rojo sourcemap --absolute` option to emit absolute paths in the sourcemap


ffrostfall:
Useful when you need to move sourcemap files between directories

Dekkonot:
Idea seems sound to me. I've ran into this myself a few times.

PR #1092, opened 2025-08-01
Add option for emitting absolute paths to `rojo sourcemap`

ffrostfall:
I know that this is a *little* premature for a pull request, but I thought I would take the initiative since I need this for myself anyways and can't wait until the next release.

This PR implements `rojo sourcemap --emit-absolute-paths`. It uses `path::absolute()` over `.canonicalize()` because canonicalize, on Windows, emits UNC paths which are not well supported. `.canonicalize` also hits the file system, which could incur performance hits, which is not good.

Closes issue #1091

### DIFF
--- a/src/cli/sourcemap.rs
+++ b/src/cli/sourcemap.rs
@@ -1,7 +1,7 @@
 use std::{
     io::{BufWriter, Write},
     mem::forget,
-    path::{Path, PathBuf},
+    path::{self, Path, PathBuf},
 };
 
 use clap::Parser;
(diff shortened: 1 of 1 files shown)


src/cli/sourcemap.rs
```diff
@@ -178,20 +196,32 @@ fn recurse_create_node<'a>(
...
         .metadata()
         .relevant_paths
         .iter()
         // Not all paths listed as relevant are guaranteed to exist.
         .filter(|path| path.is_file())
         .map(|path| path.strip_prefix(project_dir).expect(PATH_STRIP_FAILED_ERR))
-        .map(|path| path.to_path_buf())
         .collect();
 
+    let mut file_path_buffs: Vec<PathBuf> = Vec::new();
```
Dekkonot, line 208: We can use `Vec<Cow<Path>>` here over `Vec<PathBuf>` to save allocations when the flag isn't passed. It'd involve changing the type of `SourcemapNode` a bit but it isn't a public interface so there's no concern there.

Dekkonot (approved):

PR #1096, opened 2025-08-04
Improvements to sync reminder UX

boatbomber:
## Overview
- On startup, if we find an active server and it matches the project you last synced, it can auto reconnect to it (off by default)
- Sync reminders can now be fullscreen to ensure you don't miss it (small popups by default)
- Sync reminders will check if there's an active server
- Sync reminders can poll in addition to the usual check on startup (on by default)
  - This can help users who open Studio first as their workflow
- Small fixes to notifications that only occurred when sending several notifications within a millisecond

Sync reminders behavior flow:
```mermaid
flowchart TD
    INIT(Plugin initialization) --> REMINDER_ENABLED
    INIT --> POLL[Poll]
    POLL --> POLL_ENABLED[Is polling enabled]
    POLL_ENABLED --> POLL
    POLL --> REMINDER_ENABLED
    REMINDER_ENABLED[Check if reminders are disabled] --> |Enabled| SCAN
    SCAN[Check for active server] -->|Server found| NOTIFY_SERVER[Project 'Example Here' is serving at localhost:34872. Would you like to connect?]
    SCAN --> |No server| CHECK_PRIOR[Check if this place has synced before]
    CHECK_PRIOR --> |Has synced before| NOTIFY_PRIOR[You synced project 'Example Here' to this place 2 days ago. Did you mean to run 'rojo serve' and then connect?]
    CHECK_PRIOR --> |Has not synced before | DO_NOTHING[Do nothing]
```

Settings:
image

Notify mode
image

Fullscreen mode
Screenshot of fullscreen notification


Dekkonot (commented):
The idea of a full-screen notification seems excessive to me but people seem to genuinely need it, so I like this. I imagine the auto-connect feature is going to be popular. I'm probably going to enable it on my work computer, if nothing else.

One concern is whether the auto-connect will still require the initial patch confirmation if that's enabled. I think as-implemented it still will, which might be annoying. We could add an option to bypass it but I think that gets dangerously close to "too many settings" territory.

Dekkonot:
> We already allow users to choose their confirmation behavior so auto connect will respect their choice

Yes, but there's no option to turn it off _only_ when auto-connect triggers. It kinda defeats the point of auto connect if you have to remember to confirm a patch still. But I very much would like to have a patch confirmation most of the time.

boatbomber:
Well when there's no changes in the initial sync then there's no confirmation popup anyway. Hopefully that smooths out most use cases of resuming where you left off

Dekkonot:
Going to merge it despite my earlier comments because honestly life is too short and we can always fix it later if it annoys people other than me

PR #1098, opened 2025-08-07
Add Arm64 builds to CI/Releases + build on ubuntu 22.04

Dekkonot:
As of not very long ago, GitHub supports arm64 runners for Linux and Windows in public repositories. That's us! We should support this now that we can.

I'm defaulting to `ubuntu-24.04-arm` rather than `ubuntu-22.04-arm` because it's also the version that `ubuntu-latest` uses.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -53,12 +53,12 @@ jobs:
         # https://doc.rust-lang.org/rustc/platform-support.html
         include:
           - host: linux
-            os: ubuntu-latest
+            os: ubuntu-22.04
             target: x86_64-unknown-linux-gnu
             label: linux-x86_64
 
           - host: linux
-            os: ubuntu-24.04-arm
+            os: ubuntu-22.04-arm
             target: aarch64-unknown-linux-gnu
             label: linux-aarch64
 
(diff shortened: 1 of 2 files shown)


boatbomber (approved):

.github/workflows/release.yml
```diff
@@ -57,11 +57,21 @@ jobs:
             target: x86_64-unknown-linux-gnu
             label: linux-x86_64
 
+          - host: linux
+            os: ubuntu-24.04-arm
```
kennethloeffler, line 61: Should we consider building on Ubuntu 22.04 for both x86 and ARM? iirc we've had some reports about incompatible glibc versions

kennethloeffler (commented):
Should we start using

.github/workflows/release.yml
Dekkonot: I was thinking that while I was making this PR, yeah. Wasn't sure if it was appropriate to change here but I do think we should do our builds on 22.04; I chose 24.02 for parity with `ubuntu-latest`.

Want me to change that for both x86 and arm64 in this PR, or should I do it in a new one?

kennethloeffler: Don't see anything wrong with making that change here, title would just need an update

kennethloeffler (approved):

Dekkonot:
I did fail to consider that the ubuntu-latest test is required to pass so merging this PR is going to require changing a repo setting, but I can take care of that once this is approved.

Issue #1107, opened 2025-08-28
Implement a flag to not automatically initialise a git repo


cosinewaves:
When creating a new project using rojo's built in `init` command, rojo automatically initialises a new git repo. This is useful I will not lie however in a few circumstances, I do not wish to use source control, especially for small test projects or working solo. I often find myself running commands in the terminal and deleting the related git files.

PR #1108, opened 2025-08-28
Add a flag to disable to auto-initialisation of git 

cosinewaves:
I created a issue regarding this feature, then tried to implement it myself. The issue is here #1107 for viewing.
All I basically did was create a new flag then pass that flag to each of the respective `Init_*` functions (place, plugin, model) with that boolean parameter then run a simple check.

### DIFF
--- a/src/cli/init.rs
+++ b/src/cli/init.rs
@@ -63,9 +63,9 @@ impl InitCommand {
         };
 
         match self.kind {
-            InitKind::Place => init_place(&base_path, project_params)?,
-            InitKind::Model => init_model(&base_path, project_params)?,
-            InitKind::Plugin => init_plugin(&base_path, project_params)?,
+            InitKind::Place => init_place(&base_path, project_params, self.no_git)?,
+            InitKind::Model => init_model(&base_path, project_params, self.no_git)?,
+            InitKind::Plugin => init_plugin(&base_path, project_params, self.no_git)?,
         }
 
         println!("Created project successfully.");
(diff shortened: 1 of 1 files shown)


src/cli/init.rs
```diff
@@ -1,31 +1,36 @@
...
-static PLUGIN_PROJECT: &str =
-    include_str!("../../assets/default-plugin-project/default.project.json");
+static PLUGIN_PROJECT: &str = include_str!(
+    "../../assets/default-plugin-project/default.project.json"
+);
 static PLUGIN_README: &str = include_str!("../../assets/default-plugin-project/README.md");
 static PLUGIN_GIT_IGNORE: &str = include_str!("../../assets/default-plugin-project/gitignore.txt");
 
+/// Initializes a new Rojo project.
+#[derive(Debug, Parser)]
```
Dekkonot, line 33:
```suggestion
```

These lines probably aren't intended to be here.

src/cli/init.rs
```diff
@@ -1,31 +1,36 @@
-use std::io::{self, Write};
-use std::path::{Path, PathBuf};
-use std::process::{Command, Stdio};
+use std::io::{ self, Write };
+use std::path::{ Path, PathBuf };
+use std::process::{ Command, Stdio };
```
Dekkonot, line 3: You seem to have your formatter configured in a way that isn't the rustfmt default; it's causing noise in here and will cause CI to fail.

Dekkonot (changes requested):
Your formatter is configured incorrectly, and this needs a changelog entry.

PR #1117, opened 2025-09-22
Refactor `init` command

Dekkonot:
This refactors the `init` command to make it less specialized per template type. The templates are now bundled in the executable in a manner similar to the plugin, and they're loaded from the bundled snapshot rather than stored as text.

The benefit of this is that the file-system structure of the templates can be directly copied, with only some basic logic done to e.g. rename `gitignore.txt` to `.gitignore` and render the templates.

Notably, this does not actually change the contents of the templates, nor does it improve QOL for the command any. This is just a pure refactor.

There are no tests for this refactor. I can add some if you want, but you'll be the one reviewing the snapshots so your call. 😉


Dekkonot:
Okay, this seemingly does not actually work for some reason. Perhaps tests are in order. Who knew.

PR #1129, opened 2025-10-13
Add musl release builds for x86_64 and aarch64

### LINKED ISSUE #1118: `rojo serve` on Ubuntu Linux throws error "`GLIBC_2.39' not found"
Apparently, Rojo doesn't work on Ubuntu Linux 22.04. A blog post [here](https://achris.me/posts/fix-glibc/) says that I have to use an older version of the OS to make this work.

This is the command output:
```bash
andrew@MSI:~/workspace/github.com/andrewens/joist$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 22.04.2 LTS
Release:        22.04
Codename:       jammy
andrew@MSI:~/workspace/github.com/andrewens/joist$ rojo serve
/home/andrew/.aftman/tool-storage/rojo-rbx/rojo/7.5.1/rojo: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.39' not found (required by /home/andrew/.aftman/tool-storage/rojo-rbx/rojo/7.5.1/rojo)
```

(mismatched backtick / apostrophe is not a typo -- it actually says that)

wengchaoxi:
Add musl static linking build support for Linux x86_64 and aarch64, providing better portability (no glibc dependency, closes #1118)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -62,6 +62,16 @@ jobs:
             target: aarch64-unknown-linux-gnu
             label: linux-aarch64
 
+          - host: linux
+            os: ubuntu-22.04
+            target: x86_64-unknown-linux-musl
+            label: linux-x86_64-musl
+
+          - host: linux
+            os: ubuntu-22.04-arm
+            target: aarch64-unknown-linux-musl
+            label: linux-aarch64-musl
+
           - host: windows
             os: windows-latest
             target: x86_64-pc-windows-msvc
@@ -90,6 +100,12 @@ jobs:
       - uses: actions/checkout@v4
         with:
           submodules: true
+
+      - name: Install musl build dependencies
+        if: contains(matrix.target, 'musl')
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends musl-tools
 
       - name: Install Rust
         uses: dtolnay/rust-toolchain@stable


wengchaoxi:
Hi @Dekkonot , thanks for your reply.

This PR introduces musl builds to solve **cross-distribution compatibility issues**.

**Benefits**: Static linking eliminates glibc version requirements, allowing a single binary to run on all Linux distributions.

**Potential downsides**:
- Increased binary size (overhead from statically linking libc)
- Cannot benefit from system glibc security updates (requires recompilation)

This is an additional build option that doesn't replace the existing GNU builds - users can choose the appropriate version based on their needs.

Reference: https://github.com/wengchaoxi/rojo/releases

wengchaoxi:
To provide more context on the specific problem:

Currently rojo is built on Ubuntu 22.04, which makes the binary unusable on older systems (for example, I encountered `./rojo: /lib/x86_64-linux-gnu/libm.so.6: version 'GLIBC_2.29' not found` error on an Ubuntu 18.04 server) due to different glibc versions. Other distributions (like CentOS, Debian, etc.) have similar issues.

Screenshot: 
image

Dekkonot:
I'm hesitant to accept this PR because I don't have any way to test the releases to make sure they work. Normally that's not a problem because we run CI, but I _also_ do not want to run an additional two Linux CI workflows for every PR and commit. We cache builds to improve CI times, and we are already close to the limit for caches on GitHub.

I also worry about how this would impact Foreman, Aftman, and Rokit since they use the architecture to pull release artifacts. For that reason, I'm going to close this PR. We may revisit this if we get more issues like the one you linked but for now, that's how it is. Thank you for the time though!

PR #1135, opened 2025-10-18
Improve sync fallback robustness

boatbomber:
# Changes

- Check that the instance id actually maps to something
- Don't crash if a child is not reparentable
- Revert if the instance itself is not reparentable


Dekkonot (approved):
I can't validate that this has fixed the problem people had because I don't have a consistent repro for that problem, but it seems sound to me.

PR #1140, opened 2025-10-23
Slightly improve initial sync hangs

boatbomber:
The majority of the hang is actually just `HttpService:JSONDecode` taking ages on massive trees.

This PR does not address that problem. It does, however, make a couple small improvements.

1. The patch tree is computed while on the loading page, rather than switching to a blank confirmation page while waiting for the tree to compute.
2. The patch tree computation runs across frames to avoid hanging Studio.
3. The loading page gets a line of text saying the current step.

https://github.com/user-attachments/assets/edf7f1fe-7d93-4cec-907e-b9154ddc446d

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,26 +1,56 @@
 # Rojo Changelog
 
+<!-- DIRECTIONS
+Thanks for contributing to Rojo! Please add your change to the "Unreleased" section below.
+
+Use the following format:
+
+```
 ## Unreleased
 
-* Added `sourcemap.json` into the defualt `.gitignore` files ([#1145])
+* Made a change to Rojo ([#0000])
+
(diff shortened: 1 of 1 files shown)


plugin/src/PatchTree.lua
```diff
@@ -16,6 +16,14 @@ local Types = require(Plugin.Types)
 local decodeValue = require(Plugin.Reconciler.decodeValue)
 local getProperty = require(Plugin.Reconciler.getProperty)
 
+local function yieldIfNeeded(clock)
+	if os.clock() - clock > 1 / 20 then
```
Dekkonot: 1/20 of a second feels like an arbitrary time. `task.wait()` is something like 1/30th of a second, right?

Dekkonot (approved):

PR #1149, opened 2025-11-04
Use buffer for ClassIcon EditableImages

gaymeowing:
Makes the ClassIcon component now use the EditableImage buffer API, instead of the old depricated table based writing API.

### DIFF
--- a/plugin/src/App/Components/ClassIcon.lua
+++ b/plugin/src/App/Components/ClassIcon.lua
@@ -47,8 +47,8 @@ local function getRecoloredClassIcon(className, color)
 
 	if iconProps and color then
 		--stylua: ignore
-		local success, editableImageSize, editableImagePixels = pcall(function(iconProps: { [any]: any }, color: Color3): (Vector2, buffer)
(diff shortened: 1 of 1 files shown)


plugin/src/App/Components/EditableImage.lua
```diff
@@ -12,15 +12,17 @@ function EditableImage:init()
...
 	if not image then
 		return
 	end
-	if not self.props.pixels then
+	if not props.pixels then
 		return
 	end
 
-	image:WritePixels(Vector2.zero, self.props.size, self.props.pixels)
+	image:WritePixelsBuffer(Vector2.zero, props.size, props.pixels)
```
Dekkonot: Most of these changes are superfluous. I appreciate the initiative for optimization, but creating images isn't a bottleneck to the point where we need to optimize table indexes to this extent.

Would rather that this was just changing `WritePixels` to `WritePixelsBuffer`.

Dekkonot (changes requested):
1. Please run StyLua on this so that CI passes
2. The style for this codebase is camelCase. Please swap your new function and variables to camelCase.

PR #1150, opened 2025-11-07
Remove pairs() and ipairs()

gaymeowing:
Removes legacy usage of `pairs()` and `ipairs()`, didn't fully touch EncodedValue because it seems like it relies on `ipairs()` only iterating over the array part for some of it.


Dekkonot (changes requested):
rbx_dom_lua is sourced from [rbx-dom](https://github.com/rojo-rbx/rbx-dom/tree/master/rbx_dom_lua). Changes to it will need to be done upstream and then the next time we update rbx_dom_lua it will be changed in this repository.

Otherwise, looks good.

PR #1159, opened 2025-11-12
Support .jsonc extension for all JSON files

Dekkonot:
In #1144, we swapped our parser over to JSONC for all JSON projects. However, we did not support `.jsonc` as an extension for any of Rojo's files. This causes problems with linters and other niceties.

This PR fixes that! It adds support for:
- .meta.jsonc
- .project.jsonc
- .model.jsonc
- .jsonc

This turned out to be more complicated than anticipated because a lot of the support for `default.project.json` and `.meta.json` files were hardcoded.

To that end:
- ServeSession no longer hardcodes `default.project.json` and instead uses a function built into `Project`
- Several of the built-in project functions now support both `default.project.json` and `default.project.jsonc`
- Metadata support has been refactored so that the metadata itself handles loading it rather than individual middleware calling `from_slice` and applying it, so that `.meta.jsonc` support could be done in one place

### DIFF
--- a/src/snapshot_middleware/meta_file.rs
+++ b/src/snapshot_middleware/meta_file.rs
@@ -165,6 +165,11 @@ pub struct DirectoryMetadata {
 }
 
 impl DirectoryMetadata {
+    /// Attempts to read an `init.meta`` file for the provided path, and if
+    /// one exists applies it.
+    ///
+    /// Also inserts the potential metadata paths into the snapshot's relevant
+    /// paths for convenience purposes.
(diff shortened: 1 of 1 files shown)


kennethloeffler (approved):
Nice, I quite like the new helpers for meta files!

ryanlua:
Looks like `.model.jsonc` isn't supported? I've been looking at this PR and see the code for `.model.jsonc,` but for some reason it doesn't want to sync.

PR #1162, opened 2025-11-16
Change rbx_dom_lua to use EncodingService's base64 functions

Redusofficial:


Dekkonot:
rbx-dom-lua is pulled from upstream here: https://github.com/rojo-rbx/rbx-dom/tree/master/rbx_dom_lua/

If you submit a PR there, it'll be pulled into the Rojo plugin when we next do a release.

Issue #1182, opened 2025-12-01
Bug: MeshPart/UnionOperation fails to sync on reconnect - "URL too long" error


Froredion:
## Summary

When reconnecting Rojo for initial synchronization, MeshPart and UnionOperation instances are created without their mesh data (MeshId/MeshContent is empty). This is caused by two issues:

1. **Roblox behavior change**: `Instance.new("MeshPart")` now **succeeds** instead of failing, creating an empty MeshPart
2. **URL length limit**: The fallback mechanism fails with "URL too long" when serializing many instances

## Environment

- Rojo version: 7.6.x?
- Roblox Studio: Latest (as of November 2025)
- OS: Windows 11

### Issue 1: `Instance.new("MeshPart")` behavior changed

The current code in `plugin/src/Reconciler/reify.lua` assumes `Instance.new("MeshPart")` will **fail**:

```lua
local createSuccess, instance = pcall(Instance.new, virtualInstance.ClassName)

if not createSuccess then
    -- Mark this instance and all its children as unapplied
    -- The fallback sync mechanism will handle it
    addAllToPatch(unappliedPatch, virtualInstances, id)
    return
end
```

**Expected behavior**: `Instance.new("MeshPart")` fails → instance goes to fallback → SerializationService creates proper MeshPart with mesh data

**Actual behavior**: `Instance.new("MeshPart")` **succeeds** → creates empty MeshPart → attempts to set `MeshContent` property → **fails because MeshContent is read-only after creation**

Debug output showing the issue:
```
[MESHDBG] reify: Instance.new('MeshPart') for id= 5ef5631551f6a828dc02e4daa5df4001 name= Ball
[MESHDBG] reify: createSuccess= true instance= Ball
...
[MESHDBG] reify: setProperty MeshPart . MeshContent = Content success= false  ← FAILS!
```

### Issue 2: Fallback fails with "URL too long"

Even though the instance ends up in the unapplied patch (due to failed properties), the fallback mechanism in `ServeSession:__replaceInstances()` fails when there are many instances because all instance IDs are passed in the URL:

```
[MESHDBG] __replaceInstances: calling serialize API with 1983 ids (includes Ball)
[MESHDBG] __replaceInstances: serialize API ERROR: Unknown HTTP error: URL too long
```

The serialize API endpoint puts all IDs in the URL query string, which exceeds the HTTP URL length limit (~2000 characters) when there are many instances.

## Steps to Reproduce

[...]

PR #1184, opened 2025-12-01
Fix MeshPart/UnionOperation losing mesh data on initial sync

Froredion:
This fix addresses two issues that cause MeshPart and UnionOperation instances to lose their mesh data (MeshId/MeshContent) when reconnecting:

1. Force fallback for mesh-based classes: Instance.new("MeshPart") now succeeds in Roblox (behavior change) but creates an empty mesh with read-only MeshContent property. Force MeshPart, UnionOperation, IntersectOperation, and NegateOperation to use SerializationService fallback mechanism which can properly create these instances with their mesh data.

2. Batch serialize/refPatch API calls: The fallback mechanism was failing with "URL too long" errors when many instances needed serialization because all IDs were concatenated into a single URL. Now processes IDs in batches of 50 to stay within URL length limits.

Fix for this issue: https://github.com/rojo-rbx/rojo/issues/1182

**NOTE: Maybe how it's coded can be coded better?**

(PS: Untested)
(Only copied code from our working forked Repo)

### DIFF
--- a/plugin/src/Reconciler/reify.lua
+++ b/plugin/src/Reconciler/reify.lua
@@ -46,7 +46,16 @@ function reifyInstanceInner(unappliedPatch, deferredRefs, instanceMap, virtualIn
 	-- something that requires higher security than we have.
 	local createSuccess, instance = pcall(Instance.new, virtualInstance.ClassName)
 
-	if not createSuccess then
+	-- MeshPart and UnionOperation require SerializationService to properly
+	-- set their mesh data (MeshId/MeshContent). Instance.new() now succeeds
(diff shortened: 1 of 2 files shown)


kennethloeffler (commented):
Hello, thanks for opening a PR! 

I was able to reproduce the bug, and it appears entirely caused by URI length limits. The proposed solution (making multiple requests) does work, but I'd prefer an alternative solution that doesn't incur additional latency or network traffic. 

I will create an issue with reproduction steps and open a different PR with an alternative solution, but thanks for pointing out this problem.

Froredion:
More robust fix continuation: https://github.com/rojo-rbx/rojo/pull/1192

PR #1216, opened 2026-01-24
Fix "Always" confirmation behavior setting which currently works like "Initial"

warsam:
Summary
Implements the "Always" confirmation behavior setting, which prompts for confirmation on every sync patch (not just the initial sync). Changes that arrive during confirmation are merged into the pending patch and the UI updates in real-time.

Implementation
- Add "Always" option to confirmationBehavior setting
- Route WebSocket patches through the confirmation callback when "Always" is enabled
- Refactor ServeSession to connect WebSocket before showing confirmation, allowing patches to merge during confirmation
- Add PatchSet.merge() function that properly reconciles conflicts:
  - Cancels additions when they're subsequently removed
  - Combines updates for the same instance
  - Removes changes that revert to current instance state
- Auto-dismiss confirmation dialog when all changes have been reverted (patch becomes empty)
- Fix UI not transitioning back to Connected after confirming ongoing patches
- Fix DomLabel Flipper motor cleanup on unmount to prevent Roact errors
- Add unit tests for PatchSet.merge() and related functions

Test Plan
- [ ]  Set confirmationBehavior to "Always" in plugin settings
- [ ]  Connect to a Rojo server and verify the initial sync prompts for confirmation
 - [ ]  Accept the initial sync and verify the UI transitions to "Connected"
 - [ ]  Make a file change and verify a new confirmation dialog appears
 - [ ]  While confirmation is showing, make additional file changes and verify they merge into the pending patch (UI updates in real-time)
 - [ ]  Revert a file change during confirmation and verify it's removed from the pending patch
 - [ ]  Revert all changes during confirmation and verify the dialog auto-dismisses
 - [ ]  Test "Abort" option disconnects from the server
 - [ ]  Verify no Roact errors appear when rapidly opening/closing the confirmation dialog


warsam:
> I've never heard of someone wanting Rojo to halt and popup after every single change.

To clarify - it doesn't pop up for every individual change. Changes are batched and the confirmation dialog updates in real-time. You approve once when you're ready.

> The Initial setting has some cases where it doesn't prompt, so Always was added as a stricter Initial.

What are those cases? If they both only apply to initial sync, why have two options instead of just fixing "Initial"?

boatbomber:
>What are those cases? If they both only apply to initial sync, why have two options instead of just fixing "Initial"?

It's actually the other way around, Initial was added because Always was annoying people. Initial only prompts you on the first time you sync a particular project per Studio session.

See #773 and #774 for more context.

warsam:
Thanks for the context @boatbomber! Looking at #773 and #774, the goal was making confirmation less annoying by default, not removing it. I see that Dekkonot noted it's worth having "for new games" and you said disabling it entirely "is just avoiding addressing the root issue." That's why "Always" was kept as an option. I can confirm that by my own desire for it as a new rojo user.

This PR just makes that option work as labeled when someone deliberately enables it. Right now selecting "Always" silently behaves like "Initial", which I found confusing.

It also addresses the "hot workflows of constant connect/disconnect" you mentioned in #774. Patches now merge while the confirmation dialog is open, so you see the final merged state when you hit "accept". So you now only need to accept once at the end of multiple changes. The dialog auto-dismisses if all changes revert (as I didn't think it was useful to accept a change that does nothing), so it shouldn't be as noisy as the old "Always" behavior was.

If "Always" shouldn't exist at all, I think the right fix is removing it from the UI rather than keeping a broken option. Happy to do that instead if preferred?

Dekkonot:
> If "Always" shouldn't exist at all, I think the right fix is removing it from the UI rather than keeping a broken option. Happy to do that instead if preferred?

As a point of clarity, the setting isn't broken, it's just misleading. It's behaving exactly as it was implemented to do, which is prompting the user for the initial sync every time. In hindsight though, we probably could have named the setting something better.

I don't really see the use case of a setting for prompting on every sync. I'd much rather we just rename the setting to make it more obvious what it does.

boatbomber:
I agree, it's behaving exactly as intended. We just need a better name for it.

If we changed its behavior _that_ would be broken since people who had Always selected would suddenly have confirmation popups unexpectedly.

warsam:
Gotcha, makes sense. I'll leave "Always" as-is. I've reworked this to add "Every Change" as a new option instead. No existing behavior is changed.

warsam:
Rebased on master and updated the CHANGELOG. To recap after the discussion above: Always is left untouched, and I added Every Change as a new opt-in option that prompts on every sync patch. Patches merge into the pending confirmation dialog while it's open, so it stays usable for hot edit workflows rather than popping repeatedly.

Happy to also rename Always to something clearer (per @Dekkonot's point that it's misleading) and bundle it here, if that'd help land this.

PR #1220, opened 2026-01-28
Fix server crash when deleting folders

warsam:
Handle the case where a folder's parent path no longer exists when processing VFS removal events. This can happen when a folder and its contents are deleted together, causing a race between removal events.

Added a regression test for this scenario.


warsam:
CI should be passing now with the cross-platform snapshot fix. Could someone please approve the workflow run?

PR #1226, opened 2026-02-14
feat: add --base flag to upload command for merging with place files

ardrous:
Currently, `rojo upload` builds the project and does not take any pre-existing place files. My implementation is to add a --base flag for backwards compatibility.


Dekkonot (commented):
Thanks for the PR! A few things:

1. Supporting this just for `rojo upload` isn't desirable. Adding support to `rojo build` would make this significantly easier to test, and improve the usability.
2. This is akin to just putting a Rojo project over top a Roblox file. This means that unknown instances aren't deleted. This could be a problem if you were e.g. updating packages and expecting this command to take care any lingering files.
3. There aren't any tests for this to verify it's working. Writing tests for `rojo upload` is on the agenda still, but it'd be nice to at least verify the outputted rbxl is what you want it to be (see: adding support for this to `rojo build`)
4. I have performance concerns with the recursive design here. Rojo projects and Roblox games can hundreds of thousands of Instances in them and Instance trees can be arbitrarily deep. Those combine to make recursive navigation of the DataModel a problem.
5. We support uploading model files, so limiting `--base` to only using rbxl and rbxlx files is wrong.

ardrous:
Thank you for your swift response, I have made the changes based on your feedback. I ran the tests and they all work fine, and I have been using it on one of my games and it has been working perfectly without any issues. If there are any issues let me know.

src/merge_base.rs
```diff
@@ -0,0 +1,371 @@
...
+use crate::snapshot::RojoTree;
+
+/// A pair of (base instance, Rojo instance) whose children still need merging.
+struct MergePair {
+    base_ref: Ref,
+    rojo_ref: Ref,
+}
+
+/// Reads a base Roblox file (.rbxl, .rbxlx, .rbxm, .rbxmx) into a WeakDom.
+pub fn read_base_dom(path: &Path) -> anyhow::Result<WeakDom> {
```
Dekkonot, line 20: This code is essentially identical to code that exists in the syncback CLI but it doesn't do any of the required post-processing for model files that is generally required. While I have long-term ambitions to move the code elsewhere, for now this should probably just be a 1:1 rip of the code from syncback. It simplifies the logic by making it so you don't have to treat models and places differently when performing matching.

src/merge_base.rs
```diff
@@ -0,0 +1,371 @@
...
+            extension
+        ),
+    }
+}
+
+/// Merges the Rojo project tree into a base WeakDom.
+///
+/// For place projects (root class is "DataModel"), iteratively merges Rojo's
+/// children into matching base children by name and className. For model
+/// projects, clones the entire Rojo tree as a child of the base root.
```
Dekkonot, line 51: Once the `read_base_dom` function is changed to pre-process model DOMS, we can just treat them the same. This is almost certainly the desired and expected behavior for merging a Rojo project into a file anyway, so it works out.

src/merge_base.rs
```diff
@@ -0,0 +1,371 @@
...
+                    base.transfer_within(cloned_ref, pair.base_ref);
+                }
+            }
+        }
+
+        // Delete unmatched base children if ignore_unknown_instances is false.
+        let ignore_unknown = rojo
+            .get_metadata(pair.rojo_ref)
+            .map(|m| m.ignore_unknown_instances)
+            .unwrap_or(true);
```
Dekkonot, line 134: The default for `ignore_unknown_instances` is `false`, not `true`.

src/merge_base.rs
```diff
@@ -0,0 +1,371 @@
...
+        .expect("Rojo tree should have a root");
+    let base_root_ref = base.root_ref();
+
+    // For model projects (root is not DataModel), clone entire Rojo tree
+    // as a child of the base root.
+    if rojo_root.class.as_str() != "DataModel" {
+        let cloned_ref = rojo_inner.clone_into_external(rojo_root_id, &mut base);
+        base.transfer_within(cloned_ref, base_root_ref);
+        return Ok(base);
+    }
```
Dekkonot, line 71: This behavior is entirely unexpected and should be removed. I would rather this behave the same for both models and places.

src/merge_base.rs
```diff
@@ -0,0 +1,371 @@
...
+
+    // For place projects: iteratively merge children using a stack.
+    let mut stack = vec![MergePair {
+        base_ref: base_root_ref,
+        rojo_ref: rojo_root_id,
+    }];
+
+    while let Some(pair) = stack.pop() {
+        let rojo_children: Vec<Ref> = rojo_inner
```
Dekkonot, line 80: This doesn't need to be a vector.

Dekkonot (changes requested):
Overall looks good, but I think we should try to make the behavior for models and places behave the same.

PR #1228, opened 2026-02-17
Fix change_processor crash when path is a directory during two-way sync

### LINKED ISSUE #1206: Change processor tries to write to directory as if it were a file sometimes
Logs:
```
Rojo server listening:
  Address: localhost
  Port:    34872

Visit http://localhost:34872/ in your browser for more information.
[ERROR rojo] Rojo crashed! You are running Rojo 7.6.1.
[ERROR rojo] This is probably a Rojo bug.
[ERROR rojo] 
[ERROR rojo] Please consider filing an issue: https://github.com/rojo-rbx/rojo/issues
[ERROR rojo] 
[ERROR rojo] Details: called `Result::unwrap()` on an `Err` value: Os { code: 21, kind: IsADirectory, message: "Is a directory" }
[ERROR rojo] in file src/change_processor.rs on line 229
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```
Error happens as soon as I connect

adv-andrew:
Fixes #1206

I was running into the same crash described in the issue where Rojo would panic during two-way sync. After looking into it, the problem is that `fs::write` and `fs::remove_file` in `handle_tree_event` can receive a directory path instead of a file path when dealing with init scripts.

What I did:
- Added a function `resolve_init_path` that checks if the path is a directory and if so, looks for the init file inside it (init.lua, init.luau, etc.)
- Used this in the write path so it writes to the actual init file instead of trying to write to a directory
- For the remove path, if it's a directory I used `remove_dir_all` instead of `remove_file`
- Instead of panicking, it now logs a warning if something goes wrong

I tried to match the same init file priority order that's used in `snapshot_middleware/mod.rs`.

Let me know if there's anything I should change!

### DIFF
--- a/src/change_processor.rs
+++ b/src/change_processor.rs
@@ -2,7 +2,7 @@ use crossbeam_channel::{select, Receiver, RecvError, Sender};
 use jod_thread::JoinHandle;
 use memofs::{IoResultExt, Vfs, VfsEvent};
 use rbx_dom_weak::types::{Ref, Variant};
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::{
     fs,
     sync::{Arc, Mutex},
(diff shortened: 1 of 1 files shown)


src/change_processor.rs
```diff
@@ -244,7 +256,17 @@ impl JobThreadContext {
                                 match instigating_source {
                                     InstigatingSource::Path(path) => {
                                         if let Some(Variant::String(value)) = changed_value {
-                                            fs::write(path, value).unwrap();
+                                            match resolve_init_path(path) {
```
Dekkonot, line 259: You should be able to use `snapshot_middleware::get_dir_middleware` here instead. You'll have to change the visibility of that function but I'd prefer that to recreating the same behavior of finding an init path. Keeps things in one place and all that.

Dekkonot (changes requested):
This needs a changelog entry. There's instructions for that in [CHANGELOG.md](https://github.com/rojo-rbx/rojo/blob/master/CHANGELOG.md?plain=1).

PR #1244, opened 2026-04-03
Fix syncback not removing stale $properties at engine defaults

### LINKED ISSUE #1243: Syncback doesn't update $properties when Studio value is at engine default
## Bug

When a property in a project file's `$properties` block has a non-default value, and Studio changes that property to its engine default, `rojo syncback` doesn't update or remove the stale entry. The project file retains the old non-default value.

## Repro

**1. Create `test.project.json`:**
```json
{
  "name": "SyncbackTest",
  "tree": {
    "$className": "DataModel",
    "Workspace": {
      "$className": "Workspace",
      "TestPart": {
        "$className": "Part",
        "$properties": {
          "Transparency": 1.0,
          "Anchored": true,
          "Color": [1.0, 0.0, 0.0]
        }
      }
    }
  }
}
```

**2. Build and open:**
```bash
rojo build -o test.rbxl test.project.json
```
Open `test.rbxl` in Studio.

**3. In Studio:** Set `TestPart.Transparency` to `0` (engine default) and `Color` to blue. Save.

**4. Syncback:**
```bash
rojo syncback -y --input test.rbxl test.project.json
```

**5. Result:**
```json
"$properties": {
  "Anchored": true,
  "Color": { "Color3uint8": [0, 0, 255] },
  "Transparency": 1.0
}
```

`Color` updated to blue correctly. **`Transparency` stayed at `1.0`** even though Studio has it at `0`.

## Expected

`Transparency` should be removed from `$properties` (since it's at the engine default) or updated to `0`.

## Root Cause

The Roblox binary format omits properties at their engine default values. When rojo deserializes the rbxl, default-valued properties are absent from the instance data. `filter_properties` in `property_filter.rs` correctly excludes them from the filtered set (they match the default). But `project_node_property_syncback` in `project.rs` only inserts/updates from the filtered set — it never removes stale entries that are no longer present.

## Impact

Affects any `$properties` entry changed to its engine default in Studio. We hit this with Lighting service properties (`ColorShift_Bottom`, `EnvironmentDiffuseScale`, `FogEnd`, `FogStart`) — syncback retained old preset values after switching to a lighting preset where those properties are at defaults. Rebuilding the place then produced incorrect lighting because the stale project file values overrode the defaults.

## Fix

I have a fix ready that adds a `retain` pass after the insertion loop in `project_node_property_syncback`. It removes entries from `node.properties` where:
1. The property is NOT in the filtered set (meaning it's at default or absent)
2. The property has a known default in `rbx_reflection_database` (safety check — unknown properties are conservatively kept)

Also updates the reserialization logic to detect when stale properties were removed, since `project_node_should_reserialize` only iterates forward through current properties and wouldn't detect a removal.

Compiles clean, all 88 existing tests pass. Will open a PR shortly.

floydhorng:
## Summary

Fixes #1243

When a property in a project file's `$properties` has a non-default value and Studio resets it to the engine default, syncback didn't remove the stale entry. Rebuilding the place then produced incorrect values because the stale project file entry overrode the default.

**Root cause:** The Roblox binary format omits default-valued properties. `filter_properties` correctly excludes them, but `project_node_property_syncback` only inserted/updated from the filtered set — it never removed stale entries.

## Changes

- **`project_node_property_syncback`**: Added a `retain` pass after the insertion loop that removes entries where the property is NOT in the filtered set AND has a known default in `rbx_reflection_database` (unknown properties are conservatively kept)
- **Return value**: `project_node_property_syncback` now returns `bool` indicating whether stale properties were removed
- **Reserialization**: Added `removed_stale_properties` flag so the project file gets rewritten when entries are removed (`project_node_should_reserialize` only compares forward through current properties and wouldn't detect a removal)

## Testing

- All 88 existing tests pass
- Verified with a manual repro: Part with `Transparency: 1.0` in `$properties`, changed to `0` in Studio, syncback now correctly removes the entry
- Verified `rbx_reflection_database` includes inherited defaults (e.g. `Part` includes `Transparency` from `BasePart`)


kennethloeffler:
Nice, [this did come up in review for #937](https://github.com/rojo-rbx/rojo/pull/937#discussion_r1677168769), but we didn't end up creating an issue to follow up on it. Thank you!

Code looks good to me, but @Dekkonot should probably also take a look at it. I also want to run this on some real projects to see if anything erroneous happens, will report back once I've done that.

Dekkonot (changes requested):
I would love a test for this to verify it was fixed and prevent regressions. I have one written if you don't mind me just pushing it to your branch, but it was pretty straightforward. Your repro on the issue worked well.

Dekkonot:
Going to just assume you're fine with me pushing to your branch. Apologies if you are not.

Dekkonot (approved):

PR #1008, opened 2025-01-23
Add Support for Plugin Scripts

sasial-dev:
The plugin RunContext was added recently [here](https://devforum.roblox.com/t/3401339). This PR adds support for them in rojo with a new `.plugin.lua(u)` extension.

This one *does* force RunContext regardless of the `emitLegacyScripts` setting because as per the devforum post these plugin scripts have special capabilities that cannot be replicated in a legacy manner (actors support, autocomplete for `plugin` etc).

PR #1252, opened 2026-04-29
Fix init.plugin.lua/init.plugin.luau not being supported 

### LINKED ISSUE #1242: `init.plugin.lua` creates a Folder instead of a Script
If a file is named `init.plugin.lua`, it emits a Folder containing a Script (with the proper "Plugin" RunContext), instead of the expected Script containing siblings of the directory.

kennethloeffler:
Closes #1242.

It looks like when we added #1008, we did not add support for init.plugin.lua/init.plugin.luau. This PR adds the necessary logic to support this pattern, and a build test to make sure it continues working.


boatbomber (dismissed):

Dekkonot (changes requested):
This requires some changes to the syncback code for Luau. Specifically, `syncback_lua_init` doesn't accept the new middleware as input so it'll raise an error if it encounters a plugin init file.

Dekkonot (dismissed):

Dekkonot (approved):

PR #1256, opened 2026-05-05
Add ignorable glob support

EgoMoose:
This PR changes `globIgnorePaths` such that it now supports gitignore-style negation: a leading `!` re-includes paths matched by an earlier glob. The last matching pattern in the list determines whether a path is ignored. Use `\!` to match a literal `!` at the start of a pattern.

This should be backwards compatible with existing projects as opposed to adding something akin to extended globbing. 

This makes it possible to do things like have a single source folder which contains both your client and server code, but be able to sync those code paths to separate locations. 

You can still use existing glob ignore path functionality:
```json
{
	"name": "client-src",
	"globIgnorePaths": [
		"**/*Server.{lua,luau}"
	],
	"tree": {
		"$path": "src"
	}
}
```

Or you can use the new negation feature:
```json
{
	"name": "server-src",
	"globIgnorePaths": [
		"**/*.{lua,luau}",
		"!**/*Server.{lua,luau}"
	],
	"tree": {
		"$path": "src"
	}
}
```

Using these together you can do some pretty useful things when syncing:
```json
{
	"name": "example",
	"tree": {
		"$className": "DataModel",
		"ReplicatedStorage": {
			"ClientSource": {
				"$path": "client-src.project.json"
			}
		},
		"ServerScriptService": {
			"$className": "ServerScriptService",
			"ServerSource": {
				"$path": "server-src.project.json"
			}
		}
	}
}
```


Dekkonot (changes requested):
We also use globs for syncback here: https://github.com/rojo-rbx/rojo/blob/master/src/syncback/mod.rs#L389

It seems prudent to also support negation for it as well.

Dekkonot (approved):
Looks good to me. Thank you!

PR #1264, opened 2026-06-01
Remove Reflection Database Cow Remnants

krakow10:
- Depends on https://github.com/rojo-rbx/rbx-dom/pull/528

A minor cleanup after the breaking database change is merged that includes some bits that don't show up in any default clippy lints.


Dekkonot (approved):

Dekkonot:
This will be merged when we update rbx-dom

PR #1267, opened 2026-06-02
Graceful errors instead of crashing

boatbomber:
Full disclosure, I used Claude for this one. It's nothing complex.

Replaces a number of user-reachable panics with proper errors so Rojo fails gracefully instead of crashing. This targets the most common panic reports we get such as read-only/locked files and the serve port already being in use.

### Changes

- **Port already in use:** `rojo serve` now uses `Server::try_bind` and prints a clear error (with a `--port` hint) instead of panicking. The "listening" message is also deferred until binding actually succeeds, so you no longer see a success message immediately followed by a crash.
  image
- **Read-only / locked files:** writing `Source` and removing files during live two-way sync now log an error and keep the session alive instead of aborting the whole process.
- **Inaccessible working directory:** `resolve_path` returns a clear error instead of unwrapping `current_dir()`.
- **Filesystem watcher init:** `StdBackend::new` / `Vfs::new_default` now return `io::Result`, so a watcher failure in restricted environments is a clean error rather than a panic.
- **`rojo init` templates:** bundled-template deserialization returns errors instead of `panic!`.
- **Watch mode:** runtime creation and the watch loop in `build`/`sourcemap` return errors / exit cleanly.
- **Web layer:** response builders route through a non-panicking helper; non-UTF8 paths serialize lossily instead of panicking; CSV/localization JSON encoding propagates errors.

### Notes

- `memofs`: `StdBackend::new()` and `Vfs::new_default()` now return `io::Result`. The unused fallible `Default for StdBackend` impl was removed.

### DIFF
--- a/src/change_processor.rs
+++ b/src/change_processor.rs
@@ -200,7 +200,15 @@ impl JobThreadContext {
                 if let Some(instance) = tree.get_instance(id) {
                     if let Some(instigating_source) = &instance.metadata().instigating_source {
                         match instigating_source {
-                            InstigatingSource::Path(path) => fs::remove_file(path).unwrap(),
+                            InstigatingSource::Path(path) => {
+                                if let Err(err) = fs::remove_file(path) {
+                                    log::error!(
+                                        "Failed to remove file {}: {}",
+                                        path.display(),
+                                        err
+                                    );
+                                }
+                            }
                             InstigatingSource::ProjectNode { .. } => {
                                 log::warn!(
                                     "Cannot remove instance {:?}, it's from a project file",
@@ -244,7 +252,13 @@ impl JobThreadContext {
                                 match instigating_source {
                                     InstigatingSource::Path(path) => {
                                         if let Some(Variant::String(value)) = changed_value {
-                                            fs::write(path, value).unwrap();
+                                            if let Err(err) = fs::write(path, value) {
+                                                log::error!(
+                                                    "Failed to write file {}: {}",
+                                                    path.display(),
+                                                    err
+                                                );
+                                            }
                                         } else {
                                             log::warn!("Cannot change Source to non-string value.");
                                         }
--- a/src/cli/init.rs
+++ b/src/cli/init.rs
@@ -9,7 +9,7 @@ use std::{
     io::{self, Write},
 };
 
-use anyhow::{bail, format_err};
+use anyhow::{bail, format_err, Context};
 use clap::Parser;
 use fs_err as fs;
 use fs_err::OpenOptions;
@@ -42,9 +42,9 @@ pub struct InitCommand {
 
 impl InitCommand {
     pub fn run(self) -> anyhow::Result<()> {
-        let template = self.kind.template();
+        let template = self.kind.template()?;
 
-        let base_path = resolve_path(&self.path);
+        let base_path = resolve_path(&self.path)?;
         fs::create_dir_all(&base_path)?;
 
         let canonical = fs::canonicalize(&base_path)?;
@@ -128,7 +128,7 @@ pub enum InitKind {
 }
 
 impl InitKind {
-    fn template(&self) -> InMemoryFs {
+    fn template(&self) -> anyhow::Result<InMemoryFs> {
         let template_path = match self {
             Self::Place => "place",
             Self::Model => "model",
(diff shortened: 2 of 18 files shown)


src/web/mod.rs
```diff
@@ -53,9 +59,25 @@ impl LiveServer {
...
-        let rt = Runtime::new().unwrap();
+        let rt = Runtime::new().context("Failed to start the async runtime for the web server")?;
         let _guard = rt.enter();
-        let server = Server::bind(&address).serve(make_service);
-        rt.block_on(server).unwrap();
+        let server = Server::try_bind(&address)
+            .with_context(|| {
+                format!(
+                    "Could not start the Rojo server on {address}.\n\
+                     The address may already be in use \u{2014} another Rojo server might already \
```
Dekkonot: Something very funny about an AI inserting an em dash like this.

src/change_processor.rs
```diff
@@ -200,7 +200,15 @@ impl JobThreadContext {
                 if let Some(instance) = tree.get_instance(id) {
                     if let Some(instigating_source) = &instance.metadata().instigating_source {
                         match instigating_source {
-                            InstigatingSource::Path(path) => fs::remove_file(path).unwrap(),
+                            InstigatingSource::Path(path) => {
+                                if let Err(err) = fs::remove_file(path) {
```
Dekkonot, line 204: While we're here do we want to swap this to Vfs? Not sure it's worth the hassle but it's a thought.

src/path_serializer.rs
```diff
@@ -7,13 +7,12 @@ use serde::{ser::SerializeSeq, Serialize, Serializer};
 
 /// Converts the provided value into a String with all directory separators
 /// converted into `/`.
+///
+/// Paths that contain invalid Unicode are converted lossily (invalid sequences
+/// become the replacement character), since such paths cannot be represented
+/// faithfully in a UTF-8 string regardless.
 pub fn display_absolute<T: AsRef<Path>>(path: T) -> String {
```
Dekkonot, line 10: I would actually rather this fail, because I'm not entirely sure where we use this function and I am weary of potentially mangling a path.

src/cli/init.rs
```diff
@@ -128,28 +128,32 @@ pub enum InitKind {
...
-                fs.load_snapshot("", template)
-                    .expect("loading a template in memory should never fail");
-                fs
-            } else {
-                panic!("template for project type {:?} is missing", self)
-            }
-        } else {
-            panic!("Rojo's templates were packed as a file instead of a directory")
-        }
+            .context("Rojo's templates were not properly packed into Rojo's binary. This is a bug in Rojo; please file an issue.")?;
```
Dekkonot, line 139: The panic handler for Rojo already suggests filing an issue. Seems redundant.

Dekkonot (commented):
Needs changelog entry for memofs and the errors in the `init` subcommand are redundant. Otherwise seems reasonable. Long-term, we probably should try to switch to a dedicated error type instead of using Anyhow because I'm not sure how efficient this all is, but that's the stretchiest of stretch goals.

src/change_processor.rs
boatbomber: I'd like to keep this PR scoped to only modifying the error handling

src/web/mod.rs
boatbomber: The escaped em dash goes hard

src/path_serializer.rs
boatbomber: Fair enough!

src/cli/init.rs
boatbomber: Well the point is we're not panicking anymore so we include the message in the error

Dekkonot: If you had asked me five minutes ago if we panicked eventually even on errors I would have said we did. Shows what I know!

Fair enough. I withdraw the complaint.

Dekkonot (approved):

src/cli/init.rs
boatbomber: That's because we just use panic everywhere and have no handling so you're used to always seeing it 😆 That's why I made this PR, it's silly that Rojo freaks out and and asks you to report a bug for everything, even user mistakes

PR #1270, opened 2026-06-02
Add origin/host validation and warning for exposed serves

boatbomber:
### Summary

`rojo serve` exposes an unauthenticated HTTP API. This PR adds defense-in-depth so that a local bind is safe against malicious web pages, and so that exposing the server to the network is an informed, explicit choice.

Three changes:

1. DNS-rebinding protection (`src/web/origin.rs`).
  When the server is bound to an address we can build an allowlist for, every request's `Host` header (and its `Origin` header when present) must refer to an address we recognize, or the request is rejected. This blocks DNS rebinding, where an attacker's page repoints its own hostname at the bind address so the browser treats it as same-origin: the rebound request still carries the attacker's hostname in `Host` (a domain name, not one of the IP literals we accept), and the browser won't let page JS forge that header.

   The allowlist depends on the bind:
   - Loopback (the default): `localhost` and the loopback literals `127.0.0.1` / `::1`, with a matching port.
   - A specific private/LAN address (e.g. `192.168.1.5`, for testing from another machine): the above plus that exact bind IP. Clients connect to it by IP, since a hostname is indistinguishable from an attacker domain.
   - Unspecified (`0.0.0.0` / `::`) or public: enforcement is disabled unless allowed hosts are specified (see trade-offs).

   The allowlist can be extended with explicit hostnames or IPs, either with the `--allowed-hosts` CLI option or a project's `serveAllowedHosts` field (for example `mypc.lan`, to reach a LAN server by name instead of by IP). The CLI option overrides the project field when both are set. Listing any host also turns validation on for an unspecified or public bind that would otherwise disable it, so that bind then accepts only localhost, the bind IP, and the listed hosts.

   A rejected request gets a generic `404` with no Rojo-identifying body, so a prober (or a rebound page's same-origin script, which could read the body) learns nothing about whether a Rojo server is here.

2. Local-only gate on `/api/open` (`src/web/api.rs`).
  `/api/open` launches a local program (the user's editor), so it must never be reachable remotely even on an exposed bind. It now requires the immediate TCP peer (`remote_addr`) to be loopback, returning `403` otherwise. A connection forwarded over loopback by an SSH/Tailscale tunnel or a local reverse proxy passes this check, and is treated as delegated trust: the user stood up that hop and it has its own auth.

3. Startup warning for exposed binds (`src/cli/serve.rs`).
  Binding to a non-loopback address now prints a warning that the API is unauthenticated and that anyone who can reach it can read and modify the project, recommending localhost plus tunneling instead.

### Threat model and trade-offs

- Host/Origin validation is enforced for loopback and for a specific private/LAN bind. For an unspecified (`0.0.0.0` / `::`) or public bind we can't know which hostnames legitimately resolve to the server, so enforcement is disabled and the user is warned at startup instead. Consequence: such a bind has no rebinding protection, even for a browser on the same machine. This is an accepted trade-off, since exposing the unauthenticated API to the network is already the dominant risk.
- On a private/LAN bind, clients connect by IP. Accepting arbitrary hostnames would reopen the rebinding hole, since the check can't tell a legitimate hostname from an attacker's. Even on a private bind, this check does nothing against a hostile peer already on the LAN, who can reach the unauthenticated API directly; that is the network exposure the startup warning covers.
- Accepting a hostname is opt-in. Because a hostname can't be told apart from a rebound attacker domain, one is accepted only when the user explicitly lists it via `--allowed-hosts` or `serveAllowedHosts`. Listing a name trusts whatever resolves it, which for a LAN name like `mypc.lan` is the user's own DNS or hosts file. Listing hosts also opts an otherwise-unvalidated bind (`0.0.0.0` / `::` or public) back into validation, tightening it to localhost, the bind IP, and the listed hosts. The port is still checked against the bind port, and a listed hostname does not weaken the rejection of any other host.
- `Origin` is validated only when present, because non-browser clients (the Roblox plugin) don't send it. The `Host` check is the primary defense and applies to every request while enforcement is on.
- A rejection is a generic `404` rather than a `403` that names Rojo, so the rejection itself doesn't confirm a Rojo server exists.
- `Host` / `Origin` values carrying userinfo (`evil.com@localhost`) are rejected, so a value that only looks local after stripping userinfo can't slip through. IPv4-mapped IPv6 forms and IPv6 zone ids are normalized first, so they compare correctly against the bind IP and against listed IP hosts.
- The `/api/open` gate uses the TCP peer address, not a spoofable header like `X-Forwarded-For`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,7 @@ Making a new release? Simply add the new header with the version and date undern
 * Fixed the sync fallback scrambling sibling order; replacements are now re-parented ancestors-first and in their original child order. ([#1265])
 * Instances that share a name and class are now robustly matched on resync by comparing their properties, instead of relying on child order alone. ([#1266])
 * Rojo now reports a clear error instead of panicking in several cases, including when the `serve` port is already in use, when a synced file is read-only or locked, when the filesystem watcher can't be created, and when the working directory is inaccessible. ([#1267])
-* `rojo serve` now validates the `Host`/`Origin` headers on loopback binds to protect the local server against DNS rebinding, gates `/api/open` to local clients, and warns when bound to a non-loopback (network-reachable) address. ([#1270])
+* `rojo serve` now validates the `Host`/`Origin` headers to protect the local/private server against DNS rebinding, gates `/api/open` to local clients, and warns when bound to a network-reachable address. ([#1270])
 
 [#1176]: https://github.com/rojo-rbx/rojo/pull/1176
 [#1179]: https://github.com/rojo-rbx/rojo/pull/1179
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ Making a new release? Simply add the new header with the version and date undern
 * Fixed the sync fallback scrambling sibling order; replacements are now re-parented ancestors-first and in their original child order. ([#1265])
 * Instances that share a name and class are now robustly matched on resync by comparing their properties, instead of relying on child order alone. ([#1266])
 * Rojo now reports a clear error instead of panicking in several cases, including when the `serve` port is already in use, when a synced file is read-only or locked, when the filesystem watcher can't be created, and when the working directory is inaccessible. ([#1267])
+* `rojo serve` now validates the `Host`/`Origin` headers on loopback binds to protect the local server against DNS rebinding, gates `/api/open` to local clients, and warns when bound to a non-loopback (network-reachable) address. ([#1270])
 
 [#1176]: https://github.com/rojo-rbx/rojo/pull/1176
 [#1179]: https://github.com/rojo-rbx/rojo/pull/1179
@@ -54,6 +55,7 @@ Making a new release? Simply add the new header with the version and date undern
 [#1265]: https://github.com/rojo-rbx/rojo/pull/1265
 [#1266]: https://github.com/rojo-rbx/rojo/pull/1266
 [#1267]: https://github.com/rojo-rbx/rojo/pull/1267
+[#1270]: https://github.com/rojo-rbx/rojo/pull/1270
 
 ## [7.7.0-rc.1] (November 27th, 2025)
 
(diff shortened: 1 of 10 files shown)


Dekkonot (commented):
This all looks sound to me, but I have concerns.

My first one is that we should probably just return a generic 404 if an external request hits the server without permission. It's technically correct to return a 403, but from a security standpoint it feels like a better idea to just lie and say there's nothing there in case someone is e.g. trying to probe for Rojo users. My understanding is that this is standard practice for web development anyway.

The second one is that that we treat private addresses (e.g. `192.168.0.1`) as an external binding, which disables the host/origin check. I've personally had to serve to my local network before (for testing on another machine), so this isn't great. I'm not sure if this is actually something we can solve, but it's worth looking at.

boatbomber:
> My first one is that we should probably just return a generic 404 if an external request hits the server without permission. It's technically correct to return a 403, but from a security standpoint it feels like a better idea to just lie and say there's nothing there in case someone is e.g. trying to probe for Rojo users. My understanding is that this is standard practice for web development anyway.

Good point, updated to a generic 404.

> The second one is that that we treat private addresses (e.g. `192.168.0.1`) as an external binding, which disables the host/origin check. I've personally had to serve to my local network before (for testing on another machine), so this isn't great. I'm not sure if this is actually something we can solve, but it's worth looking at.

I've added Host validation for this case. It does now mean that you can't use `mypc.lan` or whatever, you must use the exact address that is being served ie `192.168.0.1` in your example. We cannot tell `mypc.lan` apart from an attacker domain.

Maybe I should add an `--allowed-hosts` arg to `serve` where you can add `mypc.lan` as an allowed host?

Dekkonot:
> Maybe I should add an `--allowed-hosts` arg to `serve` where you can add `mypc.lan` as an allowed host?

It's probably worth doing, yes. We should probably also add a field in projects to support it (we already have `port` and `address`) as well just so that you can potentially automate passing that flag away if you need to for whatever reason.

boatbomber:
Nice, then we can also protect public binds since they can specify allowed hosts.

Dekkonot (dismissed):

Dekkonot (approved):

AidenMohan:
thanks for working with me, great handling of the vuln on your side, you guys make security work easy and rewarding!
https://rustsec.org/advisories/RUSTSEC-2026-0279.html

PR #1272, opened 2026-06-02
Return 400 for missing /api/serialize IDs

### LINKED ISSUE #1271: Return an error from /api/serialize when a requested ID is missing
## Problem

`/api/serialize` detects when a requested instance ID is not present in the serve session tree, but the bad-request response is constructed without being returned. The handler then continues and can return `200 OK` with a serialized model that silently omits the missing ID.

## Why this matters

The Studio plugin relies on this endpoint when serializing requested instances. Returning success for a partially invalid request hides client/server state mismatches and makes failures harder to diagnose.

## Expected behavior

If any requested ID is not present in the serve session tree, `/api/serialize` should return `400 Bad Request` and not serialize a partial response.

AK-Khan02:
## Summary

- return the bad-request response when `/api/serialize` receives an ID that is not in the serve session tree
- add a raw `/api/serialize` POST helper for serve tests that need to inspect HTTP status codes
- add a regression test for missing serialize IDs returning `400 Bad Request`

## Validation

- `cargo test --locked serialize_missing_id`
- `git diff --check`

Closes #1271

### DIFF
--- a/tests/rojo_test/serve_util.rs
+++ b/tests/rojo_test/serve_util.rs
@@ -224,16 +224,6 @@ impl TestServeSession {
         }
     }
 
-    pub fn get_api_serialize(
-        &self,
-        ids: &[Ref],
-        session_id: SessionId,
-    ) -> Result<SerializeResponse, reqwest::Error> {
-        let body = self.post_api_serialize(ids, session_id)?.bytes()?;
-
-        Ok(deserialize_msgpack(&body).expect("Server returned malformed response"))
-    }
-
     pub fn post_api_serialize(
         &self,
         ids: &[Ref],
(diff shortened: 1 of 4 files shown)


tests/rojo_test/serve_util.rs
```diff
@@ -229,6 +229,16 @@ impl TestServeSession {
         ids: &[Ref],
         session_id: SessionId,
     ) -> Result<SerializeResponse, reqwest::Error> {
+        let body = self.post_api_serialize(ids, session_id)?.bytes()?;
+
+        Ok(deserialize_msgpack(&body).expect("Server returned malformed response"))
+    }
+
+    pub fn post_api_serialize(
```
Dekkonot, line 231: Rather than introducing a new function here, I'd rather we just rename `get_api_serialize` to be `get_post_serialize`. It's only used in two places, so a rename is low diff.

Dekkonot (changes requested):

tests/rojo_test/serve_util.rs
AK-Khan02: sounds good, lmk if it checks out now.

AK-Khan02: @Dekkonot @boatbomber

Dekkonot (dismissed):

Dekkonot (approved):

PR #1322, opened 2026-09-17
Headless API, rebased, with connection results and persistent permissions

Peteralm:
The motivation is to let agents interact with Rojo autonomously, without depending on a human clicking Connect.

I built on top of @boatbomber's fork, which adds the headless API to Rojo (#639) and already did most of the work.

**Changes**

- `ConnectAsync` now returns the result of the connection. It returned nothing before, so a caller could not tell "still connecting" apart from "wrong port".
- A granted permission now survives a restart. Before, it had to be granted again every time Studio restarted, which kept an agent from working on its own.

**Tests**

`plugin/src/HeadlessAPI.spec.lua` is new and covers:

- the result of an attempt that settles inside `startSession`;
- an attempt that settles later;
- an attempt nobody is waiting on;
- the refusal for a caller without a grant;
- the stored shape of a grant.

**Verification**

I built the plugin from this branch, installed it, and drove the API from a companion plugin. Connecting to a live `rojo serve` returns `true` in `0.240 s`; a dead port returns `false` with `Couldn't connect to the Rojo server.` in `2.269 s`. After restarting Studio, the grant is honoured without prompting.


boatbomber:
Please don't create a massive PR to main when you're really just creating a single commit to my branch. Open the PR to my branch.

